PR #58, opened 2015-11-18
Improve compability with Embarcadero C++-Builder 10 Seattle

tschoening:
I'm using Embarcadero C++-Builder 10 Seattle to build QPDF 6.0.0, which fails in some places because of missing includes, different behavior of the compiler regarding forward declarations and such. In this issue I would like to keep track of the changes needed to get your code compiled in my environment, because I'm pretty sure that most of them can easily be merged without any negative side effects. So I would love to see them considered for merging. Thanks!

### DIFF
--- a/include/qpdf/QPDF.hh
+++ b/include/qpdf/QPDF.hh
@@ -17,17 +17,17 @@
 #include <list>
 #include <iostream>
 
-#include <qpdf/QPDFObjGen.hh>
-#include <qpdf/QPDFXRefEntry.hh>
+#include <qpdf/QPDFExc.hh>
 #include <qpdf/QPDFObjectHandle.hh>
+#include <qpdf/QPDFObjGen.hh>
 #include <qpdf/QPDFTokenizer.hh>
+#include <qpdf/QPDFXRefEntry.hh>
 #include <qpdf/Buffer.hh>
 #include <qpdf/InputSource.hh>
 
 class QPDF_Stream;
 class BitStream;
 class BitWriter;
-class QPDFExc;
 
 class QPDF
 {
--- a/include/qpdf/QPDFWriter.hh
+++ b/include/qpdf/QPDFWriter.hh
@@ -24,6 +24,7 @@
 
 #include <qpdf/Constants.h>
 
+#include <qpdf/QPDFObjectHandle.hh>
 #include <qpdf/QPDFObjGen.hh>
 #include <qpdf/QPDFXRefEntry.hh>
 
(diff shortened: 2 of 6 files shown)


jberkenbilt:
Thanks. I will include these fixes in the next version.

tschoening:
I have another issue with test qutil: The expected output differs in some lines, because under Windows it's documented that the used [strerror](https://msdn.microsoft.com/en-us/library/zc53h9bh.aspx) or [strerror_s](https://msdn.microsoft.com/en-us/library/51sah927.aspx?f=255&MSPPError=-2147217396) function adds a newline to the error message. So if the exception occurs because missing files, on Windows the error message already contains one newline and gets another one using std::endl.

	----
	before remove
	exception: remove file: No such file or directory
	----
	before fopen
	exception: open /this/file/does/not/exist: No such file or directory
	----

vs.

	----
	before remove
	exception: remove file: No such file or directory

	----
	before fopen
	exception: open /this/file/does/not/exist: No such file or directory

	----

I see some possible solutions: Remove the addition of std::endl in the mentioned tests on Windows by checking _WIN32 or such, always remove an added newline in QUtil::throw_system_error or change your TestDriver to remove empty lines only containing whitespace altogether. Your already available NORMALIZE_WHITESPACE documents to remove newlines, but doesn't actually do so and because you read input files line by line one can't easily implement thing like collapsing newlines to exactly one, because each line is terminated by one newline and afterwards a completely empty line only containing the added newline from std::endl is read.

Optionally removing lines with only whitespace and using that in qutil.test seems reasonable to me.

What do you think?

jberkenbilt:
qtest lets you add filtering to the output. I'd just add filtering for that test case to filter out the extra blank lines, or I'd otherwise modify the test code in a way that doesn't require any _WIN32 checks or anything like that.

tschoening:
If I understand your filter logic correctly, it depends on invoking seperate programs? In that case I find an approach like your already available NORMALIZE_NEWLINES easier to use for every user. If you could have a look at my former commits, they look exactly like what you have donme already and are only some lines of code...

jberkenbilt:
I'll definitely take a look. I'm not sure when I'm going to have time to get to it though. Not for a few weeks at least I'm afraid.

jberkenbilt:
I took all the changes and incorporated the RM_WS_ONLY_LINES into qtest as well. Note that in the PR with the constant in void context, the correct fix was to remove then "\n", not to include it in the print. This makes sense because it was always ignored, so preserving functionality would be to remove it. qtest has a test suite that fails if the "\n" is included. Anyway, these are good changes, and I've tested that they don't break any of the other compilers I'm using. Thanks again!

PR #60, opened 2015-11-19
Useless use of a constant in void context

tschoening:
Any statement after line 390 showed the warning: Useless use of a constant in void context at /c/Users/[...]/src/qtest/module/TestDriver.pm line 390. The problem was the "\n" which surely should be included in the call to "print".


jberkenbilt:
Accepted!

PR #62, opened 2015-11-20
Improve compatibility with Windows shell

tschoening:
Please review these commits to improve compatibility with the Windows shell, even while using MSYS and UnxUtils.

### DIFF
--- a/README-what-to-download.txt
+++ b/README-what-to-download.txt
@@ -50,3 +50,34 @@ download.
    addition to including unmodified distributions of pcre and zlib, it
    includes a README file and some scripts to help you build them for
    Windows.
+
+
+Additionally for Windows, some tests require 3rd party software which
+most likely is not available if not provided by an environment like
+MSYS, Cygwin or such, but can be downloaded separately as well for an
+environment not using both of the former mentioned. Just extract or
+install the following software into separate folders each and add the
+"bin" folder to your "PATH" env variable to make executables and DLLs
+available. If installers are provided, they might do that already by
+default.
+
+ * LibJpeg (http://gnuwin32.sourceforge.net/packages/jpeg.htm)
+
+   This archive provides some needed DLLs for LibTiff if you want to
+   use the image comparison tests.
+
+ * LibTiff (http://gnuwin32.sourceforge.net/packages/tiff.htm)
+
+   This archive provides some needed binaries and DLLs if you want to
+   use the image comparison tests and is only working if DLLs from
+   LibJpeg are available as well.
+
+ * GhostScript (http://www.ghostscript.com/download/gsdnld.html)
+
+   GhstScript is needed for image comparison tests and it's important
+   that the binary is available as "gs", while its default name is
+   "gswin32[c].exe". So you need to provide an additional "gs.exe"
+   by copying one of the original files or use "mklink" to create a
+   hard-/softlink, or need to provide a custom "gs.cmd" wrapper which
+   simply forwards all args to one of the original binarys. Using
+   "mkink" with "gswin32c.exe" should be preferred.
--- a/qpdf/qtest/qpdf.test
+++ b/qpdf/qtest/qpdf.test
@@ -374,7 +374,7 @@ $td->runtest("check output",
 	     {$td->FILE => "test14-out.pdf"});
 # Test 14 also exercises writing to memory without static ID.
 $td->runtest("check non-static ID version",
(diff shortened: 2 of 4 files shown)


jberkenbilt:
Great stuff. I'll almost surely accept this. It would be great if you can let me know when you think you've fixed what you're going to fix. At that time, I can gather everything up, test and incorporate your changes, and have you do one more round of testing before I release 6.0.1. It's really hard for me to find blocks of time to work on qpdf, but usually I manage to squeeze some work in near the end of the year.

tschoening:
For now I think I'm done, I seem to be able to build everything and all tests seem to pass. That was my primary goal for now and coming to a point of using your lib, and maybe run into other issues, is still some months away. But with all your beautiful tests in place I'm very relaxed for now... :-)

	Overall test suite                                             ... PASSED

	TESTS COMPLETE.  Summary:

	Total tests: 2215
	Passes: 2215
	Failures: 0
	Unexpected Passes: 0
	Expected Failures: 0
	Missing Tests: 0
	Extra Tests: 0

jberkenbilt:
That's great. The tests are extremely comprehensive. I don't know if you've tried the large file tests or not. Please see "Additional Notes on Test Suite" in README for details. There are two categories of tests that are off by default: image comparison tests and large file tests. I don't know if the image comparison tests work on Windows at all, but there is no reason for you to run them unless you are changing significant code. They use ghostscript to render PDF files and compare resulting images, so if you didn't change any of the PDF files in the test suite, the only reason image comparison tests would fail is if there were something wrong with the way the tests are implemented. Large file support, on the other hand, could have some potential of being problematic with a previously untested compiler, so it might not be a bad idea to test it. The large file tests are tested without large files by default to make sure that the test code itself is right. Then if you enable large file tests as described in the README, it will create files of significantly more than 4 GB. I think something like 30 GB is required to run the large file test suite, and it's very slow. But for the kind of stuff you're doing here, it may be worth the effort to run them one time just to make sure everything is working with large files. Usually large files "just work" with a modern OS and a 64-bit environment, but qpdf supports large files even when built for 32-bit.

tschoening:
Finally?

	TESTS COMPLETE.  Summary:

	Total tests: 2228
	Passes: 2228
	Failures: 0
	Unexpected Passes: 0
	Expected Failures: 0
	Missing Tests: 0
	Extra Tests: 0

Tests for large file support are executed now, those for image comparison before already. I just needed to provide [LibJpeg](http://gnuwin32.sourceforge.net/packages/jpeg.htm), [LibTiff](http://gnuwin32.sourceforge.net/packages/tiff.htm) and create a link to gswin32c.exe from my already available Ghostscript to make the hard coded gs.exe available. All 3 dirs need to be available in PATH and there you go.

You were absolutely right about large file support, your code didn't recognize my compiler to have 64 Bit versions of fseek and ftell available and those tests showed that. I changed that as well in my other branch and now this test passes like all others. My SSD handles the large files very well. ;-)

Just to have it mentioned: For various reasons I don't use a Cygwin/MSYS/whatever environment and your build/test infrastructure, but created my own IDE projects and a litte cmd.exe script to run all the tests in a pre defined dir structure of my own. But like my pull requests show, only small changes were needed to support my use case as well.

jberkenbilt:
Great, thanks! If you think there are any remarks that should be made in README-windows.txt that might help other users, could you possibly do that? No problem if not -- it may be that no changes are required. Otherwise, I will likely accept this as is (or maybe with only cosmetic/whitespace changes) next time I have a chance to do some work on qpdf. I really appreciate your taking the time to do this, and it's good to know that only small changes were required and that all those were to the test suite.

tschoening:
I've added some notes about the needed libs for the tests in README-what-to-download.txt. Thought it fits a bit better there, because in the end what I did is a somewhat environment specific custom build and those people need to kow what they are doing anyways. In such cases it's most helpful to just know which binaries and DLLs are need, where to find them and let people put all that stuff together on their own. Else you might confuse some users with many different build instructions not relevant to most of them.

jberkenbilt:
Great, thanks. I may move things around a bit. I don't think I've ever run the image comparison tests on Windows. Maybe I'll incorporate this into my own testing so I can make sure it continues to work properly.

jberkenbilt:
I've accepted this change. With all three pull requests, I just took the changes, made a few tweaks, and linearized the commit history rather than merging, but your changes are all in, and they are properly attributed. Thanks!

PR #88, opened 2016-08-16
added page rotation example (it was missing & examples for this are s…

Quasilyte:
…omewhat hard to find)

### DIFF
--- a/examples/pdf-rotate.cc
+++ b/examples/pdf-rotate.cc
@@ -1,0 +1,365 @@
+#include <qpdf/QPDF.hh>
+#include <qpdf/QPDFWriter.hh>
+
+#include <vector>
+#include <regex>
+#include <algorithm>
+
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
+#include <cctype>
+#include <cerrno>
+
+#include <sys/stat.h>
+
+using std::vector;
+
+typedef QPDFObjectHandle QpdfObject;
+
+/*
+ * Rotates clockwise/counter-clockwise selected pages or page range.
+ * It is also capable of setting absolute page rotation.
+ * Check `usage` for details.
+ *
+ * User should check program return value to handle errors.
+ * Check `ErrorCode` for enumeration and `error_message` for descriptions.
+ */
+
+enum class ErrorCode: int {
+    Success, // <- Not an error
+    InvalidArgCount,
+    InputFileNotExist,
+    InvalidDegreeArg,
+    InvalidPageSelectorArg,
+    InvalidRangeSelector,
+    InvalidPagesSelector,
+    BadPageNumber,
+    BadLowRangeBound,
+    BadHighRangeBound,
+    UnexpectedException,
+    InternalRangeError,
+};
+
+struct Rotation {
+    Rotation() = default;
+    Rotation(const char* digits);
+
+    int degree;
+    bool absolute;
+};
+
+struct Arguments {
+    const char* in_file_name;
+    const char* out_file_name;
+    char* pages = nullptr;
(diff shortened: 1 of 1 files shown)


Quasilyte:
If I missed tests or build files it could be fixed.

jberkenbilt:
Thanks for this, I really appreciate your taking the time to provide this example. I have been very starved for time to work on qpdf, and I have a big backlog of things to do. Sometime I'm going to just take some time off and work on it. I don't know when. But when I do, I will definitely give this a good look and consider including it with the other examples. I can't promise when that's going to be. Thanks again.

jberkenbilt:
I've taken a little more of a look at this. This looks like a very useful example program. For various reasons, I'm not inclined to include it in "examples", but I'd like to create a "contrib" directory and include it there with the caveat that it is not "supported" but is provided in hopes of its being useful. I would not include it in the build or test suite, but I would reference the contrib directory in the README or documentation or wherever it seems appropriate. In order to do this, I would like to add comments to the top of the file indicating who contributed it (if you want, or just that it was externally contributed) and indicating that it can be used according to the same licensing terms as qpdf itself. Currently qpdf is licensed with [Artistic-2.0](https://opensource.org/licenses/Artistic-2.0) but I'm planning on relicensing it with the even more permissive [MIT license](https://opensource.org/licenses/MIT).

Here are a few reasons that I'm not planning on including this in "examples". Most of these are because it's "better" than the examples, not because it doesn't meet some standard.

* It requires c++11, and qpdf does not. I'm planning on making qpdf use some c++11 features, like <regex>, if available, but there are people who are using qpdf on a very wide range of compilers and systems, and I don't want to force a c++11 compiler.
* It doesn't have an automated test suite
* It is actually a fully useful program in its own right and is a bit complex. Most of the examples are intended to just illustrate how to use some feature of the library
* I don't want to have to support or maintain it myself.

An alternative to including it in the qpdf distribution would be for you to distribute it on your own. I'd be happy to link to it from somewhere. But I think it's good enough and useful enough that, with your permission, I'd like to include it in qpdf under these terms.

Take a look at #89 and see if this is acceptable to you. If so, I will take that, which will include your original commit as offered. Thanks again for doing this!

jberkenbilt:
Closing in favor of #89, which accepts the program into contrib. Thanks!

jberkenbilt:
I think I am going to add some kind of page rotation to qpdf 7.0.0. See #132.

jberkenbilt:
I think I may exclude the rotation example from contrib now that qpdf will do its own page rotation. It is a fine example, and I really appreciate your offering it. I used it for inspiration for my rotate option. In particular, I took the idea of allowing absolute or relative rotation from your example.

To implement actual page rotation, I added a rotatePage method to QPDFObjectHandle, so anyone who would actually want to do page rotation in qpdf would do it using that method rather than hand-coding it. Also, I found two small problems with the code when I was looking through it more carefully. One is that it superfluously calls `addPage()` after mutating the page dictionary. This is not necessary because qpdf will modify the object in place. Also, it doesn't check up the page tree hierarchy for inherited /Rotate keys. /Rotate is one of the keys that can be inherited through the pages hierarchy. The code I added to QPDFObjectHandle does this.

Even so, your example was great, and I'm sorry that it took me so long to get around to including it only to then turn around and not include it.

Thanks again for taking the time to write the code, submitting it, and inspiring me to add the feature to qpdf itself.

Quasilyte:
@jberkenbilt, I can not resist the desire to express my gratitude for this attitude towards fellow developers.

Also glad to hear about proper page rotation inside qpdf. In practice, that means that qpdf utility can be used instead of pdftk more easily.

PR #124, opened 2017-07-22
Make windows includes lowercase for cross compiling. Fixes #123

### LINKED ISSUE #123: Cannot cross compile with Windows.h (uppercase W) includes
I'm currently cross compiling qpdf from a docker container based on dockcross and MXE. Everything works (I had to pull pcre & zlib, but that's another topic) however there is an error at the compilation stage because some includes are in the form
```cpp
#include <Windows.h>
```
and due to the fact that on Linux it's case sensitive, it fails. I have created a patch to solve this issue that I will attach as a pull request to this ticket.

slurdge:


jberkenbilt:
Thank you. I will definitely accept this. I am planning some updates this summer and will ensure that this change is made, either through the PR or by tacking it onto some other Windows toolchain updates that are in my fork. Thanks for taking the time to report. I haven't tried cross-compiling qpdf for Windows. Were there any surprises? I may give it a shot myself and see if the resulting executable passes its test suite in Windows.

slurdge:
I tried cross compiling to see if it would be useful for a project like qpdf (e.g. producing exe for releases without needing a windows machine).

I ended up using dockcross which is based on docker. Basically there was no real difficulties. What I did:

1. Use the script dockcross-win64
2. git clone https://github.com/madler/zlib
3. svn co svn://vcs.exim.org/pcre/code/trunk pcre
4. In zlib and pcre ./autogen, configure, make install (with a special prefix)
5. cp *.a and *.h in external/lib and external/include
6. ./dockcross-win64 ./configure CPPFLAGS="-Iexternal/include"  LDFLAGS="-Lexternal/lib" --host x86_64-w64-mingw32 --with-random=/dev/urandom
7. ./dockcross-win64 make install CPPFLAGS="-Iexternal/include"  LDFLAGS="-Lexternal/lib"

(Roughly) It produces .exe which are quite strange in .libs but if I make make install, I get a qpdf file which, renamed to .exe, works well enough from my tests. 

My plan was to create a full docker image (based on dockcross) with dependencies already installed which purpose would be to produce the final windows executable. If you're interested, I'll post it here or some other location.

BTW: Great work on qpdf. It handles the pdfs much better than many freeware/shareware or paid applications.

jberkenbilt:
Sure, feel free to post anything you have. I have a contrib area, though maybe I haven't put it on github. Someone did a VC6 port of qpdf once which, while I didn't want to make such invasive changes to the code, I did include in the contrib area. If you have something like this, I can certainly mention it in the download area, link to it, etc.

I'm glad you are enjoying qpdf. As you use it on a wider range of files, you will find its principal weakness is that it sometimes bails out on processing files with errors. My next major area of focus is to make it much more resilient at handling invalid files.

jberkenbilt:
I took this commit and tweaked the comment just a bit for format consistency, so this is in master.

jberkenbilt:
Regarding the cross compilation changes, if you get something you think is ready for prime time, please create a new issue and I'll see about including it or reference it. :-)

PR #126, opened 2017-07-30
Included <cstdlib> in libqpdf/QPDFTokenizer.cc for strtol()

psmlbhor:
Solves Make error in libqpdf/QPDFTokenizer.cc and removes SIGSEGV from pdf-split-pages example


jberkenbilt:
I've taken the first commit and implemented the second one slightly differently -- basically following the pattern in pdf-split-pages that is done in all the other examples. Thanks!

PR #128, opened 2017-08-01
Add PCLm support to QPDF.

sahilarora535:
PCLm is a strict subset of PDF. PCLm was created by HP and is a printer friendly language. It is supported by [WifiDirect](http://www.wi-fi.org/discover-wi-fi/wi-fi-direct) and [Mopria](http://mopria.org/) printing standards for printing over the [Internet Printing Protocol](https://en.wikipedia.org/wiki/Internet_Printing_Protocol) and hence is supported by millions of printers for printing. PCLm can also be viewed by a standard PDF viewer.

Since it is a strict subset of PDF, it would be an enhancement for QPDF to support this format. The main structural differences between PDF and PCLm are:

1. For PCLm to be viewed by a normal PDF viewer, it needs some unreferenced objects to exist in the file at specific positions.
2. The file must start with a page rather than a page tree of a file catalog. The contents of every page must end before another page starts. There cannot be any ambiguous ordering in the contents of the PCLm file since the ordering is well defined. Some ambiguous ordering may show up correctly in a PDF viewer, but a printer that supports PCLm will abort the job if it get an ambiguous ordering.
3. The file body catalog and page tree must be at the end of the file just before xref.
4. PCLm files contain a magic number in the second line, which makes it look like:
```
%PDF-1.3
%PCLm-1.0
```

This pull request aims at adding PCLm support for QPDF so that QPDF can produce PCLm files as well.

### DIFF
--- a/libqpdf/QPDFWriter.cc
+++ b/libqpdf/QPDFWriter.cc
@@ -66,6 +66,7 @@ QPDFWriter::init()
     encrypted = false;
     preserve_encryption = true;
     linearized = false;
+    pclm = false;
     object_stream_mode = qpdf_o_preserve;
     encrypt_metadata = true;
     encrypt_use_aes = false;
@@ -318,6 +319,20 @@ void
 QPDFWriter::setLinearization(bool val)
 {
     this->linearized = val;
+    if (val)
+    {
+        this->pclm = false;
+    }
+}
+
+void
+QPDFWriter::setPCLm(bool val)
+{
+    this->pclm = val;
+    if (val)
+    {
+        this->linearized = false;
+    }
 }
 
 void
@@ -2254,6 +2269,13 @@ QPDFWriter::write()
 	this->qdf_mode = false;
     }
 
+    if (this->pclm)
+    {
+        this->stream_data_mode_set = true;
+        this->stream_data_mode = qpdf_s_preserve;
+        this->encrypted = false;
+    }
+
     if (this->qdf_mode)
     {
 	if (! this->normalize_content_set)
@@ -2388,6 +2410,10 @@ QPDFWriter::write()
     {
 	writeLinearized();
     }
+    else if (this->pclm)
+    {
+        writePCLm();
+    }
     else
     {
 	writeStandard();
@@ -2459,6 +2485,26 @@ QPDFWriter::writeHeader()
     // within the first 1024 characters of the PDF file, so for
     // linearized files, we have to write extra header text after the
     // linearization parameter dictionary.
+}
+
+void
+QPDFWriter::writePCLmHeader()
+{
+    setMinimumPDFVersion(pdf.getPDFVersion(), pdf.getExtensionLevel());
+    this->final_pdf_version = this->min_pdf_version;
+    this->final_extension_level = this->min_extension_level;
+    if (! this->forced_pdf_version.empty())
+    {
+	QTC::TC("qpdf", "QPDFWriter using forced PDF version");
+	this->final_pdf_version = this->forced_pdf_version;
+        this->final_extension_level = this->forced_extension_level;
+    }
+
+    writeString("%PDF-");
+    writeString(this->final_pdf_version);
+    // PCLm version
+    writeString("\n%PCLm 1.0\n");
+    writeStringQDF("%QDF-1.0\n\n");
 }
 
 void
@@ -3161,6 +3207,85 @@ QPDFWriter::writeStandard()
     {
 	QTC::TC("qpdf", "QPDFWriter standard deterministic ID",
                 this->object_stream_to_objects.empty() ? 0 : 1);
+        popPipelineStack();
+        assert(this->md5_pipeline == 0);
+    }
+}
+
+void
+QPDFWriter::writePCLm()
+{
+    if (this->deterministic_id)
+    {
+        pushMD5Pipeline();
+    }
+
+    // Start writing
+
+    writePCLmHeader();
+    writeString(this->extra_header_text);
+
+    // Image transform stream content for page strip images.
+    // Each of this new stream has to come after every page image
+    // strip written in the pclm file.
+    std::string image_transform_content = "q /image Do Q\n";
+
+    // enqueue all pages first
+    std::vector<QPDFObjectHandle> all = this->pdf.getAllPages();
+    for (std::vector<QPDFObjectHandle>::iterator iter = all.begin();
+            iter != all.end(); ++iter)
+    {
+        // enqueue page
+        enqueueObject(*iter);
+
+        // enqueue page contents stream
+        enqueueObject((*iter).getKey("/Contents"));
+
+        // enqueue all the strips for each page
+        QPDFObjectHandle strips = (*iter).getKey("/Resources").getKey("/XObject");
+        std::set<std::string> keys = strips.getKeys();
+        for (std::set<std::string>::iterator image = keys.begin();
+                image != keys.end(); ++image)
+        {
+            enqueueObject(strips.getKey(*image));
+            enqueueObject(QPDFObjectHandle::newStream(&pdf, image_transform_content));
+        }
+    }
+
+    // Put root in queue.
+    QPDFObjectHandle trailer = getTrimmedTrailer();
+    enqueueObject(trailer.getKey("/Root"));
+
+    // Now start walking queue, output each object
+    while (this->object_queue.size())
+    {
+        QPDFObjectHandle cur_object = this->object_queue.front();
+        this->object_queue.pop_front();
+        writeObject(cur_object);
+    }
+
+    // Now write out xref.  next_objid is now the number of objects.
+    qpdf_offset_t xref_offset = this->pipeline->getCount();
+    if (this->object_stream_to_objects.empty())
+    {
+        // Write regular cross-reference table
+        writeXRefTable(t_normal, 0, this->next_objid - 1, this->next_objid);
+    }
+    else
+    {
+        // Write cross-reference stream.
+        int xref_id = this->next_objid++;
+        writeXRefStream(xref_id, xref_id, xref_offset, t_normal,
+                0, this->next_objid - 1, this->next_objid);
+    }
+    writeString("startxref\n");
+    writeString(QUtil::int_to_string(xref_offset));
+    writeString("\n%%EOF\n");
+
+    if (this->deterministic_id)
+    {
+        QTC::TC("qpdf", "QPDFWriter standard deterministic ID",
+                    this->object_stream_to_objects.empty() ? 0 : 1);
         popPipelineStack();
         assert(this->md5_pipeline == 0);
     }
--- a/include/qpdf/QPDFWriter.hh
+++ b/include/qpdf/QPDFWriter.hh
@@ -308,6 +308,11 @@ class QPDFWriter
     QPDF_DLL
     void setLinearization(bool);
 
+    // Create PCLm output. Enables writing unreferenced objects,
+    // set PCLm header and writes pages before file catalog and page tree.
+    QPDF_DLL
+    void setPCLm(bool);
+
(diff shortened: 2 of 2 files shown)


libqpdf/QPDFWriter.cc
```diff
@@ -2451,6 +2476,26 @@ QPDFWriter::writeHeader()
...
+	QTC::TC("qpdf", "QPDFWriter using forced PDF version");
+	this->final_pdf_version = this->forced_pdf_version;
+        this->final_extension_level = this->forced_extension_level;
+    }
+
+    writeString("%PDF-");
+    writeString(this->final_pdf_version);
+    // PCLm version
+    writeString("\n%PCLm-1.0\n");
+    writeStringQDF("%QDF-1.0\n\n");
```
jberkenbilt: Let's not forget to make sure fix-qdf works with this or document that it doesn't. My quick reading of fix-qdf suggests that it may require a tweak to make it work.

libqpdf/QPDFWriter.cc
```diff
@@ -3154,3 +3199,88 @@ QPDFWriter::writeStandard()
...
+
+    // Start writing
+
+    writePCLmHeader();
+    writeString(this->extra_header_text);
+
+    // Image transform stream content for page strip images.
+    // Each of this new stream has to come after every page image
+    // strip written in the pclm file.
+    std::string image_transform_content = "q /image Do q\n";
```
jberkenbilt: It's hard for me to imagine how this is going to work. I guess there must be something in the cups filter that is tightly controlling how these pdfs will look?

libqpdf/QPDFWriter.cc
```diff
@@ -3154,3 +3199,88 @@ QPDFWriter::writeStandard()
...
+        enqueueObject((*iter).getKey("/Contents"));
+
+        // enqueue all the strips for each page
+        QPDFObjectHandle strips = (*iter).getKey("/Resources").getKey("/XObject");
+        std::set<std::string> keys = strips.getKeys();
+        for (std::set<std::string>::iterator image = keys.begin();
+                image != keys.end(); ++image)
+        {
+            enqueueObject(strips.getKey(*image));
+            enqueueObject(QPDFObjectHandle::newStream(&pdf, image_transform_content));
```
jberkenbilt: I wouldn't have thought of the approach of actually creating the stream at this point, but I can't think of any reason why it won't work. If this is working, it's a lot simpler than my original thought of pre-creating these and having a map to them. I guess you must be doing something external to these changes to ensure that /image points to the right thing.

libqpdf/QPDFWriter.cc
```diff
@@ -3154,3 +3199,88 @@ QPDFWriter::writeStandard()
...
+    {
+	QPDFObjectHandle cur_object = this->object_queue.front();
+	this->object_queue.pop_front();
+	writeObject(cur_object);
+    }
+
+    // Write out the encryption dictionary, if any
+    if (this->encrypted)
+    {
+	writeEncryptionDictionary();
```
jberkenbilt: Does it make sense to have encrypted PCLm files? If so, fine. If not, it might make sense to also clear the encryption parameters. There should be code in QPDFWriter that does this.

libqpdf/QPDFWriter.cc
jberkenbilt: I just saw your response to my comment on the other commit. Thanks for checking. I guess this is resolved then.

libqpdf/QPDFWriter.cc
sahilarora535: In a PDF, the page resources object looks like: 
```
 /Resources << /XObject << /I 5 0 R >> >> 
``` 
In PCLm, they are strips, similar to:
```
/Resources << /XObject << /Image0 3 0 R /Image1 5 0 R /Image2 7 0 R /Image3 9 0 R >> >>
```

In between two objects in the above PCLm example, we need the following image transform stream:
```
4 0 obj
<< /Length 14 >>
stream
q /image Do Q
endstream
endobj
```
Since the contents of this stream will always be the same, and equal to the string `q /image Do Q` according to the PCLm specifications, I though it will be better if I directly add them to where they should be. Does this make sense?

libqpdf/QPDFWriter.cc
sahilarora535: @jberkenbilt @tillkamppeter
It does not make sense to me to have encrypted PCLm files, since PCLm would not support many of the encryption keys. It may be visible in a standard PDF editor, but will not be able to print in a PCLm supported printer. Also, I just got to know that QDF mode is also not supported by the printer. Both of these phenomenon imply that neither a PCLm with QDF mode enabled nor PCLm with encryption will be a valid PCLm since both of them will be unable to print on a printer. So what do you suggest, do we drop these two methods for PCLm?

sahilarora535:
@jberkenbilt 
The PCLm page resouces xobject should look like this below, where keys starting with the string `/Image` are sorted by the number after the key:
```
/Resources << /XObject << 
/Image0 3 0 R 
/Image1 5 0 R 
/Image2 7 0 R 
/Image3 9 0 R 
.
.
.
/Image8 19 0 R 
/Image9 21 0 R 
/Image10 23 0 R 
/Image11 25 0 R 
/Image12 27 0 R 
.
.
.
```

However, in the current implementation of **`unparseObject`** in **QPDFWriter.cc**, this cannot be achieved. The keys from the current implementation of **`unparseObject`** will be lexicographically sorted in this way:
```
/Resources << /XObject << 
/Image0 3 0 R 
/Image100 5 0 R 
/Image101 7 0 R 
.
.
/Image2 9 0 R 
.
.
```
To fix this, we have three ways:
1. Not touching the current `writeObject` and `unparseObject` functions and create new `writePCLmObject` and `unparsePCLmObject` function in which key are sorted in the required manner using a custom comparator for `std::set`
2. Adding conditions in `unparseObject` and check if the file is `pclm` by checking if `this->pclm == true`. If yes, then use the custom comparator for writing the objects.
3. Replace all occurences of use of `std::set<std::string>` for child set of keys with the custom comparator similar to `std::set<std::string, comparator>`,  since PDF does not anyways care of the order of keys in the file(please correct me if I am wrong here, or if there are some test cases), but PCLm does. 

The first method will not be touching the current implementations at the cost of being code heavy. The other two will be easier to implement. What do you suggest in this case?

libqpdf/QPDFWriter.cc
jberkenbilt: I think my comment was unclear. What I'm talking about here is the placement of the call to QPDFObjectHandle::newStream. I actually think your code here is fine.

libqpdf/QPDFWriter.cc
jberkenbilt: I think you should force the file to not be encrypted. If you convert an encrypted file to PCLm, the result should not be encrypted. QPDF doesn't check restrictions on encrypted files. There's no point in doing so since it can also easily override them. I would just force the file to not be encrypted.

Don't worry about that -- I can take it at the end when integrating everything.

jberkenbilt:
> Adding conditions in unparseObject and check if the file is pclm by checking if this->pclm == true. If yes, then use the custom comparator for writing the objects.

Your option 2 is the only acceptable option. Option 1 duplicates too much logic, and option 3 will break compatibility of output files which, although not necessarily bad from a semantics standpoint, will cause a huge amount of effort for anyone who is included files created by qpdf in test suites. This includes qpdf's own test suite. I don't want to gratuitously change the output.

smilingthax:
Idea: Wouldn't it be enough to use `/Image0001`, etc. to get the correct ordering and still be PCLm compilant?

jberkenbilt:
> Idea: Wouldn't it be enough to use /Image0001, etc. to get the correct ordering and still be PCLm compilant?

Somehow I thought that was previously suggested and rejected, but if we can do it, it would be the best option. You'd just have to know how many images there are so you know how many zeroes to pad with. But using std::sort or custom comparitor isn't that hard either.

jberkenbilt:
> it would be the best option

I should say it might be the best option. Depends which ends up being easier.

sahilarora535:
@smilingthax @jberkenbilt 
This was indeed previously proposed but never rejected. We had a lot of other problems that time, and it somehow got off my mind. Thanks a lot for this. It was very easy to implement it in cups-filters itself rather than meddling with a lot of code in QPDF(see implementation here: https://github.com/sahilarora535/cups-filters/commit/144475b2b51a7269c2f8d51fc273f403944439be), and it definitely is the best method. Though it may give problems if someone tries to create a PCLm file using QPDF without using cups-filters, but we can surely add this note in some readme. 
I have pushed final changes from my side, and it would be great to have suggestions from you if I need to make any more changes.

libqpdf/QPDFWriter.cc
```diff
@@ -3165,3 +3211,82 @@ QPDFWriter::writeStandard()
...
+
+    // Start writing
+
+    writePCLmHeader();
+    writeString(this->extra_header_text);
+
+    // Image transform stream content for page strip images.
+    // Each of this new stream has to come after every page image
+    // strip written in the pclm file.
+    std::string image_transform_content = "q /image Do Q\n";
```
jberkenbilt: Is the literal name `/image` meaningful? In other words, does this work in general, or does it depend upon the incoming file looking a certain way?

jberkenbilt:
These changes look pretty good. I can take your commit, do a little work on it to integrate it into other changes I've made, and take it. I'll rebase it such that you are preserved as the commit author.

Before I do this, I need to be able to test it. Can you provide a few representative files that can be written with PCLm mode along with their expected outputs? Also, please specify the exact constraints to be imposed on the input files to make them PCLm compliant. For example, you may need to say that the pages' contents are to consist of a series of strips encoded in a certain way and named with XObject names that start with /Image and, when sorted lexically, appear in top to bottom order.

Once I have this, I can try to integrate it along with my other changes and have you do some final testing and any additional fixups you may have.

This is looking pretty close now.

libqpdf/QPDFWriter.cc
sahilarora535: @jberkenbilt This entire string `q /image Do Q` is mentioned in the PCLm Specifications sheet and works as is. This is something which needs to be there for all PCLm files regardless of their content and works in general (as per the specifications file I am using). Hence I have added it here itself.

jberkenbilt: Great, thanks for the clarification.

jberkenbilt:
The only thing remaining for me to take this (and modify it somewhat) is some way to test it. You don't have to update the test suite. Just provide the input and output files.

sahilarora535:
@jberkenbilt Please find the PCLm files in this repository I created to test them. I have tested the ones which are supported by my printer and they get printed fine: https://github.com/sahilarora535/raster-to-pclm/tree/master/pclm-files/64pixels/600dpi

jberkenbilt:
> @jberkenbilt Please find the PCLm files in this repository I created to test them. I have tested the ones which are supported by my printer and they get printed fine: https://github.com/sahilarora535/raster-to-pclm/tree/master/pclm-files/64pixels/600dpi

Are there input PDF files that generated these via writePCLm? I need the inputs and the outputs so I can have the test suite verify.

sahilarora535:
> Are there input PDF files that generated these via writePCLm? I need the inputs and the outputs so I can have the test suite verify.

Please find the original PDF files here: https://github.com/sahilarora535/raster-to-pclm/tree/master/originals
The PCLm files are converted from raster images, and you can find the raster files here: https://github.com/sahilarora535/raster-to-pclm/tree/master/raster-images/600dpi

Also, for the cases when PCLm is incorrect, it is given in **Annex B(PCLm Error Conditions)** in the specifications sheet shared with you.

jberkenbilt:
I've taken these changes by cherry-picking the commit of your other branch.

PR #134, opened 2017-08-06
Always insert newlines when newline-before-endstream (fixes #133)

### LINKED ISSUE #133: PDF/A endstream issues
This is a follow-up of bug #103. I'm using the current master version (with the `--newline-before-endstream` switch), but qpdf still breaks PDF/A compliance, apparently for the same or a very similar reason.

First, we start with a PDF/A compliant file, that I want to linearize after adding some metadata using `exiftool` (I've tried with acrobat pro dc and it also reports the file is valid PDF/A-2b):

```
# verapdf 0005d.pdf 
...
      <validationReport profileName="PDF/A-2B validation profile" statement="PDF file is compliant with Validation Profile requirements." isCompliant="true">
...
```
Now I linearize the file using qpdf. Here I try to tell qpdf to make as few modifications as possible, but I've tried with other stream options (compress, generate object streams) with the same results.
```
# qpdf --stream-data=preserve --object-streams=preserve --normalize-content=n --linearize --newline-before-endstream 0005d.pdf 0005d.tmp.pdf
# verapdf 0005d.tmp.pdf 
...
<validationReport profileName="PDF/A-2B validation profile" statement="PDF file is not compliant with Validation Profile requirements." isCompliant="false">
        <details passedRules="120" failedRules="1" passedChecks="74066" failedChecks="5">
          <rule specification="ISO 19005-2:2011" clause="6.1.7" testNumber="1" status="failed" passedChecks="0" failedChecks="5">
            <description>The value of the Length key specified in the stream dictionary shall match the number of bytes in the file following the LINE FEED (0Ah) character
			after the stream keyword and preceding the EOL marker before the endstream keyword.</description>
            <object>CosStream</object>
            <test>isLengthCorrect</test>
            <check status="failed">
              <context>root/indirectObjects[98](105 0)/directObject[0]</context>
            </check>
            <check status="failed">
              <context>root/indirectObjects[205](161 0)/directObject[0]</context>
            </check>
            <check status="failed">
              <context>root/indirectObjects[230](113 0)/directObject[0]</context>
            </check>
            <check status="failed">
              <context>root/indirectObjects[249](15 0)/directObject[0]</context>
            </check>
            <check status="failed">
              <context>root/indirectObjects[264](49 0)/directObject[0]</context>
            </check>
          </rule>
        </details>
      </validationReport>
...
```

Acrobat Pro DC also reports the violation:
imatge

Is it possible that you are adding the newline to the stream's content length (which you shouldn't do as per the specification)? I am lost as to what else could be wrong here.

You can download my test file from [here](https://la3.org/~kilburn/0005d.pdf)

kilburn:
I have drilled down into #133 and found the problem.

The current implementation of the `--newline-before-endstream` option re-used the option to write newlines when working in `qdf` mode. In that mode, the newline before `endstream` is only added if the stream content itself does not end with a newline. This may be fine for `qdf`, but it does *not* comply with the PDF/A rules (which mandate that a newline should always be added before `endstream`, disregarding whether the stream itself ends with a newline or not).

This pull request fixes the issue for me and passes all the current tests, but it can be improved in two ways:

1. A test case should probably be added to verify that the functionality doesn't break in the future. I don't know how to create a simple (minimal) pdf that has a stream whose content ends with a newline, and hence I haven't added the test. If you can help on that I will add a test to the test suite.

2. The interaction between `--newline-before-endstream` and `qdf` mode may be broken now. I don't know much about `qdf` mode and how it should operate in this instance, so I've preferred to not mess around with it (at least it will operate exactly the same as before without the `--newline-before-endstream` option.

### DIFF
--- a/libqpdf/QPDFWriter.cc
+++ b/libqpdf/QPDFWriter.cc
@@ -1587,7 +1587,12 @@ QPDFWriter::unparseObject(QPDFObjectHandle object, int level,
 	char last_char = this->pipeline->getLastChar();
 	popPipelineStack();
 
-	if (this->qdf_mode || this->newline_before_endstream)
+	if (this->newline_before_endstream)
+	{
+		writeString("\n");
+	}
+
+	if (this->qdf_mode)
 	{
 	    if (last_char != '\n')
 	    {


jberkenbilt:
Your fix looks weird to me because it might insert two newlines in some cases, but I'll study this and either accept your fix or provide a different one. Thanks!

kilburn:
On the one hand, I'm pretty sure that adding two newlines at the end if both `--newline-before-endstream` and `qdf` mode are active is wrong.

On the other hand, the bug is precisely that with `--newline-before-endstream` you sometimes *have* to have two newlines before `endstream`. This must happen when the stream itself ends with a newline: in this case, the newline that belongs to the stream itself is counted in the stream's length, whereas the second newline is not. The correct layout to maintain pdf/a compliance is:
```
stream<newline><stream contents, that happens to end with a newline><newline>endstream
```
whereas currently (without any patch), the layout ends up being:
```
stream<newline><stream contents, that happens to end with a newline>endstream
```
which is not compliant anymore because there is not a not-in-sream newline before endstream.

jberkenbilt:
Got it. I will fix this and ask you to retest as I don't think I have a tool that can test compliance.

jberkenbilt:
Closing in favor of a different fix. Please see my comments in #133.

PR #157, opened 2017-09-25
Remove libjpeg dependency from pkg-config

rohieb:
Linking QPDF with the original libjpeg-9a from http://ijg.org works well, but when I then compile cups-filters with QPDF support, cups-filters' configure is confused about it, because libjpeg does not ship pkg-config definitions:

```
[...]
checking for ZLIB... yes
checking for LIBQPDF... no
configure: error: Package requirements (libqpdf >= 3.0.2) were not met:

Package 'libjpeg', required by 'libqpdf', not found

Consider adjusting the PKG_CONFIG_PATH environment variable if you
installed software in a non-standard prefix.
```

We cannot solve this problem in QPDF, so just remove the libjpeg dependency from the `Requires.private` variable.

### DIFF
--- a/libqpdf.pc.in
+++ b/libqpdf.pc.in
@@ -6,6 +6,6 @@ includedir=@includedir@
 Name: libqpdf
 Description: PDF transformation library
 Version: @PACKAGE_VERSION@
-Requires.private: zlib, libjpeg
+Requires.private: zlib
 Libs: -L${libdir} -lqpdf
 Cflags: -I${includedir}


jberkenbilt:
I'm not sure this is the right fix. qpdf does in fact require libjpeg to link if being linked statically. I'll have too review pkg-config docs to figure out whether my config is right, but I don't think libjpeg's lack of pkg-config configuration should in any way influence what qpdf's pkg-config files have. Also, I have worked very closely with cups-filers maintainers, and no one else who is configuring cups-filters with qpdf 7 has so far encountered this problem.

rohieb:
I think most distributions just patch libjpeg to include a .pc file for it, so the error does not happen there. But I'm compiling directly from source. Another solution would be to persuade upstream libjpeg to ship a .pc file, but I guess most distros have already unsuccessfully tried that.

a17r:
jpeg-9c ships with a .pc file from upstream.

jberkenbilt:
I'm going to go ahead and close this issue then since just upgrading to a newer jpeg library should resolve the problem. Thanks for the report and update.

Issue #181, opened 2018-02-18
discussion of enhanced AppImage


KurtPfeifle:
In *"man fix-qdf"* the synopsis section states:

>     qpdf < infilename > outfilename

Shouldn't this rather read as

>     fix-qdf < infilename > outfilename

?

----

<sup>(I'm currently working to prepare a pull request for the creation of an improved version of a QPDF AppImage, which includes access to its manpages and other documentation and provides an easier discoverability of all its internal functions. That's when I noticed the above small flaw...</sup>

jberkenbilt:
Thanks for catching it just in time for 8.0.0. :-)

jberkenbilt:
I will hold off on releasing until I get your AppImage pull request. I'm probably a few days away best case anyway as I still have docs, and building my releases is still manual because of all the Windows binaries.

jberkenbilt:
I just pushed a few changes to master, so please make sure your pull request includes my latest push. Thanks.

KurtPfeifle:
My PR will include the following major changes:

* Enhanced and extended custom AppRun script with the following major features:
    - Implement a '--usage' parameter for the AppImage which helps new users to familiarize with various ways to work with it and with QPDF as a CLI in general
    - Let 'fix-qdf' and 'zlib-flate' work as "sub-commands" (not only as symlinks) to the AppImage
    - Support running the payload also when extracted into an AppDir (after running the AppImage with `--appimage-extract` the AppDir is in './squashfs-root/')
    - Support to list and display embedded manpages, PDF, HTML, READMEs, licenses and binaries (invoke AppImage with '--usage' or '--list-{pdf,html,man,readme,license}' to see them all)
    - Prepared it for future embedding of additional binaries or helper scripts (currently tested by embedding 'env', 'busybox'  and 'less' -- to be removed again if this is not useful)

* The QPDF logo is now scaled up for the 512x512 icon category (using the PNG version since I'm not sure if SVG works on all distros -- need to inquire with @probonopd)

* Added support for 'AppStream' metadata (a FreeDesktop.org "standard" which unfortunately isn't very easy to follow or validate because of its frequent changes and inconsistent tooling)

* Added comment block at start of and many comments into 'appimage/build-appimage' for the benefit of casual users who happen to run it locally on their cloned Git sources

Grab my current AppImage build here:

* https://github.com/KurtPfeifle/qpdf-appimage-testing/releases

Please test it and see what remains to be improved or should be changed.

[...]

PR #184, opened 2018-02-25
Enhanced AppImage (custom AppRun)

KurtPfeifle:
@jberkenbilt:

My PR includes the following major changes:

* Enhanced and extended custom AppRun script with the following features:
    - Implement a *`'--ai-usage'`* parameter (this may for later releases have to be changed to *`--appimage-usage`*, once the AppImage developers agree to this) for the AppImage which helps new users to familiarize with various ways to work with it and with QPDF as a CLI in general
    - Let 'fix-qdf' and 'zlib-flate' work as "sub-commands" (but also continue to work as symlinks) to the AppImage (this way the AppImage can work "out of the box" without first requiring creation of symlinks for 'fix-qdf' and 'zlib-flate'... "KISS"!)
    - Support running the payload also when extracted into an AppDir (after running the AppImage with `--appimage-extract` an AppDir is in './squashfs-root/'), which is good for debugging issues
    - Support to list and display embedded manpages, PDF, HTML, READMEs, licenses and binaries (invoke AppImage with *`'--usage'`* or *`'--list-{pdf,html,man,readme,license}'`* to see them all)
    - Prepared it for future embedding of additional binaries or helper scripts (currently tested by embedding 'env', 'busybox'  and 'less' -- to be removed again if this is not useful)

* The QPDF logo is now scaled up for the 512x512 icon category (using the PNG version since I'm not sure if SVG works on all distros -- still need to inquire with @probonopd)

* Added support for 'AppStream' metadata (a FreeDesktop.org "standard" which unfortunately isn't very easy to follow or validate because of its frequent changes and inconsistent tooling)

* Added comment block at start of and many comments into 'appimage/build-appimage' for the benefit of casual users who happen to run it locally on their cloned Git sources

You can test my most recent AppImage builds here:

* https://github.com/KurtPfeifle/qpdf-appimage-testing/releases

Please test it and see what remains to be improved or should be changed.

See also my comment #181 at "discussion of enhanced AppImage"...

I didn't try to squash my commit history (and don't know if this would be desirable for you even), since I've no experience with this procedure and I vaguely remember that you had expressed an intention to do this yourself. Hence, there is quite some back and forth with the commits...


jberkenbilt:
Thanks! I'll grab this, tweak it as needed, though hopefully not very much, and include it in 8.0.0. After 8.0.0 is out, we can revisit the question of continuous releases. It could be useful, but probably not for the GSoC work because that's mostly going to be library stuff that won't be exposed in the command-line tool. I'll just have to figure out how I want it all to work. Thanks very much for all your work on this.

jberkenbilt:
I've accepted your changes and tweaked them in #185.
I'm closing this and #181. We can continue discussion in #185.

Issue #178, opened 2018-02-01
Feature request: Higher level abstractions for objects such as forms in a PDF.


sahilarora535:
It is hard to represent all types of objects in a PDF using the `QPDFObjectHandle` class. An example of this is forms in a PDF.

@smilingthax says:
> to handle Form fields, the current approach of QPDF, "everything is as QPDFObjectHandle" is probably not the best way. A better way could be to introduce a "QPDFFormFieldHandle" (containing a reference to the annotation and the page, and also to the node/field in the /AcroForms - tree), but there are certainly other options.

@jberkenbilt says:
> I'm certainly willing to think about higher level abstractions for specific kinds of things. I doubt we're going to end up with QPDFFormFieldHandle, but I agree that some kind of framework for higher level abstractions is appropriate.

The issue originated while trying to fill a PDF which contains forms using QPDF.
Issue and conversation: https://github.com/OpenPrinting/cups-filters/issues/8

CC: @tillkamppeter

smilingthax:
Clarification: A terminal field's `/Kids` array can point to multiple annotations for the same field (possibly on different pages). A merge of field dictionary and annotation dictionary (i.e. bullet point iii.) is not possible in this case. An annotation dictionary must not re-used on more than one page (and probably should not appear more than once in a given `/Annots` array).

psmlbhor:
Now that @sahilarora535 has mentioned it, I remember asking @jberkenbilt a similar question quite a long time ago where he mentioned he will be thinking about it in the future. 

Having a separate object handle for Form fields, I think, will greatly reduce the work that needs to be put in while handling those fields. For eg. we need to keep track of the child of every object that comes under `/Annot`. A new object handle can just manage it allowing the user to focus on the task that he wishes to perform rather than managing the intricacies of the form objects.

jberkenbilt:
@sahilarora535 You asked: "Jay, what do you think? Do you want to think this through for yourself first, or would should we provide some ideas on how a form API for QPDF could look like?"

Feel free to provide some ideas if you want, though just be aware that I am very particular and may or may not agree with them. :-) I'm polishing up qpdf 8, and then I'll have to take another break, but next time I get some time to put into qpdf, I'll probably pick this up. I will give it some thought after thoroughly reading and digesting everything here, and I will be interested in any input.

sahilarora535:
[...]

PR #213, opened 2018-06-18
Helper api

jberkenbilt:
This pull request is here to serve as a place to discuss and review proposed API enhancements for higher level APIs as discussed in #178. This branch will be rebased and reworked multiple times. I'll push to it periodically when I have something in a reviewable state.


jberkenbilt:
@sahilarora535 @smilingthax Here is the pull request I mentioned in #178. At the time of this writing, I have pushed up changes to support a page-level API using the helper approach. Questions and comments welcome.

smilingthax:
Hmm, `QPDFPageDocumentHandler::getAllPages()` (still) returns a `vector<QPDFObjectHandle>`, whereas I'd most certainly would want / "need" a `vector<QPDFPageObjectHelper>`...(?)

One of the strengths of the new paradigm, as I currently understand it, would be that e.g. the page cache would finally become part of the `QPDFPageDocumentHandler` -- i.e. the Handler becomes the "container" for all the internal bookkeeping used to implement certain operations ... (?)

Other than that the concept looks good, and I'm looking forward to seeing how well this works for the form stuff.

(minor sidenote -- probably not solvable without major changes to libqpdf: You might have seen the `node-libqpdf` binding (on my github) I experimented with a few months ago... I found that it is nearly impossible to create a resource-managing wrapper ("smart pointer") around a `QPDF` object ... the new classes add just more of those internal references to QPDFs that are hard to track...)

jberkenbilt:
Regarding your comment about getAllPages returning a vector of QPDFPageObjectHelper, I definitely agree that that would be a more natural interface. I may put this in a comment, or I may provide an alternative interface. The reason I ended up not doing it stems from a design decision we made a long time ago: getAllPages actually returns a reference to an internal structure that is updated dynamically by things like addPage. In order to make all those semantics continue to work, we would have to double work. There are comments in the code that talk about this, and there are a number of tests the explicitly exercise the mutability of the vector returned.

Personally I think this is a bad pattern, and I'm not sure why we decided to do it this way. I would be surprised if many people are relying on this behavior or would consider it the right behavior.

Maybe I should change the method in `QPDFPageDocumentHelper` to return `vector<QPDFPageObjectHelper` and change the one that calls the other method to something like `getMutablePageObjectVector` for emphasis. It would be an easy change.

jberkenbilt:
Actually, I am not going to expose the old `getAllPages` or the update cache method in `QPDFPageDocumentHelper`. I am going to push up a new version in a few minutes for your comments. Your input is very valuable. I thought of this while I was doing, but your comment is compelling me to fix it. The result is going to be better because of your review and comment.

smilingthax:
Well, I deliberately ignored any legacy constraints that could also play a role in my comment.

One way (you seem to be heading in another direction) would be to put all mutators that also affect the cache (like addPage) (only) on the `QPDFPageObjectHelper`. Basically the legacy API on `QPDF` would have to internally use a `QPDFPageObjectHelper` as well...

But cache invalidate is inherently a hard problem. At some point one has to rely on the programmer to (e.g.) not  use a `QPDFPageObjectHelper` after he loaded an new pdf file into the parent `QPDF` (or at least to call the page cache invalidation method).

jberkenbilt:
I don't want to get distracted by reworking the page API. I would surely create extra work in maintaining compatibility. Unfortunately, certain API and design decisions stick around, and we get stuck with them. I will provide a newer, preferred interface for future use, and use this to validate the helper pattern. When we work out the pattern for forms, we will not repeat those mistakes.

But you definitely looking at this the right way, and your comments are validating my approach. It is in fact the case that state handling and so forth can be confined to the helper classes. It was a mistake to ever expose an internal cache. Had the cache been kept private, and had getAllPages returned a vector instead of a vector const&, we would have been in better shape.

smilingthax:
From a programmer perspective, it'd be worse, when there is an internal cache, but I can't invalidate it, although it has become stale. In the new approach it would make sense that I have to create a new `QPDFPageDocumentHelper` after manually tinkering with some of the underlying pdf objects.
The unexpected thing with it currently being a `const &` is that it could change under my feet.
OTOH, other programming languages tend towards an iterator-based interface for thinks like `getAllPages` - supported by language features like generators / coroutines and spreads / expansion...

jberkenbilt:
I'm almost done with the changes. The new interface won't expose a cache. I think it will be better.

jberkenbilt:
The cache is still exposed by the older methods. I can't do anything about that.

jberkenbilt:
Changes pushed. Care to take a fresh look?

qpdf/qpdf.cc
```diff
@@ -1905,14 +1905,13 @@ static void do_show_pages(QPDF& pdf, Options& o)
...
+    std::vector<QPDFPageObjectHelper> pages = dh.getAllPages();
     int pageno = 0;
-    for (std::vector<QPDFObjectHandle>::iterator iter =
-             pages.begin();
+    for (std::vector<QPDFPageObjectHelper>::iterator iter = pages.begin();
          iter != pages.end(); ++iter)
     {
-        QPDFObjectHandle& page = *iter;
-        QPDFPageObjectHelper ph(page);
+        QPDFPageObjectHelper ph(*iter);
```
smilingthax: Why the extra `QPDFPageObjectHelper` and not use `iter->getObjectHandle()` directly (or make `ph` a reference to *iter)?

jberkenbilt: oh, editing error in updating the interface. Good catch, thanks.

smilingthax:
We can always come back to the page stuff, as we gain new insights from the form stuff. I have no further comments for now.

jberkenbilt:
Thanks. As soon as make check finishes, I'll push up some fixups for the last around of commits that get rid of extraneous copies and conversions between object handle and object helper. Thanks for your comments. I'm going to take a first crack at defining some method signatures for forms tonight before I stop so I can hit the ground running on those tomorrow.

jberkenbilt:
Assuming everything goes well, my next push will squash some of these commits.

jberkenbilt:
@smilingthax I have pushed a series of commits with the first iteration of a proposed API for form helpers. There are no setters yet, just methods to query stuff. It should not take me long at all to implement what I have so far. Then I will see about adding a few methods to set some things such as field values or appearance streams.

smilingthax:
Re: "TBD: where do methods to alter field and annotation values go"
This is where it gets tricky...
- It might make sense to put those methods that can be executed locally onto the QPDFFormFieldObjectHelper, and only the other ones on QPDFAcroFormDocumentHelper... -- esp. things like `setValue()`: As a programmer it feels a bit unnatural to require two different objects for get vs. set... one might inquire: "Why aren't they just all on QPDFAcroFormDocumentHelper, when that's where the setter has to be?" ... I do understand your desire to not reference the DocumentHelper from ObjectHelper; OTOH, a QPDFObjectHandle does know the QPDF it came from (e.g. for stream content, I think) after all...
- Now, in most cases `/NeedAppearances` has to become `true`, thus there will be no setters left on the ObjectHandle ... unless we change the design and do reference the DocumentHelper (or at least "certain important QPDFObjectHandles" ...) ... hmm... I'm quite undecided...

---
A few use cases to think through ... let's fast forward to the point where qpdf would support creating new forms from scratch:

1a. user first creates a new form field, possibly not at the root, but below some QPDFObjectHandle / QPDFFormFieldObjectHelper
   -> (e.g.) `QPDFFormFieldObjectHelper QPDFAcroFormDocumentHelper::createField(const string& name, ..., QPDFFormFieldObjectHelper *parent = nullptr)`
1b. alternatively, user reuses an existing form field
2. user creates a new form(widget) annotation - on a certain page (or: must be added later to *exactly one* page ?)
   -> probably `QPDFAnnotationObjectHelper QPDFAcroFormDocumentHelper::createAnnotation(const QPDFFormFieldObjectHelper &field)`  [NOTE: const & or by value ?] and
        `QPDFAcroFormDocumentHelper::addAnnotationToPage(QPDFFormFieldObjectHelper &field, const QPDFPageObjectHelper &page)`
       (or a method that does both at once).

To remove an annotation or field, things would go the other way round.
- What happens when the user tries to remove a form field that is still associated with an annotation?
- A merged field+annotation dictionary means that a selective removal of only the annotation means removal of a certain set of keys from the dictionary.
- Adding a second annotation to an existing field also requires extracting those keys into a separate dictionary.
- Maybe: additional method that "removes field *and all* associated annotations".

---
Re: "QPDFFormFieldObjectHelper::getQuadding" ...
-> Seems to make sense: When implementing the appearance stream (re-)generator, it either has to look up each key it needs via (e.g.) `...ObjectHelper::getWithInheritance(key,default=QPDF_Null)`, or we add a specific method for each key we're interested in (and that's specifically marked as inheritable in the pdf spec!)... at least, we can *transparently* put caching logic (analogous to the "push inherited fields to pages" implementation) into QPDFAcroFromDocumentHelper, if we want to.

smilingthax:
Re: "QPDFObjectHandle::Rectangle"
This rings a bell... 
In pdftopdf, I basically use `Rectangle QPDFPageObjectHelper::getMediaBox()`, `getCropBox()`  (i.e. `if (hasKey(/CropBox)) return /CropBox; else return /MediaBox;`), and similar for getTrimBox/getArtBox ...

I also have `makeBox(double x0,y0,x1,y1)` ...

jberkenbilt:
> Re: "TBD: where do methods to alter field and annotation values go"

You're thinking in the right way. I'm pretty sure they will go in the doc helper because most of them will need to go there and, as you say, it's not good for the programmer to have to go to two places. Regarding the object handle knowing the QPDF it comes from, this is not true of direct objects that are created by the programmer until they are actually hooked in, so I don't want to ever rely on this.

>  When implementing the appearance stream (re-)generator, it either has to look up each key it needs via (e.g.) ...ObjectHelper::getWithInheritance(key,default=QPDF_Null), or we add a specific method for each key we're interested in (and that's specifically marked as inheritable in the pdf spec!)

I fully expect to provide a method to look things up with inheritance and also to provide a few helper methods for specific, common cases. A method to push inheritable things down may be useful at some point, but as you point out, we can't flatten the structure without potentially breaking fully qualified names.

Right now, I am writing the code to assume that an annotation can be referenced by exactly one field, which is what the spec says. If an annotation is referenced by more than one field, I will warn and just keep one of them arbitrarily. As I traverse the fields tree, I am going to sanity check /Parent. If /Parent is absent and shouldn't be absent, I'm going to set it. If it is present and set to the wrong value, I'm going to modify it and give a warning. I am going to take care not to add /Parent keys to annotation dictionaries that are not merged with field dictionaries. The main purpose of all of this is that it makes it safe to walk up the parent structure for inheritable resources. While everything that can go wrong does go wrong in the wild somewhere, I suspect inconsistencies in the tree structure will be very rare in the wild since mostly things won't work if you do that.

jberkenbilt:
> I also have makeBox(double x0,y0,x1,y1) ...

I'll add a constructor like this to Rectangle.

jberkenbilt:
@smilingthax Do you have any test PDF files with forms that span the range of structures for AcroForm, such as a multi-level hierarchy, terminal fields with /Kids that contain multiple annotations, and non-trivial inheritance? If not, I will hand-generate some and test them with various readers to make sure I have the semantics right.

jberkenbilt:
It will probably only take me half an hour to generate them, so no big deal if you don't. I am going to have to generate files with various types of errors in them anyway.

smilingthax:
> I'm pretty sure they will go in the doc helper because most of them will need to go there and, as you say, it's not good for the programmer to have to go to two places. Regarding the object handle knowing the QPDF it comes from, this is not true of direct objects that are created by the programmer until they are actually hooked in, so I don't want to ever rely on this.

Well, my line of thinking was: Every property of a form field that does imply changes to either pages/annotations or the (fields) tree structure, should not only be gettable, but also settable via the QPDFFormFieldObjectHelper -- even if this might mean that we have to somehow reference the `/NeedAppearances` key (which should probably be cached...) from the `QPDFFormFieldObjectHelper`, possibly by referencing the `QPDFAcroFromDocumentHelper` -- unless that would be too much of a burden ...
For all the other operation I fully agree that they have to be on the DocumentHelper.

> A method to push inheritable things down may be useful at some point, but as you point out, we can't flatten the structure without potentially breaking fully qualified names.

What I had in mind was pushing down the fields (possibly only in the cache copy) without actually flatten the structure (i.e. different from what happens with the pages tree).

> Right now, I am writing the code to assume that an annotation can be referenced by exactly one field, which is what the spec says. If an annotation is referenced by more than one field, I will warn and just keep one of them arbitrarily.

I'm not referring to annotations, which can only be associated with a single field, as your correctly said,  but to a field, which references more than one annotation in its /Kids array.
When one deletes such a field, and there is at least one annotations for it, these would either block the deletion, or that have to be deleted, too -- i.e. removed from their page's /Annots array (if they have an page -- I'm not sure whether we should allow annotations apart from a page they are on).
The underlying assumption here is: I would not allow (widget) annotations not associated with a form field at all (this does not generalize to non-widget annotations, though). All the annotation and appearance properties basically depend on the underlying field type. One could imagine re-associating annotations to a different form field, but only *when the field type matches*, but the user could still create a *new* annotation from the desired form field, or do such tricky operation using low-level QPDFObjectHandles.

> I'll add a constructor like this to Rectangle.
... and some way to convert a Rectangle to a QPDFObjectHandle? 
Because at some point I probably want to call (e.g.) `QPDFObjectHandle::replaceKey(key, ...rect...)` ...

> Do you have any test PDF files with forms ...
No. I haven't actually seen such pdfs in the wild yet ... (but I haven't searched a lot, either).

PS: Your newly added `insertRootFieldsForPage(QPDFPageObjectHelper, std::set<QPDFObjectHandle>&)` method puzzles me a bit... Why the "root objects" and not the "terminal fields" ... and why `set<QPDFObjectHandle>` and not `set<QPDFFormFieldObjectHelper>` ... (things will probably become clearer with your next commits)

jberkenbilt:
> PS: Your newly added insertRootFieldsForPage(QPDFPageObjectHelper, std::set<QPDFObjectHandle>&) method puzzles me a bit... Why the "root objects" and not the "terminal fields" ... and why set<QPDFObjectHandle> and not set<QPDFFormFieldObjectHelper> ... (things will probably become clearer with your next commits)

Page splitting loses form fields. I'd like to fix that, and this is exactly the information that will be needed to fix it. For example, `qpdf --empty --pages file-with-forms.pdf -- out.pdf` loses form functionality, and in order to know which root elements to preserve in the document's /AcroForm dictionary, I need this information. Simply creating `/AcroForm` and copying those objects to it should be sufficient.

jberkenbilt:
Regarding your other comments, generally speaking, caches are evil. In my professional life, I have seen countless weeks wasted tracking down obscure bugs caused from caches. Caching stuff like `/NeedAppearances` is not worth it. It's very cheap to look that up every time. I think the qpdf library strikes a pretty decent balance between performance and maintainability. There are certainly extraneous memory allocations, function calls, and object copies, but the amount of time spent doing this stuff is most likely small in comparison to the file operations that are at the heart of working with PDF. Caches are suitable for cases where it's very expensive to compute things or where there's lots of derived information. Anyway, I'm pretty clear about where I will and won't use caches, and my default is to trade away performance in favor of simplicity where possible.

These are good thoughts about the setting interfaces, but I doubt I'm going to have time to implement anything comprehensive this week. In addition to establishing these APIs, there are some bugs to fix and a handful of unrelated enhancements. I'll get as far as I can get in another day or two of work, and that will be it for the current iteration. At the end of the week, I will likely release 8.1. At this point, I expect not to have to break any binary interfaces.

Regarding setting and higher level interfaces in general, while I agree that special methods for getting and setting fields would be valuable, given the realities of my ability to work on qpdf, I am strongly favoring code for things that are complicated to do. So operations that require updating /AcroForm and also /Annots, knowing when to change /NeedAppearances, potentially unmerging annotation dictionaries from field dictionaries, and those kinds of things are much higher priority than just providing methods that do nothing but call replaceKey. Pull requests are always considered, of course, but as you know, I'm pretty tight about qpdf's API and very conservative about doing things that increase maintainability. Though adding simple one-line methods is, of course, low risk and easy...but unless it's comprehensive, it doesn't actually save the user from having to call replaceKey....

smilingthax:
Well, you have the page annotations, then:
1. Find (via the cache) the terminal form fields: `getAnnotationsForPage` + `getFieldForAnnotation` .(maybe a combined method might be useful in other places and more efficient?)
2a. Filter the form field tree to contain only those root + internal nodes that are on a path to one of the terminal form fields. Also filter the `/Kids` of each terminal field to contain only annotations from (one of) the pages in the (whole...) page set (...to be extracted).
2b. Alternatively, for each terminal field extract the path to the root and filter `/Kids` to only the annotations that are on the given page. Finally, for all pages, merge all the paths back together to create the new fields tree of the output pdf.

Just copying all the root nodes would be too much. At least all the referenced annotations on pages that are not copied *have to* be removed. Which in turn could lead to inner or terminal fields that could be pruned (but keeping them is rather harmless) ...
[*All* the root nodes would be enough, when we want *all* the pages... but then one would not need the page argument ...]

When merging pages from different pdfs, things become more difficult. Merging of two field trees has to happen; unfortunately the pdf specification has some restrictions on fields with the same field name (e.g. "In particular, field dictionaries with the same fully qualified field name must have the same field type, value and default value."). What this boils down to (disregarding special tricks/cases) is: When merging two trees with root nodes having the same name, either they are merged into one output node (only possible when the contents of the dictionaries match!), or one of the fields has to be renamed...

jberkenbilt:
Yeah, handling all the document-level things like outlines, threads, forms, etc., in splitting and merging is pretty hard. I may not end up implementing that method until I actually need it.

smilingthax:
> > I'll add a constructor like this to Rectangle.
>
> ... and some way to convert a Rectangle to a QPDFObjectHandle?
> Because at some point I probably want to call (e.g.) QPDFObjectHandle::replaceKey(key, ...rect...) ...

Not sure you got my intention here... the quoting indentation ended up a bit misleading. As it does not make sense to overload each array/object setter (like replaceKey) that currently receives a QPDFObjectHandle to also recognize `Rectangle`, either the Rectangle class should contain a method to convert it to a QPDFObjectHandle (you probably dislike the conversion "operator QPDFObjectHandle", so let's be more explicit), or we add a `QPDFObjectHandle::newFromRectangle`, or something similar. All I'm saying is that a `Rectangle` that does not round-trip back to a PDF representation seems half baked...

Forget about getMediaBox/getCropBox/getTrimBox/getArtBox... these are convenience methods that *could* find a home on the newly invented QPDFPageObjectHelper, but there are definitely by far more important things to work on.

jberkenbilt:
> All I'm saying is that a Rectangle that does not round-trip back to a PDF representation seems half baked...

You're right, I didn't get that you were saying that. You are correct. I will make it possible to create an object handle from a rectangle, most likely `QPDFObjectHandle::newArray(Rectangle)`, though your suggestion of `QPDFObjectHandle::newFromRectangle(Rectangle)` is also good. I'll add that too and have it just call the other one.

jberkenbilt:
Pushed *untested* implementation of QPDFAcroFormDocumentHelper

jberkenbilt:
Test fail because of coverage errors. I've started adding QTC coverage cases for things I need to remember to write tests for.

jberkenbilt:
commenting out coverage until I'm ready to test so I don't get meaningless test failures

jberkenbilt:
Pushed *untested* implementations of object helpers. I'm going to take a short break before I start writing some tests.

smilingthax:
Re: QPDFAnnotationObjectHelper::getAppearanceStream(...bool on)
 - The names `/On` and `/Off` in section 8.4 (Annotations) of the pdf reference are only exemplary. 
    In fact 8.6.3 (Field types), under "Check Boxes", it says: "The appearance of the off state is optional but, if present, must be stored [...] under the name `Off`. The **recommended (but not required) name** for the on state is `Yes`". 
   So how does one know which name to use (a bool is certainly not enough)? -> "The `V` entry [...] holds a **name** object [...] which is used to select the appropriate appearance [...]".
   Probably the method should take another string for this subname, as it already does for `which` ...
---
Re: QPDFFormFieldObjectHelper::getFullyQualifiedName
- I think you have to concat until you reach the root (no more `/Parent`) even if on the path some `/T` are missing...

---
Re: QPDFAcroFormDocumentHelper.cc 
 - `this->m->cache_valid` is never set to true?
 - in `analyze()`: `if ((! acroform.isDictionary()) && (acroform.hasKey("/Fields")))`  should probably be something like `(!dictionary)||(!hasKey(...))` ?
 - "// All annotations should have been encountered by traversing /AcroForm"
   -> No. All `/Subtype /Widget` annotations could be expected to be found from `/AcroForm`, but not the other types (like Links, Underlines, Movies, ...)
       This raises another interesting question...: Will the non-AcroForm annotations (which could be present without `/AcroForm` even existing in the document catalog) have (in the future) their own DocumentHelper? Are `QPDFAnnotationObjectHelper` actually only really used for form-related (i.e. `/Widget`) annotations? I could imagine an `AnnotationDocumentHelper`, which allows queries like "getAllAnnotationsOfType" or "getPageForAnnotation" -- but then things become tricky when `QPDFAnnotationObjectHelpers` are "shared" by the AcroFromDocumentHelper (cache 1) and the AnnotationDocumentHelper (cache 2) ... 
 - "// Treat the annotation as its own field."
    "          this->m->annotation_to_field[og] =  QPDFFormFieldObjectHelper(annot);"
   -> Well, considering that `annot` could be a pure annotation dictionary, not merged with a field dictionary, the resulting QPDFFormFieldObjectHelper then is a rather odd creature... OTOH, we can easily create such things by calling the public QPDFFormFieldObjectHelper(QPDFObjectHandle) -ctor with an object that certainly is not a form field - or even a dictionary at all ...
 - in `traverseField()`: "// [...] Otherwise, the annotation fields might be there￼ to be inherited by annotations below it."
    -> I don't think this is possible... Inheritance is only defined for form dictionary keys; also non-terminal fields can only be pure field dictionaries (Table 8.69 Entries common to all field dictionaries, under /Kids).
 - You might want to output warnings when some of the checks like `! field.isIndirect()` in `traverseField()` cause an early return...

jberkenbilt:
Good comments. I already caught the cache_valid never being set mistake in my testing. I'll look through the other ones. Some of them are likely pointing out errors in my code, but keep in mind that there are millions of ways that people write non-compliant PDFs. I've gotten pretty good at anticipating the kinds of mistakes people make. I can almost guarantee that PDF creators will get wrong the stuff about which fields can be inherited and when to merge annotation and field dictionaries. I'm trying to anticipate some of the kinds of errors people will make. But I'll make sure my interpretations are not going to mess up on valid files. I'll push up what I have right now, which some tests coded but not yet invoked -- I'm invoking them manually at the moment. Then I will go through your comments item by item. Thanks.

jberkenbilt:
> The names /On and /Off in section 8.4 (Annotations) of the pdf reference are only exemplary. ...
My reading of the spec here is that `/AS` is to be used. I will change the method so that `on` is a string that defaults to the empty string. If not specified and there is a value in `/AS`, I will use that. Otherwise, I will not look for subkeys.

jberkenbilt:
> - I think you have to concat until you reach the root (no more `/Parent`) even if on the path some `/T` are missing...

Yeah, I guess so. The spec is not clear about what to do if some levels are missing. I'm just going to skip missing levels including the "." rather than generating multiple "."'s. It's hard for me to see how this is going to matter for us, really.

jberkenbilt:
> in `analyze()`: `if ((! acroform.isDictionary()) && (acroform.hasKey("/Fields")))`  should probably be  ...

Yeah, I botched my parentheses. Good catch.

jberkenbilt:
> No. All `/Subtype /Widget` annotations could be expected to be found from `/AcroForm`, but not the other types

Right again. I modified the check to only find missing /Widget annotations. This whole bit of code is probably superfluous, but in the simple case of a merged field/annotation dictionary, I can well imagine that someone could write code that would just forget to include it in /AcroForm, and some reader or form field extractor out there could still work. Then someone will report a bug that some xyz software can handle their file but qpdf can't see the form field.

jberkenbilt:
I'm not ready to think about how this code will work with other annotation types. I think I haven't done anything at this point to exclude other annotation types since all this special case logic about widget annotations is in the AcroForm document helper.

jberkenbilt:
@smilingthax I believe I have addressed all your comments. I enhanced the comment in the block of code that's adding orphaned widget fields to make my intent a little more clear.

jberkenbilt:
Actually I forgot the comment about the warnings for truncating when fields aren't indirect.

jberkenbilt:
Now I have addressed all your comments.

smilingthax:
When accessing an annotation dictionary, I would use QPDFAnnotationObjectHelper ... but that does not even have a method to look up a key in the inheritance chain. Generally speaking, an annotation dictionary does not have a `/Parent` key (when the key is there, it is only because of the merger with a field dictionary)... when the pdf contains non-terminal field dictionaries (i.e. esp.: containing `/Kids`) containing annotation keys ("merged dictionary"), then it certainly isn't because the erroneous pdf producer expects inheritance for annotations to happen. It even can't expect that a normal pdf viewer will even consider to try using both the merged dictionary *and* what it finds below `/Kids` -- which could be fields or annotations, or (violating the spec again) a mixture of fields and annotations, because all that would require non-trivial special case handling in the viewer... 
Don't get me wrong, I'm not against trying to repair damaged pdfs. I just think that we should limit our attempts to "reasonable" things (like missing `/Type` in an appearance stream) and not blatant ignorance of defining principles (like: a field is either *non-terminal*, i.e. has **only** fields as children, or it is *terminal*, i.e. has **no** fields as children, but only annotations [...and, for #(kids)=1, the annotation dictionary could be merged into the field dictionary, ... and #(kids)=0 is actually not allowed, ... and ...]).

> My reading of the spec here is that /AS is to be used. I will change the method so that on is a string that defaults to the empty string. If not specified and there is a value in /AS, I will use that. 

I totally forgot about `/AS` ...  from Example 8.17 (Section 8.6) I derive: The distinction between `/AS` and `/V` seems to become important for radio buttons, where each radio of a group has one distinct "on" name like `/MasterCard`, `/Visa`, ..., and they all have `/Off`. One of them has `/AS` set to its "on" name, the others have `/AS /Off`. The `/V` is on the `/Parent` and has the name of the selected radio child...

> Actually I forgot the comment about the warnings for truncating when fields aren't indirect.

That comment was meant as just one example of where a warning probably makes sense. Other places could be when encountering a loop via `/Kids` or `/Parents`, or when a dictionary is expected but the type does not match...

Another thought I had, which goes in a similar direction, was to make the constructors of (some of) the ObjectHelpers private, so that they can only be generated by the DocumentHelper, which in turns makes sure to only return ObjectHelpers pointing to sensible QPDFObjectHandles of the right type (or possibly null).

smilingthax:
> I'm not ready to think about how this code will work with other annotation types. 

Ok.

> I think I haven't done anything at this point to exclude other annotation types since all this special case logic about widget annotations is in the AcroForm document helper.

I just wanted to point out that we might end up with a problem in the future, when we have/need two DocumentHelpers (for normal annotations vs. AcroForm/widget annotations) that both try to cache information "attached to" QPDFAnnotationObjectHelpers, but these DocumentHelpers don't know of each other...

jberkenbilt:
I don't think you've properly understood my reasoning on the repair. Most of your comments in the first paragraph do not apply to the code I've written. Unless I've misread the spec, there is no inheritance for annotations, and I have not written any inheritance for annotations. I didn't create extra accessors to look up arbitrary fields in the annotation dictionary because you can just use `getObjectHandle().getKey(...)`. I don't know where you're getting the idea of annotation inheritance unless I misspoke in one of my earlier paragraphs. I am not supporting inheritance of annotation fields. I am supporting inheritance of fields in form field dictionaries that we don't know about. I am not blocking creation of one kind of helper with the wrong kind of underlying object. Doing so is a dead end. I tried that with pages, and every kind of sanity check that a page was really a page backfired with at least one malformed file that would work fine if I just treated the thing as a page.

The purpose of my handling of "orphaned" widgets in this way is to specifically handle the case of a properly merged field/annotation dictionary that is simply missing from the /AcroForm tree. I don't think it's a blatant misunderstanding. I think it's an obvious error.

I don't wish to be dismissive of your comments as they are almost always good. I just think we are misunderstanding each other.

Do you see a specific place in the code where this code will misinterpret a correct file? Do you see any place in the code where adding additional checks will have any affect other than to prevent qpdf from processing some files that it might otherwise be able to process? It's entirely possible that I have either misread something in the spec or coded something incorrectly. I haven't done thorough testing yet, and I wrote the code pretty fast. Before I consider this done, I will have run various tests across various files, some from the wild and some of my own creation, and I can also review the sections of the spec while my implementation is fresh in my mind.

jberkenbilt:
> I just wanted to point out that we might end up with a problem in the future, when we have/need two DocumentHelpers (for normal annotations vs. AcroForm/widget annotations) that both try to cache information "attached to" QPDFAnnotationObjectHelpers, but these DocumentHelpers don't know of each other...

The AcroForm document helper is not an annotation helper. It is a form helper. Since forms are partially implemented with annotations, it uses the annotation object helper. If there is some need for some other kind of document helper that also needs to know about annotations, then it will also use the annotation helper. There is nothing in the annotation helper that knows that it is being used for forms. As such, I don't think I see the problem you are alluding to. If the annotation helper were polluted with logic specifically related to forms, then the problem you are concerned about would be a risk.

jberkenbilt:
I'm tied up with other stuff tomorrow morning. I will be back at this by around 11:00 a.m. UTC-0:400. If you can present clear examples of where my API is wrong, I will adjust. Otherwise, I will proceed with testing, during which I may discover problems with my existing implementation. Once I have all the read-only stuff working, I will squash commits and, if time permits, maybe consider some of the modify stuff, but more likely, I will switch focus to the other qpdf-related issues I want to fix. Please let me know if you think these APIs will simplify any of either the form-flattening or banner code.

smilingthax:
> Please let me know if you think these APIs will simplify any of either the form-flattening or banner code.

I think this covers most of what Sahil needs, except for the modification part... and these operations should also force `/NeedAppearances true`... but I'm not sure how you would want to implement that...

It also looks pretty good wrt. to implementing (re-)generation of appearances.
Still, I'd like to hear your opinion on whether, and then how this operation could be integrated into qpdf in the future. Probably `QPDFAcroFormDocumentHelper` would be the right place... but where would you want to put the field-type specific (re-)generator implementation? Also where should this "add char widths", "convert text from field encoding to font encoding" and "parse the `/DA` string to obtain the `Tf` operators arguments"(your recent token-filter probably help here, too) go?

The actual flattening finally just has to, for each page, obtain `/Annots` and do its thing for each of the `/Widget` annotations in there (requiring only the annotation and the page dictionary; the AcroForm / fields tree is not needed here any more, as the field content was copied into the appearance stream [or `/AS` was set according to `/V`] of the annotation during the (re-)generation step).

jberkenbilt:
Thanks for your clarification. I must have had a brief brain outage when I was coding that. It's not that much code. I'm just going to reread the relevant spec sections and the reread through the code. I think that, at the beginning, the distinction between fields and annotations was murky in my mind, though it is very clear now. Some of the first bits of code probably reflected this murkiness.

Your help with this is very valuable. You're really the only other person who has studied the code in any depth and provided useful input at this level. When I'm back online later today, I will address the above issues. After sleeping on it, I think you're right that efforts to repair are probably not as useful as I was thinking. Once I actually coded analyze, it was simpler than I originally had in my mind. My main concern was just that, when iterating through /Annots and finding widget annotations, I always wanted to be able to map them back to fields. Anyway, there's nothing like a good night's sleep, so I'll take a pass through the code again later this morning when I'm back from my appointments.

smilingthax:
> I am not blocking creation of one kind of helper with the wrong kind of underlying object.

Yes, but you also won't do any check for the right type in non-trivial methods. IIRC Pranjal once did some operations on an `QPDFObjectHandle` without checking its type, and got an exception/segfault, which he didn't understand. With an `...ObjectHelper` as further wrapping around the QPDFObjectHandle, it would be even more difficult for the programmer to understand why his code / his use of qpdf then fails only for certain pdfs. In node-libqpdf I specifically added `assertInitialized()` to many QPDFObjectHandle method wrappers to obtain exceptions (which I would convert to JS exceptions) instead of the whole VM crashing with a segfault...

---
Re: QPDFAcroFormDocumentHelper::getAnnotationsForPage

In its current form, this method is not specific to AcroForms, although it is part of the AcroFormDocumentHelper interface... Now, I'm not sure whether we will ever need a full AnnotationsDocumentHelper... But what could be helpful specifically for AcroForms would be: `getWidgetAnnotationsForPage`.
The next realization: Conceptually this would only be a specialization of `getAnnotationsOfSubtypeForPage(page, subtype)`... 

Idea: `vector<QPDFAnnotationObjectHelper> QPDFPageObjectHelper::getAnnotations()` Possibly with overload `...PageObjectHelper::getAnnotations(const string &subtype)`, and/or `AcroForm::getWidgetAnnotationsForPage` as special case for /Subtype /Widget ...

jberkenbilt:
> In node-libqpdf I specifically added assertInitialized() to many QPDFObjectHandle method wrappers to obtain exceptions (which I would convert to JS exceptions) instead of the whole VM crashing with a segfault...

In qpdf 8, I replaced almost all cases where qpdf called assert and replaced it with calls to various exceptions or with sensible defaults. It's a very tough call, and it's one thing I would change if I could change it in a reasonable way....one has to strike a balance between helping programmers that use the library vs. allowing qpdf to handle broken files. This is your classic static vs. dynamic typing. The annoyance with QPDFObjectHandle is very much like programming in groovy or perl where treating something as a type that it isn't results in a runtime error. At one point, I actually considered changing the API to make it much harder to do this by allowing people pass defaults when querying objects and/or by giving warnings. The problem is that QPDFObjectHandles and the underlying objects can exist in a vacuum, and also, making those kinds of changes would require radical rewrite of parts of the library. I'm afraid it's a weakness we have to live with because there's no way I am going to have time to improve it much beyond its current state. However, at least in qpdf 8, any QPDFObjectHandle created through parsing a file contains enough context that it can be used in a good warning. Most of the time, if you access on object as something of the wrong type, you get both a warning and a sensible default. All that said, there's only so much a library can do to protect against programmer errors. The qpdf library does require a higher level of knowledge of PDF than, say, poppler, itext, pdfbox, or similar libraries that try to abstract away knowledge of the PDF spec. All that said, your comments are valid, and I will take another pass though code with them in mind and see what I can tighten up. At least qpdf should make it easier for a programmer to notice if they are accidentally doing the wrong thing.

----

I like the idea of a specialized method to get all annotations of a particular subtype. I will add that.

jberkenbilt:
getAnnotationsForPage belongs in page document helper. I will move it there, add a specialized version, and change the one in the form helper to specifically request widget annotations.

jberkenbilt:
I'm getting rid of the code that adds missing parents.

jberkenbilt:
@smilingthax I've pushed up more fixes. Does this feel better to you? Take a look at the specific commit 4bb0f2d66c2c8938846bd5b1b12771eac5c5cc04 if you want. Once you've looked at it, I'll squash with the other commits. I think this should help with some of your objections, and I think it simplifies and improves the behavior as well.

include/qpdf/QPDFAcroFormDocumentHelper.hh
```diff
@@ -22,6 +22,52 @@
...
+// * Interactive forms are discussed in the PDF Specification (ISO PDF
+//   32000-1:2008) section 12.7. Also relevant is the section about
+//   Widget annotations. Annotations are discussed in section 12.5
+//   with annotation dictionaries discussed in 12.5.1. Widget
+//   annotations are discussed specifically in section 12.5.6.19.
+//
+// * What you need to know about the structure of interactive forms in
+//   PDF files:
+//
+//   - The document catalog contains the key "/AcroForms" which
```
smilingthax: It's "/AcroForm"

include/qpdf/QPDFAcroFormDocumentHelper.hh
```diff
@@ -22,6 +22,52 @@
...
+//     may be merged with the field dictionary. In that case, the
+//     field and the annotation are in the same object. Note that,
+//     while field dictionary attributes are inherited, annotation
+//     dictionary attributes are not.
+//
+//   - A page dictionary contains a key called "/Annots" which
+//     contains a simple list of annotations. For any given annotation
+//     of subtype "/Widget", you should encounter that annotation in
+//     the "/Annots" dictionary of a page, and you should also be able
+//     to reach it by traversing through the "/AcroForms" dictionary
```
smilingthax: Again: "/AcroForm"

jberkenbilt:
pushed up a fix to a typo in a comment

libqpdf/QPDFAcroFormDocumentHelper.cc
```diff
@@ -128,7 +124,11 @@ QPDFAcroFormDocumentHelper::analyze()
     // All Widget annotations should have been encountered by
     // traversing /AcroForm, but in case any weren't, find them by
     // walking through pages, and treat any widget annotation that is
-    // not associated with a field as its own field.
+    // not associated with a field as its own field. This just ensures
+    // that requesting the field for any annotation we find through a
+    // page's /Annots list will have some associated field. Note that
+    // a file that contains this kind of error will probably not
+    // actually work with most viewers,
```
smilingthax: "most viewers,"  ... comma what?

jberkenbilt:
Fixed typos

smilingthax:
Nothing catches my eye after the recent changes.

There are some things in my head when it comes to modifying the structures (adding new fields, etc), but that's for another time.

jberkenbilt:
Thanks again. As I said, this is better for your review and participation. I will squash these and work on testing, then I'll assess about the modifications.

jberkenbilt:
Regarding write functionality, I can think of many possibilities around manipulation of the fields tree and so forth, but I think these two basic ones would be pretty trivial to add:
* Create a new object to contain a merged field/widget annotation; populate it with `/Subtype /Widget`, and hook it into the containing pages' `/Annots` field and the documents `/AcroForm` field, creating each as needed
* Modify the value of `/NeedAppearances`
* Set the value (`/V`) of a widget annotation while also setting `/NeedAppearances true`
I think these modifications would be sufficient to handle most of bannertopdf.

One additional thing that could help would some assistance in generating appearance streams, such as taking the value of `/DA`, doing the marked text substitution, etc. -- the stuff the lexical changes were added to support. I don't have that fresh in my mind, but read the relevant spec sections a few weeks ago and manually edited a few appearance streams in qdf files to convince myself that I understood how they worked.

Longer term, I could adding support for splitting merged field/annotation dictionaries (though this requires a lot more knowledge about keys than I think I want to deal with) and doing other manipulations on the fields tree, but I suspect other things will be of higher value for my limited available time.

Of course, support for flattening forms could also be added, though I still think it makes sense to get that stuff working outside the library first. Doing it in the general case is going to be very difficult once you deal with XFA, older formats with boundaries, and all sorts of other nuances. I have a feeling the poppler/ghostscript approach is going to be the approach for a while. But I think the existing changes will make it easier to write that.

I'd be interested to know if you have additional thoughts. I will probably distill them into a section in qpdf's TODO file in case I can't get back to it for a long time. Depending on how the next couple of days go, I could conceivably do the first set of bullets this week. We'll see.

jberkenbilt:
I know you mentioned other stuff about adding and deleting fields in earlier discussions.

jberkenbilt:
Possibly creating a field with child annotations would also be doable. This seems to be required for radio buttons.

smilingthax:
- Setting `/V` (incl. `/NeedAppearances`) is probably the most useful right now.

- "[...] some assistance in generating appearance streams [...]"... I don't feel like I currently can give a lot of input on that, without having to dig through that part of the spec again (and/or trying to implement parts of the process). BTW: XFA appearance streams *have to* be generated by the viewer application that makes changes to the form (and thus is already capable of handling rich text), so no code needed for that.

- Form flattening then simply means: Adding an image draw command with the appearance stream as source (they are actually `/XObject`s, although the  `/Subtype` seems to be missing in some pdfs -> must be added by us, when missing!) to the page's content stream. And there's a bit of bookkeeping involved, like calculating+setting the correct context matrix for the drawing command (depends e.g. on `/Rect` and the Flags), adding it to the resource dictionary without a key collisions (the /XObject can have the expected key also stored as `/Name`, which has to also be changed). All in all nothing exciting, but will require some time for testing, etc. 

- I'm not sure what you're alluding to in "older formats with boundaries"...
- But I don't think that any special cases for XFA are needed here.
- It's just that you can't flatten without having up-to-date appearance streams first.

> "Longer term, I could adding support for splitting merged field/annotation dictionaries (though this requires a lot more knowledge about keys than I think I want to deal with) and doing other manipulations on the fields tree"

Hmm... IMO that's where it actually gets interesting.... I think that the library cannot (and should not) shield the user from understanding the difference between an annotation (dictionary) and a form field (dictionary). Therefore I'm not to keen on providing a high-level "create a new field with annotation" method, together with some special handling of radio buttons... What the library can do is relieve him from the duty of having to deal with merged vs. non-merged dictionaries. Some Ideas:
 - `bool QPDFFormFieldObjectHelper::isTerminal()`
 - `void QPDFAcroFormDocumentHelper::addAnnotationToField(QPDFFormFieldObjectHelper &field, QPDFAnnotationObjectHelper &annot)`
   -> this is only possible for `field->isTerminal()`
   -> must split the field/annotation dictionary, when it is merged
   -> Where does the new QPDFAnnotationObjectHelper come from? Does the user have to create a QPDFObjectHandle from scratch and convert it to a 
   -> sanity check that the annotation is not used for two fields...
   -> third argument to add annotation to page for convenience? (via method in `QPDFPageDocumentHelper::addAnnotation(page,annot)` that makes sure that the annotations is not used twice [problem: we don't have a cache for that right now, esp. not in the PageDocumentHelper...])
 - `void QPDFAcroFormDocumentHelper::removeAnnotationFromField(...)`
   -> but still part of the page (? - convenience: `, bool removeFromPage`? [Problem: we don't know the page the annotation is on - nobody does... -> instead of putting addAnnotation/removeAnnotation in the PageDocumentHelper, an AnnotationDocumentHelper seems to be the appropriate place...])
   -> split of dictionaries might be necessary here, as well
   -> How to deal with nodes and paths in the fields tree that have no Kids? - the pdf spec does not want them to exist, and actually they are neither terminal nor non-terminal ...
 - `void QPDFAcroFormDocumentHelper::addField(QPDFFormFieldObjectHelper &field, QPDFFormFieldObjectHelper *parent = null)`
   -> `parent==null`: add to root (`/AcroForm`'s `/Fields` array)
   -> only for `!parent->isTerminal()`
 - `void QPDFAcroFormDocumentHelper::removeField(QPDFFormFieldObjectHelper &field)`
  -> has to somehow deal with the `/Kids`, which could be just annotations (remove? split field/annotation dictionary?), or even a whole subtree of fields+annotations...

=> This is probably too much to implement right now. But the gist is: A method to split a field/annotation dictionary is the low-level primitive required here... as we (currently) do not know (without brute force) the page an annotation is on (to make changes to `/Annots`), the annotation dictionary *must* keep the existing objId, and the split off field dictionary is a new object. The method also has to fix the links (`/Parent`, `/Kids`). Merging, on the other hand, is entirely optional - everything will still work when we don't implement it and generate one-element `/Kids`.

Regarding the need to hard-code a list of all the possible keys ... it's not too bad, actually:
 - Annotation of type Widget: /Type, /Subtype, /Rect, /Contents, /P, /NM, /M, /F, /AP, /AS, /Border, /C, /StructParent, /OC, /H /MK, /A, /AA, /BS
   -> easy way out: everything not in this list is moved to the field dictionary
 - Field (all possible `/FT` types): /FT, /Parent, /Kids, /T, /TU, /TM, /Ff, /V, /DV, /AA [!! must become part of both dictionaries !!], /DA, /Q, /DS, /RV, /Opt, /MaxLen, /TI, /I, /Lock, /SV

The safe way to treat unknown keys would be to simply duplicate them into both dictionaries (as needed for /AA).

Actually the splitting could become part of the existing `getFieldForAnnotation` method:
`QPDFFormFieldObjectHelper getFieldForAnnotation(QPDFAnnotationObjectHelper annot, bool separate=false)`
When `separate==true` the method makes sure that annot and the returned field are split, when they are not already.

jberkenbilt:
Thanks for your comments. Yeah, looks like we're pretty much on the same page. Your idea of duplicating keys we don't know whether are part of the field or annotation is a good idea. I hadn't thought of that, but maybe I would have if I actually started to code it. :-)

The generating appearance stream thing is basically using `/DA` and `/V` to construct an appearance stream that displays the text in `/V`. It's pretty straightforward for cases where `/V` can be represented with characters in the standard fonts and the field is a text field. For other kinds of fields, the appearance stream is more complicated. Regarding boundaries, if I recall the spec correctly, prior to the introduction of appearance streams, there were fields controlling how boundaries looked around rectangles and such. I think they are safe to ignore with newer readers, and I think I would just ignore them. The complication of generating appearance streams comes when there are characters that require additional fonts, as we have already discussed. Basically `/V` can be a Unicode string, but the string in the appearance stream has to be properly encoded/character mapped for the font, and writing that code is way out of scope.

At the end of all this, whatever I don't get to, I will copy your and my comments into a new issue and make a note in the TODO file that all previous thoughts are gathered into that issue.

I'm most of the way through testing and have hand-constructed files with non-trivial inheritance and a few other details. There are some bugs in atril where it doesn't render some fields properly that the PDF viewer in chrome does. I'm not surprised though....I ran into a problem in my day job (which doesn't ordinarily have anything to do with PDF, but I tend to get called upon when someone in the company has a PDF problem!) regarding Arabic values in forms. So many different readers all broke in different ways, as did various Java PDF libraries. Support for forms just isn't that good, I guess.

smilingthax:
Hmm, one other feature that comes to my mind (from pdftopdf, and will be necessary for form flattening, again):
When I add commands to a page's content stream, I have to make sure the graphics state is in a "known" state, ideally: the initial state. While this is trivial for prepended content, the trick to make this work for appended content is to make sure that everything before is properly wrapped in q/Q (push/pop graphics state).
Now, the "safe" option is to prepend a `q` and append a `Q` for before appending my new content. The graphics stack size is limited, though.
So a better idea would be to have a method in QPDF that uses your recent lexical additions to walk the existing content stream and checks whether any operators (or, more exactly: any graphics state changing operators - but a conservative approximation is certainly good enough) appears a the "top level" (i.e. outside of a q/Q pair). In that case the whole existing content stream has to be wrapped by prepending q/appending Q (as I current do)...

jberkenbilt:
Tests are passing. There were a few minor problems, but nothing significant. I have to stop for the day since I have rehearsal tonight, but when I resume, I will update ChangeLogs and documentation and most likely merge this increment, though until I do an actual release, I still reserve the right to change it. Or I may hold off on merge just a bit to give @sahilarora535 a little longer to look at it. I will probably proceed with the other non-form-related fixes before I move onto writing.

jberkenbilt:
I ended up deciding to just add a basic method to set the value of a field, supporting Unicode encoding and modifying /NeedAppearances. I felt that this was a small amount of work that I could do easily while it's fresh in my mind, and it would take care of a lot of what is needed for bannertopdf. I will push up my comments in a few minutes.

jberkenbilt:
Best way to look at latest changes is just to look at the last few commits, especially "Allow setting a form field's value"

jberkenbilt:
I added an example, updated docs, ChangeLog, etc. I am now moving onto the next item.

PR #222, opened 2018-07-24
Ensure zlib-flate doesn't link with an old libqpdf

ryandesign:
See: https://sourceforge.net/p/qpdf/bugs/17/

Parts of the fix for that bug report had previously been committed in 30dbf94f53a3fd9760242883bdc5bddbaa0c9f44 and 6299c64cf3351fb1935319378aed421e26ed2f90.

### DIFF
--- a/zlib-flate/build.mk
+++ b/zlib-flate/build.mk
@@ -19,4 +19,4 @@ $(OBJS_zlib-flate): zlib-flate/$(OUTPUT_DIR)/%.$(OBJ): zlib-flate/%.cc
 	$(call compile,$<,$(INCLUDES_zlib-flate))
 
 zlib-flate/$(OUTPUT_DIR)/$(call binname,zlib-flate): $(OBJS_zlib-flate)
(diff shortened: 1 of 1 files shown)


jberkenbilt:
Thanks for this. I guess I must have overlooked this when fixing the problem earlier. I will accept this change next time I work on qpdf.

kilburn:
@jberkenbilt I was about to make a pull request to macports for the latest 8.3.0, and I've seen their system is still applying this patch. Are you merging it soon? Is there some reason holding you back from doing it?

jberkenbilt:
@kilburn My checklist includes looking at `issues` but not `pulls`, and I forgot about it. I also didn't see the mention for some reason. Sorry about that. I'm updating my checklist.

jberkenbilt:
I checked my email, and for some reason, I never received the notification. Weird. Anyway, it's included in 8.4.0. I'm building 8.4.0 now and will release this morning.

PR #261, opened 2018-10-15
TODO

jberkenbilt:
testing pipeline triggers again

### DIFF
--- a/TODO
+++ b/TODO
@@ -30,11 +30,6 @@ Soon
 
     - See ../misc/broken-files
 
- * Some qpdf --check tests are fragile on Windows. The output gets
-   truncated. This happens in the loop for content preservation tests.
-   Figure out the source of the fragility.
-
-
 Next ABI
 ========
 
--- a/azure-pipelines/build-windows
+++ b/azure-pipelines/build-windows
@@ -29,8 +29,7 @@ installdir=install-$tool$wordsize
 rm -rf $installdir
 ./config-$tool --with-windows-wordsize=$wordsize --enable-show-failed-test-output
 make -j$(nproc)
-# XXX Fix test fragility -- it's always check status in the content preservation tests.
-make -k check || echo "WARNING: TESTS FAILED"
+make -k check
 make install
 
 v=`(cd $installdir; ls -d qpdf-*)`
--- a/qpdf/qtest/qpdf.test
+++ b/qpdf/qtest/qpdf.test
@@ -2847,10 +2847,36 @@ foreach my $file (@files)
 			 {$td->STRING => "",
 			  $td->EXIT_STATUS => 0});
 
+            # This test is fragile on Windows for unknown reasons.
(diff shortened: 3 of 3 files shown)


lkillgore:
Hi @jberkenbilt, I'm from MS and I'm looking into why this isn't triggering builds for you.  Installing and uninstalling the marketplace app will break existing pipeline definitions (which I can walk you through fixing), but I'm not seeing this webhook in our server logs (it *seems* GitHub isn't sending it to us), so there's something else going on here.

I'll post here when I get more information.  I'm sorry for the inconvenience.

jberkenbilt:
@lkillgore Thank you for looking into this. I'm really excited about using azure pipelines in this way and hope this enables you to find and fix some underlying problem. Regarding breaking pipeline definitions, at this point, the integration between https://github.com/jberkenbilt/qpdf and https://dev.azure.com/qberkenbilt/qpdf is working, but not the integration between https://github.com/qpdf/qpdf and https://dev.azure.com/qpdf/qpdf. I can trigger builds on https://dev.azure.com/qpdf/qpdf manually though.

jberkenbilt:
@lkillgore I have a few questions about all this. I don't want to abuse your attention and convert you to my personal tech support, so please feel free to redirect me. :-)
* After this was fixed, I wanted to remove the temporarily integration I created between my fork (jberkenbilt/qpdf) and qberkenbilt/qpdf in azure. I did this by going into github settings for my jberkenbilt account, selecting applications, configuring azure pipelines, unchecking jberkenbilt/qpdf, and saving. This took me through the whole configuration process, which I went ahead and completed, and the result is that I now have another set of builds under the pipeline with a new definition id, and there doesn't seem to be anything I can do about it.

a

Would it have worked for me to just not go through the dialog? The marketplace app is nice in that it "automatically" does the wiring, but the downside of that is always that things are obfuscated away from end users, who in this case, are often developers...

jberkenbilt:
@lkillgore Just pushed to master, and the CI build kicked off, so both problems appear to be resolved. Thanks again for your help and attention. Free CI for my open source project that covers Linux, Windows, and Mac is a dream come true.

lkillgore:
@jberkenbilt, yes, the flow we're using is tailored to the first-time pass.  We're trying to make things as easy as possible for the common case, so yes, that is an odd experience to remove a repository and end up adding a pipeline.

I think what you're looking to do is delete the pipeline:
image

In the future, adding pipelines or editing pipelines will likely be more straightforward using the UI in Azure DevOps.  (at least I do... but, I'm probably biased :)

I hope that helps!

jberkenbilt:
@lkillgore Yes, deleting the pipeline worked. A subsequent push went back to the original one, which is exactly what I wanted. I had poked around, but I had not been able to find a menu with a delete it in. I'm sort of a hard-core CLI guy...this whole world with web UIs and stuff is too confusing. :-) Anyway, thanks again for your great help. At this point, I consider my issue to be fully resolved.

PR #290, opened 2019-02-14
The kfo-PDF files for testing need to be copied using "binmode"…

ams-tschoening:
…or Windows will introduce "\r\n". The following is an example for errors logged during running the tests after they failed, which were fixed after adding "binmode" for copying the files.

    qpdf: selecting --keep-open-files=n
    qpdf: processing 001-kfo.pdf
    WARNING: 001-kfo.pdf: file is damaged
    WARNING: 001-kfo.pdf (offset 556): xref not found
    WARNING: 001-kfo.pdf: Attempting to reconstruct cross-reference table

### DIFF
--- a/qpdf/qtest/qpdf.test
+++ b/qpdf/qtest/qpdf.test
@@ -1675,6 +1675,7 @@ $n_tests += 4;
     for (my $i = 1; $i <= 201; ++$i)
     {
         open(F, sprintf(">%03d-kfo.pdf", $i)) or die;
+        binmode F;
         print F $content;
         close(F);
     }


jberkenbilt:
Thanks! As soon as CI is green, I will merge.

jberkenbilt:
I wonder why this doesn't fail for me. Anyway, it definitely appears to be a correct change.

ams-tschoening:
I must admit I'm not using plain MSYS2 to run the tests, but instead what is included in the Windows distribution of GIT. While I'm pretty sure it's MSYS2 as well, or at least parts of it, `binmode`-stuff might simply be Windows-native for me, while non-Windows even on Windows for you.

jberkenbilt:
Yes, your theory is probably right. So I could potentially make this mistake again and not notice, but your change is clearly an improvement, so assuming it doesn't introduce other problems (and I can't imagine that it would), I think it's a good change.

PR #295, opened 2019-02-14
Undefined functions because of missing std:: or header.

ams-tschoening:
My Embarcadero C++Builder RAD 10.2 was missing either headers or the explicit namespace `std::` on some function calls. Again, I think being explicit here shouldn't interfere with other compilers.

### DIFF
--- a/examples/pdf-split-pages.cc
+++ b/examples/pdf-split-pages.cc
@@ -8,9 +8,10 @@
 #include <qpdf/QPDFPageDocumentHelper.hh>
 #include <qpdf/QPDFWriter.hh>
 #include <qpdf/QUtil.hh>
-#include <string>
+
 #include <iostream>
-#include <cstdlib>
+#include <stdlib.h>
+#include <string>
 
 static char const* whoami = 0;
 static bool static_id = false;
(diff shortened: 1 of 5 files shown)


examples/pdf-split-pages.cc
```diff
@@ -49,7 +49,7 @@ static void process(char const* whoami,
 void usage()
 {
     std::cerr << "Usage: " << whoami << " infile outprefix" << std::endl;
-    exit(2);
+    std::exit(2);
```
jberkenbilt: Instead, please replace `#include <cstdlib>` with `#include <stdlib.h>`

libqpdf/QPDFTokenizer.cc
```diff
@@ -180,7 +180,7 @@ QPDFTokenizer::resolveLiteral()
                     num[0] = p[1];
                     num[1] = p[2];
                     num[2] = '\0';
-                    char ch = static_cast<char>(strtol(num, 0, 16));
+                    char ch = static_cast<char>(std::strtol(num, 0, 16));
```
jberkenbilt: Instead, please replace `#include <cstdlib>` with `#include <stdlib.h>`

ams-tschoening:
Did all changes as requested and hope you don't bother that I slightly changed order of includes as well. Some of the files clearly used alphabetical order while others didn't. Only minor changes.

PR #297, opened 2019-02-14
std::list<QPDFOutlineObjectHelper> didn't compile in C++Builder.

ams-tschoening:
The class `QPDFOutlineObjectHelper` contains a private member class storing some data and containing the following line:

> std::list\<QPDFOutlineObjectHelper\> kids;

That line doesn't compile in my C++Builder 10.2 because the type is not fully defined at the time it's needed for the list. Not sure why this works elsewhere, but it's a known limitation in theory:

http://www.drdobbs.com/the-standard-librarian-containers-of-inc/184403814
http://codeverge.com/embarcadero.cppbuilder.cpp/e2450-undefined-structure/1053917
https://stackoverflow.com/a/45554279/2055163

Because you are using `PointerHolder` at the same place for `parent` already, I simply suggest moving to that for the `std::list` as well. Because things get pretty long and difficult to maintain this way, I introduced two new `typedef`s shortening names a lot and making future replacements easier. At some places method signatures now even fully align! ;-)

I didn't care about breaking API/ABI-changes, but didn't get anything else to work as well. So if you have another solution, please let me know. Thanks!

### DIFF
--- a/examples/pdf-bookmarks.cc
+++ b/examples/pdf-bookmarks.cc
@@ -135,23 +135,23 @@ void show_bookmark_details(QPDFOutlineObjectHelper outline,
     std::cout << outline.getTitle() << std::endl;
 }
 
-void extract_bookmarks(std::list<QPDFOutlineObjectHelper> outlines,
+void extract_bookmarks(std::list<PointerHolder<QPDFOutlineObjectHelper> > outlines,
                        std::vector<int>& numbers)
 {
     numbers.push_back(0);
-    for (std::list<QPDFOutlineObjectHelper>::iterator iter = outlines.begin();
+    for (std::list<PointerHolder<QPDFOutlineObjectHelper> >::iterator iter = outlines.begin();
          iter != outlines.end(); ++iter)
     {
         ++(numbers.back());
-        show_bookmark_details(*iter, numbers);
-        std::list<QPDFOutlineObjectHelper>::iterator next = iter;
+        show_bookmark_details(**iter, numbers);
+        std::list<PointerHolder<QPDFOutlineObjectHelper> >::iterator next = iter;
         ++next;
         bool has_next = (next != outlines.end());
         if ((style == st_lines) && (! has_next))
         {
             numbers.back() = 0;
         }
-        extract_bookmarks((*iter).getKids(), numbers);
+        extract_bookmarks((**iter).getKids(), numbers);
     }
     numbers.pop_back();
 }
--- a/examples/pdf-bookmarks.cc
+++ b/examples/pdf-bookmarks.cc
@@ -135,23 +135,23 @@ void show_bookmark_details(QPDFOutlineObjectHelper outline,
     std::cout << outline.getTitle() << std::endl;
 }
 
-void extract_bookmarks(std::list<PointerHolder<QPDFOutlineObjectHelper> > outlines,
+void extract_bookmarks(std::vector<QPDFOutlineObjectHelper> outlines,
                        std::vector<int>& numbers)
 {
     numbers.push_back(0);
-    for (std::list<PointerHolder<QPDFOutlineObjectHelper> >::iterator iter = outlines.begin();
+    for (std::vector<QPDFOutlineObjectHelper>::iterator iter = outlines.begin();
          iter != outlines.end(); ++iter)
     {
         ++(numbers.back());
-        show_bookmark_details(**iter, numbers);
-        std::list<PointerHolder<QPDFOutlineObjectHelper> >::iterator next = iter;
+        show_bookmark_details(*iter, numbers);
+        std::vector<QPDFOutlineObjectHelper>::iterator next = iter;
         ++next;
         bool has_next = (next != outlines.end());
         if ((style == st_lines) && (! has_next))
         {
             numbers.back() = 0;
         }
-        extract_bookmarks((**iter).getKids(), numbers);
+        extract_bookmarks((*iter).getKids(), numbers);
     }
     numbers.pop_back();
 }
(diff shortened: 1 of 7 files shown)


jberkenbilt:
Yes, I can't take this because of the API/ABI breakage. I'll look into it a little more and see if I'm doing something that's actually wrong.

jberkenbilt:
I'm going to go ahead and close this, but if you find a good solution, please feel free to report. Thanks.

ams-tschoening:
> Yes, I can't take this because of the API/ABI breakage.

Just recognized that you are intending to release QPDF 9 as new major release next: Why is breaking the API/ABI a problem with a new major release for you? What's your overall policy on this, never change anything to become backwards incompatible at all?

jberkenbilt:
My policy is that breaking ABI causes a major release (basically semantic versioning). I like to do that as little as possible because it is disruptive to packagers as it makes people have to relink, but when necessary, I will break ABI and do a major release. qpdf 9 breaks ABI, so now is an excellent time for ABI-breaking changes, if any.

My policy for API is that I try to avoid ever breaking the API because that causes work for everyone who uses qpdf. In the past, I have broken API only when there was no good way to avoid it, and most of the time, API breakage has been on minor or advanced parts of the API.

There may come a time when I drop support for non-C++11 and consider something radical like dumping PointerHolder in favor of C++11 auto pointers, but I haven't quite gotten that far. It's also conceivable that I could break API for performance reasons. For example, there are methods right now that allow exposing an Array QPDFObjectHandle's elements as a std::vector and a Dictionary QPDFObjectHandle's elements as a std::map. If I change the underlying representations for performance reasons, which is a distinct possibility, I would necessarily remove those. Also, the worst API mistake in all of qpdf is returning a reference to the internal pages cache, and I may sometime break that. It won't be for qpdf 9 though most likely. When I break API, I always enumerate the changes in the release notes. If I break API in a big way, like killing PointerHolder, I will most likely provide some non-default approach to aid in the transition, such as allowing QPDF to be built with a preprocessor symbol that provides a backward-compatible API or something.

qpdf 9 has a few API breakages, but all for cases where the old API simply didn't work or for dark corners that normal users wouldn't touch. Breaking API for deep power user interfaces is less problematic because those people are better equipped to change their code and because it affects fewer people.

ams-tschoening:
If ABI/API-changes are possible, then what's the reason of not accepting this PR for release 9? It surely sounds and looks more difficult than what it actually is. I only did two things:

1. Using `PointerHolder` where it wasn't used before to fix some compiler issue, which you seem to are using in most places already anyway.

2. This propagated up the call stack and slightly changed some method signatures. To make those things easier in future, additional `typedef`s have been introduced.

The main changes are the following two lines only:

    typedef PointerHolder<QPDFOutlineObjectHelper> Ptr;
    typedef std::list<QPDFOutlineObjectHelper::Ptr> List;

https://github.com/qpdf/qpdf/pull/297/commits/2fa043360b18941a51529d542a1b6f5f80892880#diff-6cebeb8348d32fdf02fb4cd2ed4da6bcR40

In my opinion, the code is easier to read and maintain now. Existing users don't need to change much as well, if at all, only some additional dereferences or such.

It would be great if you could reconsider your decision, so that I don't need to revert my changes and find another solution. Thanks!

jberkenbilt:
I don't want the extraneous typedefs. They are inconsistent with the rest of the library. However, I'll look more closely at the changes and possibly select some of them and ask you to retest.

jberkenbilt:
If the typedefs are just cosmetic, you would improve your odds by resubmitting without the List and Ptr typedefs so I can see the real changes. :-)

ams-tschoening:
Will do so.

ams-tschoening:
Removed the `typedef`s as requested. In the end, the compiler only fails at the following line. Using a `PointerHolder` there fixes the error, but leads to all the other signature changes. :-/

    std::list<QPDFOutlineObjectHelper> kids;

https://github.com/qpdf/qpdf/commit/813a37f3bafa416688cc051d7593320f53078c91#diff-6cebeb8348d32fdf02fb4cd2ed4da6bcR116

The following commit implemented another approach avoiding the signature changes by copying kids into a backwards compatible structure, but not sure if it's worth it.

https://github.com/ams-tschoening/qpdf/commit/ffa8b82e369726016113ffbc4cacd9ec285b8ca5

I guess there's something wrong with the forward declaration and including of `QPDFOutlineObjectHelper` vs. `QPDFOutlineDocumentHelper` for my compiler, but I couldn't get anything else to compile.

jberkenbilt:
I don't understand why this is necessary. The type of QPDFOutlineObjectHelper *is* fully defined, so the article you referenced doesn't seem to be relevant, and I really hate to use lists of PointerHolders when the whole notion of these helpers is they are copiable in this way. I wonder whether it works if I use vector instead of list. Have you tried that? At a glance, I don't see that it would change any of the code. Nothing is depending on any capability of list that isn't there in vector. It's possible that the queue initialization may change if queue.insert won't work with vector iterators, but that would just be making a for loop...

I'd rather see list -> vector than having to introduce another level of indirection.

ams-tschoening:
I didn't test with `std::vector` because I wouldn't have expected to make any difference... But it does, `std::vector` fixes the problem as well for some reason. So I removed usage of `PointerHolder` once more and changed to `std::vector` wherever necessary. Doing that I saw you were using vectors already at multiple places, maybe for similar reasons?

jberkenbilt:
I am good with this change. The PR fails CI because of fuzzer issues. However, I am willing to take the changes manually, fix the other issues, and update the docs. This is an API change, but it is worth doing to work around this issue. I will take it for 9.0.

jberkenbilt:
I have incorporated your changes into my work branch. If everything goes well, I will push to master this evening and manually close this issue. Thanks for your work in investigating this and making my job easier.

jberkenbilt:
Fixed in master.

ams-tschoening:
Thanks for merging!

PR #302, opened 2019-03-02
Fix C++ exception handling when -fvisibility=hidden

jbarlow83:
Ensure that QPDFExc and QPDFSystemError are marked visible, so that their typeinfo will not be
suppressed compiled with `-fvisibility=hidden`.

Details:
    https://gcc.gnu.org/wiki/Visibility

### DIFF
--- a/include/qpdf/DLL.h
+++ b/include/qpdf/DLL.h
@@ -26,7 +26,11 @@
 #if defined(_WIN32) && defined(DLL_EXPORT)
 # define QPDF_DLL __declspec(dllexport)
 #else
-# define QPDF_DLL
+  #if __GNUC__ >= 4
+    #define QPDF_DLL __attribute__ ((visibility ("default")))
+  #else
+    #define QPDF_DLL
+  #endif
 #endif
 
 #endif /* QPDF_DLL_HH */
--- a/include/qpdf/DLL.h
+++ b/include/qpdf/DLL.h
@@ -25,12 +25,13 @@
 
 #if defined(_WIN32) && defined(DLL_EXPORT)
 # define QPDF_DLL __declspec(dllexport)
+# define QPDF_DLL_EXCEPTION
+#elif __GNUC__ >= 4
+# define QPDF_DLL __attribute__ ((visibility ("default")))
+# define QPDF_DLL_EXCEPTION __attribute__ ((visibility ("default")))
 #else
-  #if __GNUC__ >= 4
-    #define QPDF_DLL __attribute__ ((visibility ("default")))
-  #else
-    #define QPDF_DLL
-  #endif
+# define QPDF_DLL
+# define QPDF_DLL_EXCEPTION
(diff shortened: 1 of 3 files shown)


jbarlow83:
For more detail `-fvisibility=hidden` suppresses the typeinfo from the linker output, and this is required for dynamic casting which exception handling relies on. It seems gcc may use some semantics to figure out if it needs the typeinfo, while clang does not, hence the macOS-only failure. Or perhaps there's some heuristic involved. 

`QPDF_DLL` should be added to the class definition (`class QPDF_DLL QPDFThing {...`) for any exposed class that might be involved in a dynamic cast. I think that's just exceptions.

This does not change the ABI. It only affects how objects that link to libqpdf.so are created. (At least I assume this is the case because I was able to test it by modifying the header files without recompiling.)

jberkenbilt:
@jbarlow83 I'll accept this change. However, I want qpdf to be fully bisectable, so I need this to be squashed into a single commit. Either you can do it so that you remain the committer, or I can do it, which will keep you as author and make me committer. Or I can try the squash and merge button, which I've never used, so I'm not sure what the commit comment will look like.

jbarlow83:
Squash merge is fine with me, whatever effect it has. IIRC it will retain me as author and make you committer.

jberkenbilt:
I wonder if I should consider building with `-fvisibility=hidden` starting in 9.0, which will the first release after the next time I decide to make a non-binary-compatible change. I'll add this to my list of things to consider. I keep a list of things to do next time I break ABI, mostly consisting of things I've done in a less-than-optimal way to avoid breaking ABI needlessly....

jbarlow83:
I'm not an expert but as I understand it, there are not many drawbacks to `-fvisibility=hidden` and it would make the exported ABI equivalent to the same set that is exported to Windows.

(It's highly recommend for pybind11 because that library uses templates heavy and generates very long mangled names.)

jberkenbilt:
It's an ABI change because things that were previously exported will no longer be exported. It won't effect anyone who only calls things through normal means because they would have had no way to call hidden functions, but in theory someone could use nm to look things up and call them manually. So for practical purposes it is not an ABI change that matters, but by the letter of the law, it is. Also it is possible that someone somewhere has written code that does RTTI on pipeline classes, so those also have to be exported in this way. I would need to make sure I fully understand the consequences and write tests to make sure things work as I did for the exception case. Then I would feel safe. Often when I'm coding I forget a QPDF_DLL here and there and catch it when the Windows builds fail. This would cause those things to get caught sooner and make qpdf more consistent across environments, so I think it's worth doing when I'm ready to break ABI. I want to fix up some issues around integer size too which also breaks ABI, and there is a list of other items in the TODO file.

jberkenbilt:
@jbarlow83 qpdf 9 will use -fvisibility=hidden. I'm testing now and should be pushing to master soon.

Issue #298, opened 2019-02-14
QPDF using wmain on Windows and paths with e.g. German umlauts.


ams-tschoening:
I tried to open a PDF file using a German umlaut and got the following error executing using `cmd.exe`:

    C:\Users\[...]>qpdf --check [...]_Bärbel_[...].pdf
    open [...]_B├ñrbel_[...].pdf: No such file or directory

`[...]` is used to strip the file name to the most important details. I compiled `qpdf` using its support for `wmain` and all of your tests passed.

Looking at the code, you are converting to UTF-8 and forwarding internally. Does that mean you are forwarding UTF-8 encoded paths to `open` as well? If so, the error is correct, UTF-8 encoded file names can't be found this way on Windows. One either needs to provide the encoding in use by the system currently or Unicode, meaning `wopen` instead of open.

https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/open-wopen?view=vs-2017

Mapping UTF-8 into the file system works on most non-Windows, because they by definition simply store individual bytes, which are most likely UTF-8 encoded. But Windows/NTFS doesn't do so, it stores UTF-16 chars.

https://stackoverflow.com/questions/2050973/what-encoding-are-filenames-in-ntfs-stored-as

The reason to switch to use `wmain` was that some of your tests are passing Unicode chars to `qpdf`, which failed without `wmain`, but succeeded with. But how you handle those arguments internally is totally up to you, while forwarding UTF-8 to `open` involves the file system and such.

sz5000:
On Windows, UTF8 filenames needs to be re-encoded to ANSI if you do not use the unicode versions of the Windows functions. Otherwise file names with umlauts will not work. Prior to `wmain` we were able to pass ansi encoded files to qpdf and this worked because qpdf did not further encoded the files and passed them to the corresponding windows api function, but now you encode wmain arguments as UTF-8 internally and this makes things worse.

jberkenbilt:
You're right. I overlooked the case of filenames containing Unicode characters. I will have to figure out the right way to handle both cases. Ugh. Thanks for the report.

sz5000:
Internally it should be fine when you do a utf8 -> ANSI encoding before you pass filenames to the open method. Otherwise you have to be selective and you have to encode to ANSI only for filenames.

ams-tschoening:
[...]

PR #322, opened 2019-04-22
Perl wrongly writes individual UTF-8 bytes in test file names.

ams-tschoening:
Those file names are needed to test Unicode compatibility for Windows and the problem can easily be fixed by creating the necessary file names manually and version them. That way they always have the correct encoding in the currently available file system, as GIT-clients should handle that properly. It's not possible to alter the name bytes in the Perl test using this commit, which is a strong indication that things are handled properly.

This PR is associated with the following comment:

https://github.com/qpdf/qpdf/issues/298#issuecomment-485468105

### DIFF
--- a/qpdf/qtest/qpdf.test
+++ b/qpdf/qtest/qpdf.test
@@ -170,7 +170,7 @@ $td->runtest("extra overlay filename",
 foreach my $d (['auto-ü', 1], ['auto-öπ', 2])
 {
     my ($u, $n) = @$d;
-    copy('minimal.pdf', "$u.pdf");
+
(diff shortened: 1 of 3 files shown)


jberkenbilt:
The reason I didn't do this was because I didn't want to rely on git getting it right. Let me consider this and see if I can do something to make sure it's right, such as adding some Windows-specific code to verify that it's right.

ams-tschoening:
#298 implemented a different approach by copying the files natively using some small additional C++-app and that works as well. So I'm closing this PR as not needed anymore.

Issue #218, opened 2018-07-12
Encryption in QPDF - usage of external library?


zdohnal:
Hi Jay,
I'm Zdenek Dohnal and I'm a maintainer of QPDF in Fedora and RHEL. I would like to ask you about encryption feature of QPDF. 
IIUC QPDF implements several encryption algorithms  (RC4-40, RC4-128, RC4-256 and AES-256) by itself. Why doesn't QPDF use a cryptographic library (f.e. gnutls) instead? 
Would you mind considering an usage of cryptographic library? I can cooperate with you, if you are interested.

mschilli87:
@zdohnal : Recently, @jberkenbilt [commented](https://github.com/qpdf/qpdf/issues/200#issuecomment-379601225) on [another issue](https://github.com/qpdf/qpdf/issues/200):

> I have purposely aimed to keep external dependencies and operating system-specific code to a minimum, and I have also stayed away from C++11 or newer features, which is appreciated by some of my users who build qpdf on embedded platforms or other environments for which more modern tools are less available.

I'm not sure how widespread gnutls is and to what extend it would limit the systems QPDF could run on, but this might have been a reason to go down the road taken so far.

zdohnal:
@mschilli87 Gnutls was the first one came on my mind, if the widespreading is critical factor, then OpenSSL can be better in the aspect.
It could have been done by a ifdef preprocessor macro (so when you want crypto from external library, compile qpdf with -D_XY).

jberkenbilt:
I think it's a reasonable request since there's inherent security risk with embedding any crypto or other potentially security sensitive logic inline. I also wear a package maintainer hat, so I fully appreciate the issue. I would also be concerned about license compatibility. qpdf uses the Apache license (and/or the Artistic license). I'm not fully up to speed on the current state, but I know there are historically issues with openssl's license being incompatible with other things, I don't know what license gnutls uses.

If I would support external libraries for this, I would want it to be enabled with a ./configure option, such as `--with-gnutls` or `--with-openssl`, and the choice of underlying implementation would be hidden behind some abstraction. I don't want `#ifdef`s litering the main code. There is already support for replacing the random number generation piece, so that might establish somewhat of a pattern. It's been a few years, so I don't remember exactly how I set that up.

[...]

PR #357, opened 2019-09-02
GNU TLS support

zdohnal:
Hi Jay,

I created a patch for #218 with usage of GNU TLS,  which is available on Linux and Windows too. I tried to avoid using ifdefs as much as I was able - I took CUPS as reference for implementing the support - they solve it by:
#ifdef gnutls
#  include "gnutls.c"
#else 
...internal implementation
#endif 

Most of those ifdefs are in files which implements hashes and ciphers, but I needed to 'ifdef' Pl_AES_PDF too, because there is internal implementation of CBC mode there.

Please tell me if there is something I should enhance, change, rewrite etc. and I'll change the code according to it.
I ran qpdf test suite with the changes and it passes.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -488,6 +488,36 @@ AC_ARG_ENABLE(show-failed-test-output,
     fi],
    [SHOW_FAILED_TEST_OUTPUT=0])
 
+AC_PATH_TOOL(PKGCONFIG, pkg-config)
+
+AC_MSG_CHECKING(if cryptography will be supplied by GNU TLS)
+AC_ARG_ENABLE(gnutls,
+   AS_HELP_STRING([--enable-gnutls],
+                  [enable cryptography by GNUTLS]),
+   [if test "$enableval" = "yes"; then
+      if test "x$PKGCONFIG" != x ; then
+        AC_TRY_COMPILE([
+        #include <gnutls/crypto.h>], , , AC_MSG_ERROR([GNU TLS development headers are needed.]))
+        GNUTLSFLAGS=`pkg-config --cflags gnutls`
+        GNUTLSLIBS=`pkg-config --libs gnutls`
+        LIBS="$LIBS $GNUTLSLIBS"
+        CPPFLAGS="$CPPFLAGS $GNUTLSFLAGS"
+        AC_DEFINE([HAVE_GNUTLS], [1], [If cryptography will be supplied by GNU TLS])
+        HAVE_GNUTLS=1
+      else
+        HAVE_GNUTLS=0
+        AC_MSG_ERROR([pkg-config is needed to configure compilation with GNU TLS.])
+      fi
+    else
+      HAVE_GNUTLS=0
+    fi],
+   [HAVE_GNUTLS=0])
+if test "$HAVE_GNUTLS" = "1" ; then
+   AC_MSG_RESULT(yes)
+else
+   AC_MSG_RESULT(no)
+fi
+
 AC_ARG_WITH(docbook-xsl,
    AS_HELP_STRING([--with-docbook-xsl=DIR],
 		  [location of docbook 4.x xml stylesheets]),
--- a/libqpdf/MD5-gnutls.cc
+++ b/libqpdf/MD5-gnutls.cc
@@ -1,0 +1,58 @@
+#include <gnutls/crypto.h>
+#include <qpdf/MD5.hh>
+#include <qpdf/QUtil.hh>
+
+
+// MD5 initialization. Begins an MD5 operation, writing a new context.
+void MD5::init()
+{
+    int ret;
+
+    ret = gnutls_hash_init(&ctx, GNUTLS_DIG_MD5);
+    if (ret < 0) {
+	QUtil::throw_system_error(
+	    std::string("GNU TLS: MD5 error:") + std::string(gnutls_strerror(ret)));
+	ctx = nullptr;
+    }
+
+    finalized = false;
+    memset(digest_val, 0, sizeof(digest_val));
+}
+
+// MD5 block update operation. Continues an MD5 message-digest
+// operation, processing another message block, and updating the
+// context.
+
+void MD5::update(unsigned char *input,
+		 size_t inputLen)
+{
+    if (ctx != nullptr && input != nullptr)
+	gnutls_hash(ctx, input, inputLen);
+}
+
+// MD5 finalization. Ends an MD5 message-digest operation, writing the
+// the message digest and zeroizing the context.
+void MD5::final()
+{
+    if (finalized)
+	return;
+
+    if (ctx != nullptr)
+	gnutls_hash_deinit(ctx, digest_val);
+
+    finalized = true;
+}
+
+// MD5 basic transformation. Transforms state based on block.
+void MD5::transform(UINT4 state[4], unsigned char block[64])
+{}
+
+// Encodes input (UINT4) into output (unsigned char). Assumes len is a
+// multiple of 4.
+void MD5::encode(unsigned char *output, UINT4 *input, size_t len)
+{}
+
+// Decodes input (unsigned char) into output (UINT4). Assumes len is a
+// multiple of 4.
+void MD5::decode(UINT4 *output, unsigned char *input, size_t len)
+{}
--- a/libqpdf/MD5.cc
+++ b/libqpdf/MD5.cc
@@ -37,6 +37,10 @@
 #include <string.h>
 #include <errno.h>
 
+#ifdef HAVE_GNUTLS
+# include "MD5-gnutls.cc"
+#else
+
 int const S11 = 7;
 int const S12 = 12;
 int const S13 = 17;
@@ -294,6 +298,8 @@ void MD5::decode(UINT4 *output, unsigned char *input, size_t len)
 	    (static_cast<UINT4>(input[j+2]) << 16) |
             (static_cast<UINT4>(input[j+3]) << 24);
 }
+
+#endif /* HAVE_GNUTLS */
 
 // Public functions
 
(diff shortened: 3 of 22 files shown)


jberkenbilt:
Thank you very much for doing this. Let's definitely set a goal to have this patch accepted.

There are a few ways we could proceed. First, I am not planning to work on qpdf for a bit...I just put in a big surge, and I need to do some other things. Later in the year I will try to do another surge. I don't want to introduce any big changes for a while because I want to have the agility to release 9.0.x bug fixes with minimal impact.

Anyway, you've done most of the heavy lifting here. There are lots of minor things here that I don't care for, such as the pattern you lifted from CUPS. It's not an approach I would use. I'd have to think it through, but I'd be more inclined to either have the build pick which files to compile or to have the file include one of two different implementations. Generally I don't like files including other source files, though qpdf already does this in at least one place. But I especially don't want such asymmetric treatment. It will be very easy for someone to look at code and not realize they're reading code that's not compiled. It will also be important for me to ensure that qpdf is regularly compiled and tested both with and without gnutls if it is my intention to support both configurations.

There may need to be updates to the oss-fuzz integration.

The code does not follow my stylistic conventions. I have not documented them, but nearly the entire codebase follows the same conventions, and I'd like to keep in that way.
* No omission of optional braces
* 4-space indent level with no tabs (older code uses tabs, and fixing it is more evil than allowing it, but I don't use tabs in new code)
* braces on separate lines
* Check configure.ac -- I already have pkg-config macros so you don't have to call pkg-config, I think.

For now, avoid c++-11 features like nullptr. I will hopefully drop this restriction soon.

Everything above is subject to negotiation. I think it's important to get this in, and I can bend a little to make it a reality. If some of the things I've mentioned turn out to be impractical or defeat the point of doing this, I'll be flexible.

What I'd like to do is to avoid having you spend a lot of time making updates only to have me reject them. We should probably iterate with more granularity.

Since you've already done basically all the heavy lifting, I'm also fine with just leaving this PR open, and when I'm ready to tackle this, I can do the plumbing bit of making sure everything follows my rules. Since you've done the actual implementation work, I would still give you credit for the changes, and I could prioritize this for the next release of qpdf since it will probably just be a few hours of my time to get everything working.

What I'll do is put the "next" tag on the associated issue. If you want to try to tackle some of the things I mentioned, great, but I think you've probably already done enough to enable me to get this in. Before you spend a lot of time making additional changes, we should communicate.

Also, if any files are renamed, I like to keep the renames in separate commits, and I require each commit to be fully working so my commits are always bisectable. E.g., if the non-gnutls implementation of something were renamed, one commit would have to be only the renames and any associated build changes to make the code find the renamed files even if a subsequent commit changed some of the same files again. This is the part where it's pretty quick for me to start with someone else's change and get it into my style, and unless you were going to be making lots of contributions, it may or may not be worth the effort to get you fully up to speed. But if you want to take on more of the work, I'm happy to provide guidance.

Thanks again for all your help on this. Your changes will probably make this a reality. Without this patch, it's unlikely I was going to get to this any time soon.

jberkenbilt:
I need to give a little more thought to the build piece of this. IN #218, there are people who have suggested other alternatives. What would feel right to me would be to do this in a manner that we could omit all but one implementation at compile time but we could also build with all options simultaneously enabled, and it should be structured in a manner that would make it easy to replace gnutls with something else later if we wanted to. This would require some refactoring of the build and code, which I would probably want to take on myself, but it wouldn't be a huge amount of work.

In my day job, I have seen immense technical debt accumulate as a result of having large blocks of code enabled and disabled at compile time. It quickly becomes impossible to ever be confident that all supported configurations are still working. It may be that CI will build all allowed crypto providers and independently test them with libtests but then select one for the main qpdf, and OS packages will disable all expect gnutls, for example.

I'll give this some thought when I'm ready to take it on.

zdohnal:
> Thank you very much for doing this. Let's definitely set a goal to have this patch accepted.
You're welcome - it was really something new for me :) .
> 
> Anyway, you've done most of the heavy lifting here. There are lots of minor things here that I don't care for, such as the pattern you lifted from CUPS. It's not an approach I would use. I'd have to think it through, but I'd be more inclined to either have the build pick which files to compile or to have the file include one of two different implementations. 

Do you mean like f.e. wget does? If they run configure script with specific crypto library, they then set different object files and different source files to compile.
I was trying that approach, but I would need to fiddle in Makefile.am, so I rather chose CUPS way.

> 
> The code does not follow my stylistic conventions. I have not documented them, but nearly the entire codebase follows the same conventions, and I'd like to keep in that way.

I tried to follow the code style of original files, which I omitted by include - means if I was writing sha2-gnutls.c, I followed sha2.c code style.

> 
>     * No omission of optional braces
> 
>     * 4-space indent level with no tabs (older code uses tabs, and fixing it is more evil than allowing it, but I don't use tabs in new code)
> 
>     * braces on separate lines
> 
>     * Check configure.ac -- I already have pkg-config macros so you don't have to call pkg-config, I think.

Ok, I'll update the code accordingly. Probably Pl_AES_PDF.cc is good example how it should look like.

> 
> 
> For now, avoid c++-11 features like nullptr. I will hopefully drop this restriction soon.

Ok, I'll go for NULL for now.

> 
> Everything above is subject to negotiation. I think it's important to get this in, and I can bend a little to make it a reality. If some of the things I've mentioned turn out to be impractical or defeat the point of doing this, I'll be flexible.
> 
> What I'd like to do is to avoid having you spend a lot of time making updates only to have me reject them. We should probably iterate with more granularity.

Ok, can we then agree on what I can do to help more with the issue? I can correct code style f.e.

> 
> Since you've already done basically all the heavy lifting, I'm also fine with just leaving this PR open, and when I'm ready to tackle this, I can do the plumbing bit of making sure everything follows my rules. Since you've done the actual implementation work, I would still give you credit for the changes, and I could prioritize this for the next release of qpdf since it will probably just be a few hours of my time to get everything working.

Ok.

> 
> What I'll do is put the "next" tag on the associated issue. If you want to try to tackle some of the things I mentioned, great, but I think you've probably already done enough to enable me to get this in. Before you spend a lot of time making additional changes, we should communicate.
> 
> Also, if any files are renamed, I like to keep the renames in separate commits, and I require each commit to be fully working so my commits are always bisectable. E.g., if the non-gnutls implementation of something were renamed, one commit would have to be only the renames and any associated build changes to make the code find the renamed files even if a subsequent commit changed some of the same files again. This is the part where it's pretty quick for me to start with someone else's change and get it into my style, and unless you were going to be making lots of contributions, it may or may not be worth the effort to get you fully up to speed. But if you want to take on more of the work, I'm happy to provide guidance.
> 

Noted :) IIRC the patch did not rename any current file, just added the new ones, but it is always good to know.

> Thanks again for all your help on this. Your changes will probably make this a reality. Without this patch, it's unlikely I was going to get to this any time soon.

Happy to help :)

zdohnal:
> I need to give a little more thought to the build piece of this. IN #218, there are people who have suggested other alternatives.

Actually I chose gnutls for other reason, that it is already used in CUPS, so it kind made sense to me that two projects which works with PDF could have the same crypto library...
Licensing and availability (GNUTLS is for Win and Linux too) are not problems.
 
> What would feel right to me would be to do this in a manner that we could omit all but one implementation at compile time but we could also build with all options simultaneously enabled, 

Hmm... I'm not sure I folllow - the first part is clear - we can choose one implementation and others will be disabled, but have them all enabled? Like build f.e. 3 different libqpdf at one time? Each one with different external library?

> and it should be structured in a manner that would make it easy to replace gnutls with something else later if we wanted to. This would require some refactoring of the build and code, which I would probably want to take on myself, but it wouldn't be a huge amount of work.

Ok, since I'm not quite sure what you mean, I agree.

> 
> In my day job, I have seen immense technical debt accumulate as a result of having large blocks of code enabled and disabled at compile time. It quickly becomes impossible to ever be confident that all supported configurations are still working. It may be that CI will build all allowed crypto providers and independently test them with libtests but then select one for the main qpdf, and OS packages will disable all expect gnutls, for example.
> 
> I'll give this some thought when I'm ready to take it on.

Ok.

jberkenbilt:
Just letting you know I have started to look at this more carefully. I'm going to try to spend a little time seeing if I can get this in for 9.1. If not, then it will go in for the next minor version bump. It is the next feature I intend to work on.

jberkenbilt:
@zdohnal I am leaving this PR open so that I can harvest code from it. I will structure the commits in my final version so that you are the author of the code that you contributed. I'm going to chip away at this in my available time, but it is the next feature I am working on. Most likely this will be released in 9.2. I have a feature waiting to go for 9.1. I am also planning on requiring C++-11 for whatever release includes this enhancement.

zdohnal:
Ok, no problem. Thank you for letting me know, agreeing to add the feature and for your work on it!

If you need anything, just let me know.

jberkenbilt:
I have incorporated all your code into my crypto branch, which I am still rebasing and finishing up, but the changes are basically in

I notice that the md5 code leaks memory from inside gnutls. I will investigate, but it's probably a gnutls bug. Linearization uses md5.

jberkenbilt:
In case you're interested: https://github.com/jberkenbilt/qpdf/tree/crypto. The commit "Implement gnutls crypto provider" is the one with this code in it.

jberkenbilt:
found the memory leak.

jberkenbilt:
I have completed the implementation and have a little more doc work to do before merging. I have also been able to do this without breaking binary compatibility, which means that this could in principle go in to qpdf 9.1.0. I have taken code from this PR, but in the end, I ended up having to rewrite a good chunk of it because of minor bugs or stylistic considerations, and after refactoring to allow swapping out the crypto provider, it ended up being just one source file containing all the gnutls-specific code.

The code you provided was extremely useful and was essential to my being able to implement this quickly. I have credited you in the commit comment and will also credit you in the ChangeLog and release notes.

I am going to close this PR now and put additional updates in #218. Thanks again for your contribution!

zdohnal:
Thank you for making this possible, Jay!

PR #367, opened 2019-09-26
Fix /Contents value representation in a signature dictionary

trueroad:
Table 8.93 "Entries in a signature dictionary"
in PDF 1.5 reference describes that
the value of Contents entry is a hexadecimal string representation
when ByteRange is specified.

This commit makes
QPDF always uses hexadecimal strings representation
instead of literal strings for it.


jberkenbilt:
Looks good at first glace. It will need tests, but I can add the tests. Do you have a real example that can be shared without copyright or license concerns?

trueroad:
[input-signed.pdf](https://github.com/qpdf/qpdf/files/3662703/input-signed.pdf)
How about this file?

trueroad:
I've tried to add the tests.
Would you review it?

jberkenbilt:
As with my other comment, these tests should go somewhere in qpdf.test. For the most part, when possible, I use the qpdf CLI to exercise specific cases by creating input and output files and testing them. You can see hundreds of examples of this in qpdf.test.

I fully recognize that this style of testing is different from what most people teach these days of using unit tests for most things and using integration tests for things that can't be tested with unit tests. I deviate from this practice for several reasons. The main one is that I set up the whole testing framework used by qpdf now something like 17 years ago and was following what I perceived to be practices of the day, and now I like to keep things consistent. But also I think it's a very effective way to test command line tools. Exercising library functionality through the CLI offers several major advantages: 1) it ensures that the library functionality is accessible through the CLI, 2) it ensures that real files work correctly when used the way users use them, 3) it allows for all tests of libqpdf's core functionality to be together. Secondly, there is no overhead to integration testing in this case. It's not like this is a web application where I have stand up a servlet container or something.

As I mentioned in a different commit, I use test_driver.cc for things in the core classes where I want more of a unit test style or where I want to exercise cases that are not easily reachable through the CLI.

In the case of this change, it should be easy enough to just process files with the qpdf CLI and ensure that they look right. Again, I can help with reorganizing the tests.

jberkenbilt:
I've looked through all the tests at a glance. I think you're doing a great job of getting everything in shape here. Once you have all the tests written, whichever way you need to do it either by adding tests that depend on earlier changes to the later PRs, rebasing PRs on top of other PRs, combining into a single PR, or whatever, then I will do a deep analysis and help to get commits in order.

One option would be that I can create an integration branch and we can merge all the PRs into the integration branch. Then once you're satisfied that all your changes are in and tested, I can work in the integration branch to pull it into my standard and do any additional checks, and then we can merge that into master. Let me know how you want to do it.

I may try to get a 9.0.2 out with one fix. Then this set of changes will make up 9.1. Depending on how much time I set aside, maybe I will also put the gnutls changes in for 9.1, or maybe that will end up being 9.2. Thanks for all your work on this. Your changes will make qpdf a better library.

trueroad:

jberkenbilt:
Incorporated by #376

PR #368, opened 2019-09-27
Exclude signature dictionary from compressible objects

trueroad:
It seems better not to compress signature dictionaries.
Various PDF digitally signing tools, including Adobe Acrobat Reader DC,
do not compress signature dictionaries.

Table 8.93 "Entries in a signature dictionary" in PDF 1.5 reference
describes that /ByteRange in the signature dictionary
shall be used to describe a digest
that does not include the signature value (/Contents) itself.

The byte ranges cannot be determined if the dictionary is compressed.


trueroad:
I've fixed the commit.

Thank you for your advice.

trueroad:
For the tests for this pull request, I would like to add the condition `--object-streams=generate` to `sig_dict_contents.test` in #367.
If #367 is merged, I will write the tests.

jberkenbilt:
I'm hoping to merge these all at once in a single PR, so it would be great if you can write all the tests before I merge anything. If needed, you can rebase this PR on top of #367 and then add the updated tests in a separate commit. In the end, these tests will move out of libtests anyway.

trueroad:

jberkenbilt:
Incorporated by #376

PR #369, opened 2019-09-28
Add QPDF::getXRefTable()

trueroad:
I would like to get cross reference table/stream information of a PDF.
This pull request add QPDF::getXRefTable() which can get the cross reference table/stream.


include/qpdf/QPDF.hh
```diff
@@ -239,6 +239,8 @@ class QPDF
     QPDFObjectHandle getTrailer();
     QPDF_DLL
     QPDFObjectHandle getRoot();
+    QPDF_DLL
+    std::map<QPDFObjGen, QPDFXRefEntry> const& getXRefTable() const;
```
jberkenbilt: This has to be a non-const method that returns a copy, and it has to make sure it has been called after parse() so that the xref table is valid. I'm not adding any methods that return references to private members. I would probably accept the addition of this method as `std::map<QPDFObjGen, QPDFXrefEntry> getXRefTable()`.

jberkenbilt:
note to self: if accepting this change, add comments explaining limitations and reference from showXRefTable.

jberkenbilt:
@trueroad  I'm curious about your use case. Also, you have supplied a handful of pull requests, which is great. If you can give me a sense of when you're done making the changes you think you need, it will help me time a release. If there is something specific you're trying to accomplish, I'm willing to discuss to see if the direction you're going with qpdf seems consistent with how it's intended to be used or if there are other things I should do to the library in support of your use case.

include/qpdf/QPDF.hh
trueroad: I've fixed the commit.
Thank you.

trueroad:
Thank you for your advice.

I would like to make a PDF digitally signing tool.
It works in two phases.

In the first phase, it generates an intermediate PDF in preparation for signing.
First, it modifies and adds some dictionaries including a signature dictionary with dummy values of /ByteRange and /Contents.
Then, by QPDFWriter::write(), it creates the intermediate PDF with the dummy values.
In the intermediate PDF, the signature directory has to be uncompressed (#368) and unencrypted, and the values of /Contents has to be hexadicimal representation (#367).

In the second phase, it rewrite the intermediate PDF to obtain the final digitally signed PDF.
First, it finds file offsets of the dummy values in the intermediate PDF (#369).
From the file offset of the dummy /Content value, it calculates the correct /ByteRange value and rewrites the dummy /ByteRange value to the correct value.
It then calculates the digital signature digest and rewrite the dummy /Contents value to the correct /Contents value.
The rewritten PDF is the final digitally signed PDF.

The rewrite must not change all bytes other than dummy values in the intermediate PDF.
Moreover, it must not change the file offset and size of all items.
If I understand correctly, libqpdf does not support such rewriting.
Therefore, in this phase, libqpdf is used to find the file offset of the dummy values, but normal file I/O is used for rewriting.

QPDF::getXRefTable() can obtain the file offset of the signature directory in the intermediate PDF.
If I understand correctly, QPDF does not have an offset for each object in the dictionary.
So to get the file offsets of the dummy values, my tool has to parse the dictionary without libqpdf.

jberkenbilt:
My inclination is to approve your pull requests because they enable a certain style of working with PDF files that could be useful. For example, one could imagine writing software that discovers unreferenced parts of the PDF file, something that would not be possible with the existing API.

Your use case is interesting. It would be good if qpdf directly handled digital signing. Is this digital signing as described by the spec? This feature has been requested before, but I haven't had time to think about doing it, and I believe it would require additional external dependencies. If you look at the way linearization is implemented in QPDFWriter, you can see that the file is written in two passes with padding and so forth. Sounds like you might need a similar thing. It might be possible to make QPDFWriter support this in a general way so that your implementation could let qpdf do more of the bookkeeping, but I haven't given it much thought.

Is the thing you're working on going to be open source? (Okay either way, just wondering.)

trueroad:
Thank you for your reviewing.

If I understand correctly, PDF digital signatue spec is
"12.7.4.5 Signature Fields" and "12.8 Digital Signatures"
in PDF 32000-1:2008 (PDF 1.7 spec).

Of course, my tool will require OpenSSL for digitally signing.
I will make it open source.

jberkenbilt:
There is a pull request to add external crypto to qpdf using gnutls. I plan on accepting the bulk of the implementation, which will add gnutls as a dependency. Can gnutls be used instead of openssl? Would you be open to a deeper integration with qpdf? I was thinking of implementing digital signature myself once gnutls was added as a dependency. Maybe your work could be a starting point. There are obviously things that could be leveraged if digital signature is implemented inside qpdf that would make it easier and better than implementing it outside the library.

trueroad:
I'm not familiar with GnuTLS.
But I think that GnuTLS can be used instead of OpenSSL.

trueroad:
I've tried to add the tests.
Would you review it?

jberkenbilt:
Thank you. I will review it. It may be a bit of time. Are you done making changes other than in response to my review?

trueroad:
> Are you done making changes other than in response to my review?

No. I'm preparing tests.

jberkenbilt:
What would be ideal is if you could put all your changes in one pull request, separating into a separate commit for each change. In other words, each of your current pull requests could be a separate commit. It's fine to leave all your commits separate right up until the end, but before merge, I will want each change to be isolated along with its tests in a separate commit per change, and I will want tests to pass on each commit. However, if it's too much trouble, you can leave things separate, and I can reorganize the commits. I like to do it this way because I like the commits to be bisectable (i.e., tests should pass at each intermediate step), and I like discrete changes separated because it's much easier to figure out what changes were made for what reasons. For example, if a bug is introduced, I can see the narrowest possible context around what was happening at the time the change was introduced. I do this for my own changes as well.

I'll look at some of your tests and see if you're on the right track.

jberkenbilt:
I use libtests for testing parts of the library other than QPDF itself with rare exception. There are two reasons. The deep reason is that I use coverage cases (QTC::TC(...)) throughout the codebase to ensure that certain conditions are tested, and the scope of QTC is limited to a single test suite. The second reason is that, when I'm locally iterating on changes to qpdf and not to the helper classes, I run the qpdf test suite only to save time. I don't want library tests to be commingled with qpdf tests or qpdf tests to be isolated from qpdf.test.

The test case here looks good. Instead of creating a separate executable and putting it in libtests, you can add another case to test_driver.cc. In some cases, you could take a stand-alone executable and stick it in the qpdf directory and invoke it from qpdf.test.

All that said, I'm perfectly fine if you want to leave this the way it is, and I can just review the tests for correctness. Then when I'm ready to accept the changes, I can reorganize them according to my preferences. I would still give you credit for the work.

It's kind of up to your preference. It is my full intention to accept these changes one way or another. It will take me obviously much less than time to organize the changes according to my standards because I haven't really documented those standards as hardly anyone has ever made meaningful contributions at this scale, so the process of getting things to my preference will be an iterative one in which I am telling you to replace one perfectly good way of doing things with a different way. If you were going to be making long-term contributions to qpdf, then it would be completely worth the effort, but otherwise, it might feel like unnecessary tedium on your part.

What do you think? I'll review the other tests for correctness and disregard concerns about whether they are in the right place. When you have completed all the changes you intend to make, then I can put some effort into getting them in mergeable shape, which should be low effort because you're already doing such a thorough job. :-)

trueroad:

trueroad:
My experiment PDF digitally signing tool is here.
https://gist.github.com/trueroad/0b0a2127aff508caf583265fbef4b644

jberkenbilt:
Incorporated by #376

PR #370, opened 2019-10-01
Add QPDFWriter::getRenumberedObjGen() and QPDFWriter::getWrittenXRefTable()

trueroad:
These methods are useful for making PDF digitally signing tool (https://github.com/qpdf/qpdf/pull/369#issuecomment-536241488).


jberkenbilt:
I would be nice if the format of the xref table returned were the same as the one from QPDF. Unfortunately, QPDFWriter uses int as the map key, which is an optimization stemming from the fact that QPDFWriter never creates objects with generation != 0. I wonder whether we could construct a map from QPDFObjGen for consistency. It would make for a cleaner API. I'm writing this comment without looking at the other PR. Hopefully what I'm saying is correct and makes sense.

trueroad:
Thank you for your advice.
I've fixed it.

trueroad:
I've added the tests for QPDFWriter::getRenumberedObjGen().
For the tests for QPDFWriter::getWrittenXRefTable(), I think that QPDF::getXRefTable() in #369 is needed.
If #369 is merged, I will write the tests.

jberkenbilt:
As mentioned earlier, I want to merge everything together. If needed, rebase 370 on top of 369 so you can write the tests. This is the reason that I'd rather see everything in one PR with a separate commit for each change.

trueroad:

jberkenbilt:
Incorporated by #376

PR #372, opened 2019-10-02
Add QPDFObjectHandle::getParsedOffset()

trueroad:
This pull request makes an application using libqpdf can obtain the offset of an object in the PDF.
The method is useful for making PDF digitally signing tool (https://github.com/qpdf/qpdf/pull/369#issuecomment-536241488).


libqpdf/QPDFObjectHandle.cc
```diff
@@ -2080,6 +2081,7 @@ QPDFObjectHandle::parseInternal(PointerHolder<InputSource> input,
                 object = QPDFObjectHandle(new QPDF_Array(olist));
                 setObjectDescriptionFromInput(
                     object, context, object_description, input, offset);
+                object.setParsedOffset(offset - 1);
```
jberkenbilt: I'm not sure this handles whitespace properly.

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2159,6 +2161,7 @@ QPDFObjectHandle::parseInternal(PointerHolder<InputSource> input,
                 object = newDictionary(dict);
                 setObjectDescriptionFromInput(
                     object, context, object_description, input, offset);
+                object.setParsedOffset(offset - 2);
```
jberkenbilt: I'm not sure this handles whitespace properly.

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2175,9 +2178,26 @@ QPDFObjectHandle::parseInternal(PointerHolder<InputSource> input,
...
+QPDFObjectHandle::getParsedOffset()
+{
+    dereference();
+    return this->m->obj->getParsedOffset();
+}
+
+void
+QPDFObjectHandle::setParsedOffset(qpdf_offset_t offset)
+{
+    if (this->m->obj.getPointer ())
```
jberkenbilt: Please omit the space before `()`

libqpdf/QPDFObjectHandle.cc
trueroad: If I understand correctly, whitespace is not handled here.
`offset` points to the next character of the beginning of the array "[".
By `offset - 1`, the parsed offset is changed to point to the "[".
If you prefer the parsed offset to point the next character of the "[", I'll fix so.

libqpdf/QPDFObjectHandle.cc
trueroad: If I understand correctly, whitespace is not handled here.
`offset` points to the next character of the beginning of the dictionary "<<".
By `offset - 2`, the parsed offset is changed to point to the "<<".
If you prefer the parsed offset to point the next character of the "<<", I'll fix so.

libqpdf/QPDFObjectHandle.cc
trueroad: I've fixed.

trueroad:
Thank you for your reviewing.

I've also added parsed offset setting for stream.

I'd like the idea of how you do tests.
I'll try to provide tests for some cases.

trueroad:

jberkenbilt:
Incorporated by #376

PR #375, opened 2019-10-18
Signature dictionary /Contents encryption

trueroad:
I've noticed that signature dictionary /Contents value is not encrypted even if the PDF is encrypted.
Adobe Acrobat Reader DC handles such PDFs properly.
Conversely, it cannot handle properly an encrypted PDF with the signature dictionary /Contents value encrypted.

This pull request makes QPDF handle the signature dictionary /Contents value as unencrypted even if the PDF is encrypted.
Also tests for this is added.

Please review this pull request.


trueroad:
I've noticed that this pull request needs to change QPDFObjectHandle::getParsedOffset().
So I've merged the branch and change it.

If you do not want the two branches to merge with merge commit, but rather want the commits to be linear, make such an integrated branch / pull request.

jberkenbilt:
I am reviewing your changes and preparing pull request that I will be prepared to merge after your final okay that it meets your needs. Your changes are high quality, which makes my work easy. I'll let you know when I'm ready. Kids are playing at a neighbor's house, so I have some time. :-)

jberkenbilt:
Incorporated by #376

PR #376, opened 2019-10-20
Sig dict

jberkenbilt:
@trueroad Here is a pull request that incorporates all your changes. Here's what I did:
* Create a new branch
* Merge all your pull requests in order
* Combine each pull request into a single commit per pull request, using your text as the commit messages with some minor corrections and reformatting
* Amend commits to fix whitespace and indentation in a few spots
* Add my corrections (all cosmetic) in fixup! commits so you could clearly see them

The result is all your changes as I would merge them, except there are a few minor problems that I'm sure you'll be able to address. I am commenting on those sections in this pull request.

What I'd like you to do is the following:
* [x] See if you are okay with the changes I made in the fixup! commits. If so, I will squash those, which will mean that your name will show up on those lines of code when git blame is run. If you prefer to have my changes to your changes separate, I will not squash them but will instead change the commit messages to indicate that they are my changes.
* [x] Address the things I identify.

I think the easiest way to proceed would be for you to address my suggested changes by creating a new pull request targeted to this branch (the sig-dict branch in qpdf/qpdf). Alternatively, you can create a branch in your fork off the tip of sig-dict and just add commits to it in your own pull request, whichever you prefer.

Once I am comfortable with the changes, I will fix up the commits so that there are no longer fixup! commits by either squashing them or creating separate commits per your preference. Once you and I have both signed off, I will merge to master.

Great work.


libqpdf/QPDFObjectHandle.cc
```diff
@@ -1912,7 +1924,19 @@ QPDFObjectHandle::parseInternal(PointerHolder<InputSource> input,
 	    break;
 
 	  case QPDFTokenizer::tt_name:
-	    object = newName(token.getValue());
+	    {
+		std::string name = token.getValue();
+		object = newName(name);
+
+		if (name == "/Contents")
```
jberkenbilt: I'm feeling like this is going to prevent decryption of all /Contents dictionary entries whose values are string, when I think we only want to prevent decryption of /Contents in signature dictionaries. I feel like maybe there is a better way to handle this case. I'm happy to discuss further. There is a lot of extra bookkeeping added to handle this case. It would be nice to have a test case with a map with a /Contents value of type string that is not in a signature dictionary to make sure it is handled properly. I'm not sure whether this would actually appear in a real file, but weird things can be found in real files, and it would be a shame for qpdf to be leaving some strings encrypted because it falsely assumes they are in signature dictionaries. That would be a serious bug because it would result in loss of data.

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2080,6 +2112,10 @@ QPDFObjectHandle::parseInternal(PointerHolder<InputSource> input,
                 object = QPDFObjectHandle(new QPDF_Array(olist));
                 setObjectDescriptionFromInput(
                     object, context, object_description, input, offset);
+                // The `offset` points to the next of "[".
+                // Set the rewind offset to point to the beginning of "[".
+                object.setParsedOffset(offset - 1);
```
jberkenbilt: After reading a token, InputSource's getLastOffset() method always returns the offset of the beginning of the token. I wonder whether this could be used safely to create a cleaner way of handling this than "knowing" that you are just after the "[". When I see this code, I worry about whether the right thing happens if there is whitespace before or after the "[". If you can't use getLastOffset here because of some other reason in the code that I am forgetting, then I would want to double check that there is a test with different kinds of space around the brackets. Most likely there is already, but we should double check and make sure the comment here is specific about whitespace being handled propery.

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2156,9 +2192,25 @@ QPDFObjectHandle::parseInternal(PointerHolder<InputSource> input,
...
+		    dict["/Contents"]
+		      = QPDFObjectHandle::newString(contents_string);
+		    dict["/Contents"].setParsedOffset(contents_offset);
+		}
                 object = newDictionary(dict);
                 setObjectDescriptionFromInput(
                     object, context, object_description, input, offset);
+                // The `offset` points to the next of "<<".
+                // Set the rewind offset to point to the beginning of "<<".
+                object.setParsedOffset(offset - 2);
```
jberkenbilt: Same comment applies here as in the array case.

qpdf/test_parsedoffset.cc
```diff
@@ -0,0 +1,207 @@
...
+    {
+        ss << "direct, ";
+    }
+
+    ss << obj.getTypeName();
+
+    return ss.str();
+}
+
+void walk(size_t stream_number, QPDFObjectHandle obj,
```
jberkenbilt: This method does not recurse into dictionaries and arrays. As far as I can see, the correctness of offsets inside nested dictionaries and arrays is not tested. Also, it doesn't look like this traverses into content streams, which also contain objects.

jberkenbilt:
@trueroad I added a commit to the end to move the new tests before large file tests. I like the large file tests to be at the end because they are extremely slow when run on actual large files. Before merging, I will move them around so they appear in their new locations initially.

jberkenbilt:
@trueroad Also, if you are curious, the sig-dict-clean-merge branch in my fork (https://github.com/jberkenbilt/qpdf/tree/sig-dict-clean-merge) contains the result of merging in your pull requests without making any changes other than fixing a merge conflict.

libqpdf/QPDFObjectHandle.cc
trueroad: I've added tests for non signature dictionary /Contents string (text string) in #377.

If /Contents is not in signature dictionary, the string is not excluded encryption and decryption.
The exclusion is only for /Contents string in signature dictionary.
https://github.com/qpdf/qpdf/blob/8ce344792b40422b3ac36dac658311179ca7beca/libqpdf/QPDFObjectHandle.cc#L2195

libqpdf/QPDFObjectHandle.cc
trueroad: > After reading a token, InputSource's getLastOffset() method always returns the offset of the beginning of the token.

Unfortunately, `input->getLastOffset()` points the offset of "]" that is closing array instead of opening array.

> When I see this code, I worry about whether the right thing happens if there is whitespace before or after the "[".

`qpdf/qtest/qpdf/minimal.pdf` used in the tests has several whitespace pattern before and after the "[".

* both before and after the "["

```
  /Kids [
    3 0 R
  ]
```

* only before the "["

```
  /MediaBox [0 0 612 792]
```

* only after the "["

```
5 0 obj
[
  /PDF
```

These are working correctly.

libqpdf/QPDFObjectHandle.cc
trueroad: Also `input->getLastOffset()` points the offset of ">>" that is closing dictionary instead of opening dictionary.
`qpdf/qtest/qpdf/minimal.pdf` used in the tests has several whitespace pattern before and after the "<<".

qpdf/test_parsedoffset.cc
trueroad: > This method does not recurse into dictionaries and arrays. As far as I can see, the correctness of offsets inside nested dictionaries and arrays is not tested.

`QPDF::GetAllObjects()` gets all indirect objects.
So we only need to recurese the direct objects.
For dictionary, it worked correctly.
However, for array, it had a bug.
I've fixed it in #377.

> Also, it doesn't look like this traverses into content streams, which also contain objects.

Again, `QPDF::GetAllObjects()` gets all indirect objects.
Also the objects in content streams are got.
Thus, it traverses all objects including the objects which are in content streams.

The test file `qpdf/qtest/qpdf/digitally-signed.pdf` has some objects which are in content streams.
The output of this test `qpdf/qtest/qpdf/digitally-signed-parsedoffset.out` contains the objects which are in the content streams.

trueroad:
@jberkenbilt Thank you for your reviewing and fixing up.
I've added some commits for addressing your comments in #377.

libqpdf/QPDFObjectHandle.cc
jberkenbilt: Great, thanks for the confirmation.

jberkenbilt:
@trueroad I'm satisfied with the content of this now. Once CI passes, I will go through, update the ChangeLog and docs, squash commits, and ask for your final okay that this is what you want merged. Then I will merge. Obviously future corrections/enhancements are always possible.

jberkenbilt:
For historical record, this PR incorporates #367, #368, #369, #370, and #372.

trueroad:
Thank you for adding comments for getParsedOffset().
Please squash the commits.

jberkenbilt:
@trueroad Assuming CI is green (which I'm sure it will be), this is in the final form that I'm ready to merge. I added a commit that updates the ChangeLog and release notes in the manual, but otherwise, this is identical in content to what was here before...I just cleaned up the commits. If it looks good to you, I will merge. Please take a look at the last commit and see whether I represented your changes correctly.

After this, we can close the other pull requests. Depending on whether I am able to make time to work on the gnutls enhancement soon, I will either hold this for those changes or push out a 9.1.0 release. Either way, it will probably be a few weeks because I want to give 9.0.2 a little chance at life. :-)

Thank you for an extremely high quality submission. You did a great job dealing with all my requests, significant and nit-picky alike.

trueroad:
I've confirmed all commits.
Please merge them.

I'm looking forward to the 9.1.0 release.
Thank you for the review and helpful comments.

jberkenbilt:
For the record, the CI failure appears to be some kind of infrastructure issue with AppImage and is not related to the changes in this request. I will contact the AppImage people if the problem persists. All meaningful checks passed, and a fully passing set of checks appeared for an earlier version of this that had identical contents.

jberkenbilt:
@trueroad Thanks again for your contribution!

PR #377, opened 2019-10-21
For sig-dict branch

trueroad:
This pull request is fix for #376.


trueroad:
> Are you okay with my squashing my fixup commits and losing the distinction of your original code vs. my fixes with the result being attributed to you?

I'm okay. Please squash.
Thank you for your reviewing.

PR #385, opened 2019-11-24
Update Travis to Xenial

gliptak:
https://docs.travis-ci.com/user/reference/trusty
Trusty is EOL by Canonical


gliptak:
Travis no longer in-use?

jberkenbilt:
I'm not really using travis and never check it, but as long as I haven't deleted it, I may as well take this. Thanks.

PR #387, opened 2019-11-29
This is a true fix for 285 and 286

Cloudmersive:
The issues 285 and 286 were never truly fixed.  The root cause was that it was creating a user key instead of a machine key.  This fix also improves the error messages.

Hope this helps.


jberkenbilt:
Thanks for this! I will merge it next time I work on this. I may reformat the strings a little, but otherwise, it looks great, and fixing Windows-specific problems like this likes outside my expertise.

Cloudmersive:
Awesome, glad to hear it.  Great work on this project by the way.

PR #403, opened 2020-02-21
CIFuzz Dog Food

Leo-Neat:
OSS-Fuzz was wondering if qpdf would dogfood CIFuzz, a new feature being rolled out in OSS-Fuzz. It allows for pull requests to be fuzzed in CI using fuzz targets enabled in OSS-Fuzz. It will be active in dry_run mode meaning that it will never fail. Documentation for this feature can be found [here](https://google.github.io/oss-fuzz/getting-started/continuous-integration/). Let us know if you have any questions!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,23 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'qpdf'
+        dry-run: true
+    - name: Run Fuzzers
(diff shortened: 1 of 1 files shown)


jberkenbilt:
Yes, happy to help out in this way. I'm not sure why tests are failing right now. Looks like it might be a transient issue. I also want to make sure that, with both github actions and azure pipelines, it's obvious what's what, so I may want to tweak the workflows file. Do I have to do anything to enable actions?

Leo-Neat:
@jberkenbilt Sounds great! You don't have to do anything else to enable actions.  Once the  main.yml file is added to your  workflow directory, the test will be run  on pull requests automatically.

jberkenbilt:
CI is looking good. I'll check it later tonight or over the weekend and merge if good. I guess we hit some hiccup on Azure pipelines yesterday.

jberkenbilt:
@Leo-Neat Right now, #283 is running this action. This is my standing pull request that I update to get CI before I push to master.

PR #405, opened 2020-02-29
libqpdf/QUtil.cc: fix build without wchar

ffontaine:
Build on Linux toolchains without wchar fails since version 9.1.1 and a44b5a34a07b9f2905d419d5571fd53832c1f6c0 on:

```
libqpdf/QUtil.cc: In function 'int QUtil::call_main_from_wmain(int, wchar_t**, std::function<int(int, char**)>)':
libqpdf/QUtil.cc:2378:32: error: 'wcslen' was not declared in this scope
         for (size_t j = 0; j < wcslen(argv[i]); ++j)
                                ^~~~~~
```
To fix this error, return -1 if `WINDOWS_WMAIN` is not defined

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/libqpdf/QUtil.cc
+++ b/libqpdf/QUtil.cc
@@ -2366,6 +2366,7 @@ QUtil::possible_repaired_encodings(std::string supplied)
 int
 QUtil::call_main_from_wmain(int argc, wchar_t* argv[], std::function<int(int, char*[])> realmain)
 {
+#ifdef WINDOWS_WMAIN
     // argv contains UTF-16-encoded strings with a 16-bit wchar_t.
     // Convert this to UTF-8-encoded strings for compatibility with
     // other systems. That way the rest of qpdf.cc can just act like
@@ -2396,4 +2397,7 @@ QUtil::call_main_from_wmain(int argc, wchar_t* argv[], std::function<int(int, ch
     argc = QIntC::to_int(utf8_argv.size());
     new_argv[argc] = 0;
     return realmain(argc, new_argv);
+#else
+    return -1;
+#endif
 }


jberkenbilt:
WINDOWS_WMAIN will never be defined in QUtil.cc. Most likely there is just a missing header file. I will look into it.

jberkenbilt:
I need to include <cwchar> If I haven't already done so.

ffontaine:
Commit bb6768b does not fix the issue as the toolchain does not have wchar, build now fails on:

```
libqpdf/QUtil.cc: In function ‘int QUtil::call_main_from_wmain(int, wchar_t**, std::function<int(int, char**)>)’:
libqpdf/QUtil.cc:2379:37: error: ‘wcslen’ is not a member of ‘std’
         for (size_t j = 0; j < std::wcslen(argv[i]); ++j)
                                     ^~~~~~
libqpdf/QUtil.cc:2379:37: note: suggested alternative: ‘scalbn’
         for (size_t j = 0; j < std::wcslen(argv[i]); ++j)
                                     ^~~~~~
                                     scalbn
```

So I think there is three options:
 - update libtests/build.mk to use WINDOWS_WMAIN_COMPILE
 - add an option in the configure script to allow the user to disable libtests
 - do nothing (in this case, I'll add a wchar dependency for the qpdf package in buildroot)

jberkenbilt:
What toolchain doesn't have wchar? That is part of the standard and has been for a long time.

Updating libtests/build.mk to use WINDOWS_WMAIN_COMPILE is out of the question. It's not a library issue, it's an issue of whether you're using it in main. The library will support this functionality unconditionally as I have a hard line against compile-time changes to the API. Likewise, I'm not going to add a configure option to disable libtests since being able to build and run these tests is essential to ensuring qpdf works.

What would you do to add a wchar dependency? What platform or toolchain are you compiling for? I'm really surprised to encounter a toolchain that doesn't have wchar. Are you sure there isn't some other header that's defining some symbol that's messing things up?

I do want to come up with a suitable fix for this, but it has to be a fix that doesn't add compile-time options that change the API of the library or disable tests that are essential to verifying proper fucntionality.

ffontaine:
uclibc/uclibc-ng (https://www.uclibc-ng.org) can be built without wchar.

I opened this PR due to build failures on buildroot autobuilders with qpdf 9.1.1: http://autobuild.buildroot.org/?reason=qpdf-9.1.1.
These build failures were raised on powerpc with the following toolchain: http://autobuild.buildroot.org/toolchains/tarballs/br-powerpc-603e-basic-cpp-2019.11.tar.bz2

On my side, there is no issue to add a dependency on wchar on qpdf if the two other solutions are not acceptable. So, feel free to (re)close this PR.

yann-morin-1998:
> What toolchain doesn't have wchar? That is part of the standard and has been for a long time.

Toolchains for embedded devices where wchar supoprt is not needed. E.g. uClibc (a configurable C library for embedded systems) may lack support for wchar. Such systems may still need to run qpdf (e.g. as a backend for cups-filters to run in stand-alone printer booth for example).

So the remaining question is now: is wchar a hard-dependency of qpdf?

> being able to build and run these tests is essential to ensuring qpdf works

I totally agree with you in that part: tests are essential to ensure a good quality, and _being able to build and run tests_ **during developement** is essential.

However, I would disagree about not being able to disable them: tests are totally useless on a production system once it ships, so there would be no reason to build them in the first place.

jberkenbilt:
I'm closing this pull request and moving the discussion to #406. @yann-morin-1998 Thanks for your comments. They are helpful. I will respond in #406.

PR #409, opened 2020-03-07
Allow upscaling of the form XObject in placeFormXObject()

marcoscarpetta:
Hi, I'm the developer of PDF Mix Tool [1]. It would be useful if form XObject could be also upscaled in placeFormXObject(). I would need this for a 'resize page' feature.

I don't know if there was some reason to keep that limitation, but I tested the patch and it seems to work properly.

[1] https://gitlab.com/scarpetta/pdfmixtool

### DIFF
--- a/libqpdf/QPDFPageObjectHelper.cc
+++ b/libqpdf/QPDFPageObjectHelper.cc
@@ -753,10 +753,6 @@ QPDFPageObjectHelper::placeFormXObject(
     double xscale = rect_w / t_w;
     double yscale = rect_h / t_h;
     double scale = (xscale < yscale ? xscale : yscale);
-    if (scale > 1.0)
-    {
-        scale = 1.0;
-    }
 
     // Step 2: figure out what translation is required to get the
     // rectangle to the right spot: centered within the destination.


jberkenbilt:
Hi -- as you can see, this change caused test failures, but I'll see if I can figure out why that restriction is there and either put a comment in or remove it and adjust as needed. Is it possible you can supply some sample code that works correctly with your change but not without it? I want to ensure your use case is supported, so I'll help figure out the right solution. Thanks!

marcoscarpetta:
Hi, this is the sample code. In the `main` function, the first call to `overlay_page` is the sample code that works with my patch and doesn't without it. Basically, an example A4 page is placed over a blank A3 page using `QPDFPageObjectHelper::placeFormXObject`: with the patch the example page is extended to cover the blank A3 page, while without the patch it remains of the same size and gets centered in the A3 page. In this case, the rectangle of the A3 page was used as `rect` parameter when calling `placeFormXObject`.

The following call to `overlay_page` is an example to prove how this result (smaller page of the same size and centered in the bigger one) can be obtained with the patch also, by properly specifying the destination rectangle. The output file produced with or without the patch is (apparently) identical and it is also identical to that produces by the previous call without the patch.

Finally, the A4 page is placed over an A5 page and it gets shrunk in any case.

I included these further examples because I thought what make tests fail may be to assume that the overlayed page will not expand. If it is this the case, the problem can be solved by specifying the correct rectangle when `placeFormXObject` is called. Probably I'm completely wrong anyway... :)

I hope I was clear. Thank you!

[sample.zip](https://github.com/qpdf/qpdf/files/4314463/sample.zip)

jberkenbilt:
I've been thinking this over, and I think you're probably right. However, I need to reconstruct why I did it the way I did it and whether what's there is now is an outright bug or whether I need to somehow make it possible to achieve the old behavior or even have the old behavior as default.

jberkenbilt:
I'm adding allow_shrink (default true) and allow_expand (default false) to placeFormXObject for qpdf 10.0.0. Upon thinking about it, I think both should probably default to the same value, and that value should probably be false, but the defaults are set for backward compatibility since changing the meaning of a method call when called in the same way is one of the worst assaults an API designer can make on his/her users. However, with the new optional parameters, you should be able to achieve the result you want. Thanks for bringing this to my attention. Now if only I could discuss this with the version of me who had all this fresh in mind and know why I did it the way I did....

jberkenbilt:
A suitable fix will be in qpdf 10.0.0.

marcoscarpetta:
Thanks!

Issue #419, opened 2020-04-03
QPDF() is not deterministic


deanscarff:
In b8bdef0a, support was added to QPDFWriter for deterministic IDs.  However, 7588cac2 added code into the QPDF ctor that initializes m->unique_id using the time mixed with RNG data.  Admittedly, the non-determinism isn't exposed in the output except in the (unlikely) case of a collision.

Some solutions, none of them great:

- static counter of some sort (caller would need to guarantee thread-safety)
- callback of some sort (again, caller needs to guarantee thread-safety, but better for qpdf not having mutable static state.  Callbacks also add a lot of complexity to the API)
- content hash like QPDFWriter?  (I don't think this works, but just throwing it out there in case I'm wrong)

If this isn't a use case you're concerned about, I won't be offended if you close this issue.  However, if you are interested in a solution, I'm happy to take a stab at the implementation.

jberkenbilt:
Thanks for bringing this up. My inclination is to not care because, as you point out, this is not exposed in the output. Other than noticing the fact that QPDF has non-deterministic internal state, are you aware of a case where this matters in some way? If so, please let know since I don't want to dismiss this concern prematurely. Otherwise, I will probably just close this because I don't think it actually matters. I would point out that the code that the unique ID code replaced was using pointers, which are worse in all possible ways and which are recycled throughout the life of the system. Thanks!

deanscarff:
I haven't observed a case where this matters in any (concerning) way, I just happened to notice the extra syscalls when comparing versions (which are a small inconvenience when auditing, and also a small performance overhead).

Theoretically though, with the current implementation, you will start seeing collisions if you're generating IDs many times a second: the 32 high order bits from `time()` are identical, and the 32 low order bits have =~ 1 - exp(-k(k-1)/(2^33)) probability of collision.  This starts being plausible generating around 1000/s (about 0.012% probability of collision each second you're running).  Admittedly this is unrealistic for most users, CPUs and PDFs.
Maths: https://en.wikipedia.org/wiki/Birthday_problem#Approximations

However, a simple fix that reduces collision chances is ditching the low-entropy time() call and getting a full 64 bits from `QUtil::random()`.  Conveniently for me, this would also make the syscall go away ;)

deanscarff:
A good solution (IMO): use C++11's `std::atomic<unsigned long long>::fetch_add()`.  This is thread safe, doesn't add API complexity, but it may add portability issues depending on the platform.

[...]

PR #421, opened 2020-04-04
Use deterministic assignments for unique_id

deanscarff:
Fixes qpdf/qpdf#419

### DIFF
--- a/libqpdf/QPDF.cc
+++ b/libqpdf/QPDF.cc
@@ -1,6 +1,7 @@
 #include <qpdf/qpdf-config.h>  // include first for large file support
 #include <qpdf/QPDF.hh>
 
+#include <atomic>
 #include <vector>
 #include <map>
 #include <algorithm>
@@ -172,9 +173,8 @@ QPDF::QPDF() :
     // Generate a unique ID. It just has to be unique among all QPDF
     // objects allocated throughout the lifetime of this running
     // application.
-    m->unique_id = static_cast<unsigned long>(QUtil::get_current_time());
-    m->unique_id <<= 32;
-    m->unique_id |= static_cast<unsigned long>(QUtil::random());
+    static std::atomic<unsigned long long> unique_id{0};
+    m->unique_id = unique_id.fetch_add(1ULL);
 }
 
 QPDF::~QPDF()


libqpdf/QPDF.cc
```diff
@@ -44,6 +45,10 @@ static char const* EMPTY_PDF =
     "110\n"
     "%%EOF\n";
 
+namespace {
```
jberkenbilt, line 48: @deanscarff If it's this simple, I'd be inclined to accept this, as long as it doesn't mess up performance, but I doubt it will since QPDF objects themselves are relatively few. I am working on a performance benchmark system, so I will measure it. Please mimic the indentation style used everywhere else in the code. In other words, this should be
```
namespace
{
    std::atomic<unsigned long long> unique_id;
} // namespace
```
If you mimic the style and the code passes my benchmarks and CI, I will accept the PR as is. If the indentation style is not matched, then I will modify it before committing. In the end, the only difference is whose name is on the commit.

Thanks for doing this. I didn't know about `std::atomic` or anonymous namespaces before I saw this. I just started requiring C++11 in version 9.1.0, which makes things like this possible where they weren't before.
```

Comment (deanscarff) [reply libqpdf/QPDF.cc]: Actually I've changed this to just be a static (the are some subtle differences but they don't affect correctness), and it's more consistent with the surrounding code.  Curious to know the result of the benchmarks (I suspect on x86 architectures it should actually be faster this way, since the atomic implementations are userspace).

Comment (deanscarff) [reply libqpdf/QPDF.cc]: Uh, I should elaborate here: the function-static *is* safe in terms of initialization order, the previous (unnamed namespace / global static) was not (having a static QPDF would be undefined behaviour, and also initializing two dynamically-allocated QPDFs would cause a race on the initialization of the atomic).

The de-initialization order is still not defined, but I can't see a scenario where that would be a problem (some destructor would have to be creating a QPDF as a program terminates).

File: libqpdf/QPDF.cc [hunk 2]
```diff
@@ -172,9 +173,8 @@ QPDF::QPDF() :
     // Generate a unique ID. It just has to be unique among all QPDF
     // objects allocated throughout the lifetime of this running
     // application.
-    m->unique_id = static_cast<unsigned long>(QUtil::get_current_time());
-    m->unique_id <<= 32;
-    m->unique_id |= static_cast<unsigned long>(QUtil::random());
+    static std::atomic<unsigned long long> unique_id{0};
```
jberkenbilt, line 176: Yes, this looks right to me. I rely on safe static initialization in other parts of the code. It was the one thing I used C++-11 for that I could not have done without it unless I wanted to add dependencies on threading libraries.

jberkenbilt:
No significant impact on performance testing.

jberkenbilt:
Thanks for your contribution!

Issue #417, opened 2020-04-03
boringssl crypto provider


deanscarff:
BoringSSL is Google's fork of openssl, see: https://boringssl.googlesource.com/boringssl/

I've implemented a QPDFCryptoProvider_openssl based on boringssl, I expect it will also be compatible with openssl.  I can send a PR if you're interested in merging it in.

jberkenbilt:
No promises, but I'd be interested in taking a look.

jberkenbilt:
...and merging it in if it looks good. :-) I'm working on 10.0.0 now, but it's probably too late to make it in for that release.

PR #423, opened 2020-04-06
Add OpenSSL/BoringSSL crypto provider

deanscarff:
Fixes qpdf/qpdf#417


jberkenbilt:
@deanscarff This is a very high quality contribution. I am minutes away from preparing a 10.0.0 release, and I'm going to see if I can go ahead and include this. Do you have anything else immediately pending? I don't want to change the random number stuff for 10.0.0, but I'm open to changing the random number stuff to integrate with crypto. Most likely we can do that without breaking ABI (with only adding to it), so that could go in for 10.1. I'm going to take a good look at the  random generation code. The Windows native OS support has been a thorn in the side of qpdf for a long time as people get this problem acquiring crypt context. I don't want to break totally custom random number sources but I think the default should probably be to use the crypto provider.

jberkenbilt:
I went through the code thoroughly. I appreciate that you updated azure pipelines and README.md, and I agree with the decision to not enable it in the AppImage for now -- no need to make that larger than necessary. I'm inclined to leave it off in the debian package as well to avoid a dependency on openssl, but I will have to analyze that. In what environment are you using this? Are you building your own qpdf or using an OS package?

deanscarff:
Hi Jay, I don't have any other changes pending.  Re: the RNG, I think it's better to take your time and think about a good design than me trying to patch a new provider in right now.

I don't have any strong opinions about whether the openssl crypto provider should be included in the binary packages for linux distros; in fact I'd probably be inclined to stick with gnutls for Debian just because it's smaller.

I've been testing in two ways:
 * linking against BoringSSL & libqpdf from our own binaries; this is my main use case
 * building qpdf from source on a Debian system, against Debian's libssl-dev; this was for testing

In neither case am I redistributing any binaries or packages, and I haven't done testing on other systems.  I don't personally use Debian's qpdf often (but I don't have any reason not to).

Regarding environments: I tried to be simultaneously compatible with current BoringSSL and OpenSSL versions.  One issue I'm aware of is that OpenSSL has changed its API in ways that make it hard to be compatible with versions prior to v1.1.0 (released 2016); from the EVP_EncryptInit(3) man page:

> [in v1.1.0] EVP_CIPHER_CTX_reset() appeared and EVP_CIPHER_CTX_cleanup() disappeared.

I don't expect anyone to be building against such an old version of libssl, so I didn't write a configure.ac check.

FYI: in other projects for embedded devices I've used dlopen() to make crypto runtime-pluggable.  That allowed me to distribute small binaries and then use whatever is available on the target system.  I think it's probably overkill for qpdf though.  Sticking with gnutls by default is a good in my opinion, and if there are opinionated package maintainers, they have the controls to switch.

deanscarff:
Ah just saw you already did the RNG work; awesome!

jberkenbilt:
@deanscarff Thanks for all your comments. I think I agree with everything you said. When I did the RNG support, I didn't test with BoringSSL, but I did check the API to very the header, and it seems that openssl/rand.h is the right one for both. Though as I look at the code now, I see I should have included it in the .cc, not the .hh, but it doesn't actually make any difference in this case.

Let me know how it goes or if you have any other suggestions or pull requests!

PR #426, opened 2020-04-08
Fix copies in range-based for loops

deanscarff:
Fix unnecessary copies detected by:
https://clang.llvm.org/extra/clang-tidy/checks/performance-for-range-copy.html

### DIFF
--- a/libqpdf/JSON.cc
+++ b/libqpdf/JSON.cc
@@ -346,7 +346,7 @@ JSON::checkSchemaInternal(JSON_value* this_v, JSON_value* sch_v,
     if (sch_dict && (! pattern_key.empty()))
     {
         auto pattern_schema = sch_dict->members[pattern_key].getPointer();
-        for (auto iter: this_dict->members)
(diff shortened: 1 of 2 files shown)


jberkenbilt:
@deanscarff I'm going to have to release 10.0.1 tomorrow. Anything else pending? I broke getStreamData() (made it return raw stream data instead of throwing an exception if the data is valid but not filterable) -- a fairly serious error, so I want to get 10.0.1 out ASAP.

jberkenbilt:
The sanitizers failure is not your fault. I pushed a commit to master that wasn't ready and am going to revert it momentarily.

jberkenbilt:
After seeing this, I'm making a slightly different change that will cover your change, so while I will not be merging this PR, I will be accepting your contribution. Thanks!

jberkenbilt:
When I push my change, I'm going to close this PR with it. Feel free to retest. Thanks for all your work on this over the past several days. The timing is good since I'm actually doing some qpdf work now.

PR #433, opened 2020-04-21
fix WindowsCryptProvider fixes #432

### LINKED ISSUE #432: unable to acquire crypt context with new keyset: error number 8009000f: Object already exists.
What does this error mean when trying to encrypt:

unable to acquire crypt context with new keyset: error number 8009000f: Object already exists.

oltolm:
This fixes the problem by not creating a persistent key container to store private keys at all.
[Microsoft documentation](https://docs.microsoft.com/de-de/windows/win32/api/wincrypt/nf-wincrypt-cryptacquirecontexta) says this:

>Therefore, applications must not use the default key container to store private keys. Instead, either prevent key storage by passing the CRYPT_VERIFYCONTEXT flag in the dwFlags parameter, or use an application-specific container that is unlikely to be used by another application.

and

>For performance reasons, we recommend that you set the pszContainer parameter to NULL and the dwFlags parameter to CRYPT_VERIFYCONTEXT in all situations where you do not require a persisted key.

### DIFF
--- a/libqpdf/SecureRandomDataProvider.cc
+++ b/libqpdf/SecureRandomDataProvider.cc
@@ -42,60 +42,15 @@ class WindowsCryptProvider
   public:
     WindowsCryptProvider()
     {
-        if (!CryptAcquireContext(&crypt_prov,
-                                 "Container",
+        if (!CryptAcquireContextW(&crypt_prov,
+                                 NULL,
                                  NULL,
                                  PROV_RSA_FULL,
-                                 CRYPT_MACHINE_KEYSET))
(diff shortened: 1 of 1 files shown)


jberkenbilt:
Thanks, this seems good. I wonder why we did it the old way. Maybe this wasn't around? Maybe we just didn't know better? I'll merge this when I am next working on qpdf and can make a ChangeLog entry for it. I see CI passed, so I'm inclined to accept it as is. Hopefully this will be the end for real of this annoying issue.

jberkenbilt:
I'm rebuilding this based on current master. If all is well, I will merge.

PR #451, opened 2020-06-25
Replace memchr in BufferInputSource::findAndSkipNextEOL with manual memory search

jbarlow83:
On large files with predominantly `\n` line endings, `memchr(..'\r'..)`
may waste a considerable amount of time searching for a line
ending candidate that we won't need.

On the Adobe PDF Reference 6th ed (31 MB), this commit is 8x faster
at `QPDF::processMemoryFile()` (about 1.1s compared to 8.2s).

<hr/>

For some context in pikepdf, I have a custom input source to handle objects that
implement the IO stream protocol in Python, which means all reads and seeks
involve some Python overhead. For small reads and seeks this is expensive;
for large ones bulk storage is the overhead.

I got curious if `mmap()` would improve matters and found it to be about
2x faster when combined with this patch (without this patch, `mmap()` was
a fair bit slower, a surprising result that prompted my investigation). I thought
it's an interesting data point that QPDF might benefit from `mmap` too, although,
hard to say much.

Cheers.

### DIFF
--- a/libqpdf/BufferInputSource.cc
+++ b/libqpdf/BufferInputSource.cc
@@ -61,14 +61,15 @@ BufferInputSource::findAndSkipNextEOL()
     }
 
     qpdf_offset_t result = 0;
-    size_t len = QIntC::to_size(end_pos - this->m->cur_offset);
     unsigned char const* buffer = this->m->buf->getBuffer();
+    unsigned char const* end = buffer + end_pos;
+    unsigned char const* p = buffer + this->m->cur_offset;
 
-    void* start = const_cast<unsigned char*>(buffer) + this->m->cur_offset;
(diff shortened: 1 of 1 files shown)


jberkenbilt:
Thanks. I will almost certainly take that. My intention is that the next qpdf release will be focused heavily on performance. I believe the tokenizer is very inefficient because of the same kind of logic that you are correcting here. This could provide a good model. Using mmap when available is definitely on my radar. When I get to a natural stopping point on my primary work project, I will probably take some time and address this along with the small handful of open issues.

jbarlow83:
`pread` might help too.

jberkenbilt:
I'm rebuilding this based on current master. If all is well, I will merge.

PR #454, opened 2020-07-02
Check integer overflow in resolveObjectsInStream

deanscarff:
Fixes a crash found by fuzzing.

### DIFF
--- a/libqpdf/QPDF.cc
+++ b/libqpdf/QPDF.cc
@@ -2151,8 +2151,8 @@ QPDF::resolveObjectsInStream(int obj_stream_number)
 	}
 
 	int num = QUtil::string_to_int(tnum.getValue().c_str());
-	int offset = QUtil::string_to_int(toffset.getValue().c_str());
-	offsets[num] = offset + first;
+	long long offset = QUtil::string_to_int(toffset.getValue().c_str());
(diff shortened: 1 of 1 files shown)


libqpdf/QPDF.cc
```diff
@@ -2151,8 +2151,8 @@ QPDF::resolveObjectsInStream(int obj_stream_number)
 	}
 
 	int num = QUtil::string_to_int(tnum.getValue().c_str());
-	int offset = QUtil::string_to_int(toffset.getValue().c_str());
-	offsets[num] = offset + first;
+	long long offset = QUtil::string_to_int(toffset.getValue().c_str());
```
jberkenbilt, line 2154: QUtil::string_to_int ireturns int. If assigning to long long, use the version that returns long long.

p00ya: Assuming the current logic is correct (that toffset represents an "int" range), then I think using `string_to_int` to correct and safer (so the 32-bit range checks can take place).  That way the later addition is guaranteed not to overflow (at least on systems where `long long` is wider than `int`, which seems like a safe assumption).

To make this extra clear I can add a comment and/or a `static_cast`?

jberkenbilt:
After more careful study, I decided your change is correct.

PR #525, opened 2021-04-29
libqpdf/Pl_AES_PDF.cc: remove duplicated if branch

zdohnal:
Check for this->encrypt seems to be moved to plugged crypto
implementations, so it can be removed from Pl_AES_PDF.cc.

The issue was found out by our internal coverity scanner, here is the message:

```
Error: IDENTICAL_BRANCHES (CWE-398):
qpdf-10.3.1/libqpdf/Pl_AES_PDF.cc:212: identical_branches: The same code is executed regardless of whether "this->encrypt" is true, because the 'then' and 'else' branches are identical. Should one of the branches be modified, or the entire 'if' statement replaced?
#  210|       }
#  211|   
#  212|->     if (this->encrypt)
#  213|       {
#  214|   	this->crypto->rijndael_process(this->inbuf, this->outbuf);
```

### DIFF
--- a/libqpdf/Pl_AES_PDF.cc
+++ b/libqpdf/Pl_AES_PDF.cc
@@ -209,14 +209,7 @@ Pl_AES_PDF::flush(bool strip_padding)
         }
     }
 
-    if (this->encrypt)
-    {
-	this->crypto->rijndael_process(this->inbuf, this->outbuf);
-    }
-    else
-    {
-	this->crypto->rijndael_process(this->inbuf, this->outbuf);
-    }
+    this->crypto->rijndael_process(this->inbuf, this->outbuf);
     unsigned int bytes = this->buf_size;
     if (strip_padding)
     {


jberkenbilt:
It is amazing what kinds of things we may lurk in our code. Thanks for the PR. If true, I will merge after CI finishes. Otherwise, I will merge after CI finishes.

zdohnal:
Hi Jay,

thanks for looking into this! :)

jberkenbilt:
Very early in my programming career, I was reading some C code that had an array called widgets. There was all sorts of code that indexed into it, and there was some bug I was trying to fix (in someone else's code). I couldn't figure out where widgets was getting initialized, and I spent days pouring over the code until I eventually figured out it was getting initialized in a callback that had been set up with a pointer to some memory that widgets was pointing into. Anyway, during the process of trying to track this down, I had one of my first dreams that I woke up from remembering specific code. The code from the dream:

```C
if ((18 < 20) && (20 < 22))
{
    max_num_widgets = 15;
}
```

Issue #304, opened 2019-03-03
Build fails with "libtool:   error: specify a tag with '--tag'"


eesheesh:
I downloaded the master ZIP from GitHub just now, and after a successful `./configure`, it fails with:
```
$ make
sed -e 's:@PACKAGE_VERSION@:8.4.0:g' \
    -e 's:@docdir@:/usr/local/share/doc/qpdf:g' \
    < manual/qpdf.1.in > doc/qpdf.1
sed -e 's:@PACKAGE_VERSION@:8.4.0:g' \
    -e 's:@docdir@:/usr/local/share/doc/qpdf:g' \
    < manual/fix-qdf.1.in > doc/fix-qdf.1
sed -e 's:@PACKAGE_VERSION@:8.4.0:g' \
    -e 's:@docdir@:/usr/local/share/doc/qpdf:g' \
    < manual/zlib-flate.1.in > doc/zlib-flate.1
/bin/bash ./libtool --quiet --mode=compile g++ -Wold-style-cast -Wall -MD -MF libqpdf/build/BitStream.tdep -MP -Iinclude -Ilibqpdf  -c libqpdf/BitStream.cc -o libqpdf/build/BitStream.o; sed -e 's/\.o:/.lo:/' < libqpdf/build/BitStream.tdep > libqpdf/build/BitStream.dep
libtool: compile: unable to infer tagged configuration
libtool:   error: specify a tag with '--tag'
/bin/bash: libqpdf/build/BitStream.tdep: No such file or directory
libqpdf/build.mk:98: recipe for target 'libqpdf/build/BitStream.lo' failed
make: *** [libqpdf/build/BitStream.lo] Error 1
```

Have I missed any step? The README implies this should be enough.

Thanks!

Nardusgc:
I am stuck on the same issue, any workaround yet?

jberkenbilt:
Can you provide some information on what platform and compiler version you are using?

demsey2:
make sure you have the right compiler, 
I compiled this for Amazon Linux and I use this to get all I need to compile
```
yum groupinstall -y "Development Tools"
```

quisys:
Same

PR #528, opened 2021-05-28
Specify tag with libtool

DWagenk:
This solves
  libtool: compile: unable to infer tagged configuration
  libtool:   error: specify a tag with '--tag'
build issues when using toolchain that prevents libtools
inference rules from working correctly (e.g. ccache).

fixes: #304


DWagenk:
Thanks, that was quick!

PR #561, opened 2021-09-10
Fix QPDFEFStreamObjectHelper::{get,set}Subtype

fornwall:
The /Subtype entry that specifies the mime type of an embedded file is inside the embedded file stream dictionary directly, not it in the parameter dictionary.

See Table 45 and 46 in the PDF 1.7 specification:
https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#page=112


libqpdf/QPDFEFStreamObjectHelper.cc
```diff
@@ -141,7 +141,7 @@ QPDFEFStreamObjectHelper::setModDate(std::string const& date)
 QPDFEFStreamObjectHelper&
 QPDFEFStreamObjectHelper::setSubtype(std::string const& subtype)
 {
-    setParam("/Subtype", QPDFObjectHandle::newName("/" + subtype));
+    this->oh.getDict().replaceKey("/Subtype", QPDFObjectHandle::newName("/" + subtype));
```
jberkenbilt, line 144: Everything looks good here assuming CI passes, which it looks like it will. I have a preference for keeping my lines within 80 columns. I would break the line after `replaceKey(` and indent `"/Subtype"...` four spaces in from the previous line for stylistic consistency. See `stream.warnIfPossible` on line 159. Then I will gladly merge this change.

jberkenbilt (changes requested):

jberkenbilt:
Good catch. Thanks. The changes look good except the trivial nit of the 80+-column line. I'd fix it myself but then git annotate would not show you as the author of the fix.

libqpdf/QPDFEFStreamObjectHelper.cc
fornwall: Done, thanks!

fornwall:
Thanks for the review! I've added the suggested line break, does it look ok now?

jberkenbilt (approved):
Thanks for the fix!

jberkenbilt:
I'll merge this as soon as CI passes. It passed last time, so I'm sure we're good. I'll check it in 45 minutes or so.

Issue #422, opened 2020-04-05
Method for disabling rpath when building RPM?


edrozenberg:
When trying to build an RPM, I'm seeing the following errors related to `rpath`:

```
ERROR   0001: file '/usr/bin/qpdf' contains a standard rpath '/usr/lib64' in [/usr/lib64]
ERROR   0001: file '/usr/bin/zlib-flate' contains a standard rpath '/usr/lib64' in [/usr/lib64]
ERROR   0001: file '/usr/bin/fix-qdf' contains a standard rpath '/usr/lib64' in [/usr/lib64]
```

One suggestion is to pass `--disable-rpath` to `configure`, but this option is not available for the qpdf `configure` script:
```
configure: WARNING: unrecognized options: --disable-rpath
```

My fallback option is to tell the RPM build to proceed anyway via something like:
```
$ QA_RPATHS=$[ 0x0001|0x0010 ] rpmbuild my-package.src.rpm
```

but would prefer if there is some way to disable the use of `rpath` since that seems like the most "correct" solution.

jberkenbilt:
I'm not doing anything explicitly to get rpath in there, so I don't know how to prevent it from getting in there. I would have to do some research and digging. If this is a known, standard problem, maybe someone can shed some light. Can you provide a little more information as to what you're doing? Does the standard qpdf rpm have this problem? Are you doing this officially for a RedHat-like release, or are you just building it on your own? Thanks.

edrozenberg:
@jberkenbilt I'm building my own RPM under CentOS 7.7 using a standard approach documented in articles such as:

https://tecadmin.net/create-rpm-of-your-own-script-in-centosredhat/
https://www.techrepublic.com/article/making-rpms-part-4-finishing-the-spec-file
https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#_syntax

RedHat's documentation on rpath:
https://fedoraproject.org/wiki/RPath_Packaging_Draft

In this article they say:

> **Beware of Rpath**
> Sometimes, code will hardcode specific library paths when linking binaries (using the -rpath or -R flag). This is commonly referred to as an rpath. Normally, the dynamic linker and loader (ld.so) resolve the executable's dependencies on shared libraries and load what is required. However, when -rpath or -R is used, the location information is then hardcoded into the binary and is examined by ld.so in the beginning of the execution. Since the Linux dynamic linker is usually smarter than a hardcoded path, we usually do not permit the use of rpath in Fedora.
[...]

PR #565, opened 2021-09-16
Alternative fix for #422

tomty89:
Instead of not passing `-rpath $(libdir)` to libtool at all, only empty `hardcode_libdir_flag_spec{,_CXX}` that consist of "`rpath`" and `runpath_var{,_CXX}`. The former approach appears to prevent proper `libqpdf.la` generation and `libqpdf.lai` creation.

Ref.: https://github.com/termux/termux-packages/commit/a29657b770be854cabd2123ea86b6a2ac3cc1f36


jberkenbilt:
I think this looks fine, but I'll admit to not fully understanding exactly what the underlying issue is. I'll have to review the previous change and see what the difference is with the previous change. I will make sure I study it before whatever the next release is. I think I'm going to have to carve out some time for a 10.3.3 release as minor issues are piling up.

jberkenbilt:
If you have some kind of formula I can use to know whether a specific fix does or doesn't solve the underlying problem, it would be very helpful if you can provide it so I can update my notes. I have no issues with something like trying to create an RPM from a fedora docker image or something along those lines. My native environment is Ubuntu/Debian, but I've spent lots of times in the RedHat/Fedora/CentOS world as well and have good albeit outdated familiarity with that ecosystem.

tomty89:
Well, I am not exactly familiar with the DT_RPATH stuff, but it *seems* to me that the OP probably bumped into the problem because he  (unnecessarily?) used `--libdir /usr/lib64` for `./configure`. When I test in Arch, the rpath link flags are not actually used if `--libdir` is not set or set to `/usr/lib`, the implicit default when `--prefix` is `/usr`. (It probably has nothing to do with what the default is in `./configure`, but rather what was set as the default library path when `ld` or so was built.)

With that said, the OP (the part that one might want a way to disable the emission of DT_RPATH) still stands when for example one wants to build with prefix like `/usr/local`. I tested the PR with such case and it succeeded to prevent the executables from having DT_RPATH (confirmed with `readelf --dynamic` in addition to the build log).

jberkenbilt:
Sorry, what is `OP`?

tomty89:
Ahh sorry. I mean like the original poster / problem of #422.

jberkenbilt:
Sorry for the long delay. This patch modifies variables that are set inside of macros that I don't control, so I'm not willing to accept it as is since I think it will be too fragile. Maybe you can explain exactly what the issue is that you are trying to solve.

tomty89:
> This patch modifies variables that are set inside of macros that I don't control, so I'm not willing to accept it as is since I think it will be too fragile.

Yeah, it's kind of fragile. I think it's at least partly because libtool is more or less dead (in terms of development) and messy anyway. So everything can only go downstream.

> Maybe you can explain exactly what the issue is that you are trying to solve.

The problem is you sort of misunderstood what `-rpath` does in `ltmain.sh. (Well, to be fair, it does not separate things fine enough from the beginning.) The switch does not (only) prevent what was unwanted / complained (i.e. `DT_RPATH` / `DT_RUNPATH`) in #422. Rather its value is also crucial for the generation of `libqpdf.la{,i}`. The problem is only triggered when `--prefix` is not `/usr` though.

jberkenbilt:
I think I'm going to switch qpdf's build to cmake for version 11. This issue as well as some issues with the Windows build will probably just go away if I do that.

jberkenbilt:
I've come far enough on making qpdf build with cmake that I'm firmly committing to it for qpdf 11 and will therefore close this issue. I have verified that cmake handles rpath correctly at installation time.

PR #572, opened 2021-11-05
Fix typo in qpdf-manual.xml

neuhaus:
fix typo in manual so the given command will actually do what the previous sentence says it does.


jberkenbilt:
Thanks for pointing out this error. I will reword in a slightly different way, but I appreciate the fix!

PR #579, opened 2021-11-19
Ensure qpdf_oh handles returned by C-API functions are unique.

m-holger:
Return new qpdf_oh from qpdf_oh_wrap_in_array when input is already an array.
Update some doc comments in qpdf-c.h.

### DIFF
--- a/qpdf/qpdf-ctest.c
+++ b/qpdf/qpdf-ctest.c
@@ -530,7 +530,9 @@ static void test24(char const* infile,
     assert(! qpdf_oh_is_scalar(qpdf, mediabox));
     assert(qpdf_oh_is_array(qpdf, mediabox));
     assert(qpdf_oh_get_array_n_items(qpdf, mediabox) == 4);
-    assert(qpdf_oh_wrap_in_array(qpdf, mediabox) == mediabox);
+    qpdf_oh wrapped_mediabox = qpdf_oh_wrap_in_array(qpdf, mediabox);
+    assert(wrapped_mediabox != mediabox);
+    assert(qpdf_oh_get_array_n_items(qpdf, wrapped_mediabox) == 4);
     for (int i = 0; i < 4; ++i)
     {
         qpdf_oh item = qpdf_oh_get_array_item(qpdf, mediabox, i);
--- a/include/qpdf/qpdf-c.h
+++ b/include/qpdf/qpdf-c.h
@@ -134,7 +134,7 @@ extern "C" {
 
     /* Returns the error condition, if any.  The return value is a
      * pointer to data that will become invalid after the next call to
-     * this function, qpdf_next_warning, or qpdf_destroy.  After this
+     * this function, qpdf_next_warning, or qpdf_cleanup.  After this
      * function is called, qpdf_has_error will return QPDF_FALSE until
      * the next error condition occurs.  If there is no error
      * condition, this function returns a null pointer.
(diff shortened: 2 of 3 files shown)


qpdf/qpdf-ctest.c
```diff
@@ -530,7 +530,9 @@ static void test24(char const* infile,
     assert(! qpdf_oh_is_scalar(qpdf, mediabox));
     assert(qpdf_oh_is_array(qpdf, mediabox));
     assert(qpdf_oh_get_array_n_items(qpdf, mediabox) == 4);
-    assert(qpdf_oh_wrap_in_array(qpdf, mediabox) == mediabox);
+    qpdf_oh wrapped_mediabox = qpdf_oh_wrap_in_array(qpdf, mediabox);
+    assert(wrapped_mediabox != mediabox);
```
jberkenbilt, line 534: We will need to strengthen this test a little to make sure we are getting the same underlying object. Otherwise, this test would pass if qpdf_oh_wrap_in_array did some random crazy thing like returning a null object. I am happy to enhance the test myself. I would do two things:
* For this test, I would check in the for loop that the items in wrapped_mediabox and mediabox match
* I would add a test that wraps an array containing an indirect object and makes sure that the internal object has the same object ID as the original array's internal object.

I'll go ahead and approve CI for this, then I can either code the additional test or you can. It would take about as long to code it as it took me to type this comment, so I don't mind taking care of it if you're not comfortable with it.

jberkenbilt:
CI is good. I'm going to quickly code up an enhanced test, run that through CI once again, and then merge. Thanks for your contribution!

m-holger:
Thanks.

jberkenbilt:
FYI, this is my added test code. Sending through CI. Will merge when through.

```diff
diff --git a/qpdf/qpdf-ctest.c b/qpdf/qpdf-ctest.c
index c954a64d..9557d522 100644
--- a/qpdf/qpdf-ctest.c
+++ b/qpdf/qpdf-ctest.c
@@ -536,6 +536,9 @@ static void test24(char const* infile,
     for (int i = 0; i < 4; ++i)
     {
         qpdf_oh item = qpdf_oh_get_array_item(qpdf, mediabox, i);
+        qpdf_oh item2 = qpdf_oh_get_array_item(qpdf, wrapped_mediabox, i);
+        assert(qpdf_oh_get_int_value_as_int(qpdf, item) ==
+               qpdf_oh_get_int_value_as_int(qpdf, item2));
         printf("item %d: %d %.2f\n",
                i, qpdf_oh_get_int_value_as_int(qpdf, item),
                qpdf_oh_get_numeric_value(qpdf, item));
@@ -571,6 +574,12 @@ static void test24(char const* infile,
     assert(qpdf_oh_get_object_id(
                qpdf, qpdf_oh_get_array_item(qpdf, contents_array, 0)) ==
            qpdf_oh_get_object_id(qpdf, contents));
+    qpdf_oh wrapped_contents_array =
+        qpdf_oh_wrap_in_array(qpdf, contents_array);
+    assert(qpdf_oh_get_array_n_items(qpdf, wrapped_contents_array) == 1);
+    assert(qpdf_oh_get_object_id(
+               qpdf, qpdf_oh_get_array_item(qpdf, wrapped_contents_array, 0)) ==
+           qpdf_oh_get_object_id(qpdf, contents));
     qpdf_oh resources = qpdf_oh_get_key(qpdf, page1, "/Resources");
     qpdf_oh procset = qpdf_oh_get_key(qpdf, resources, "/ProcSet");
     assert(strcmp(qpdf_oh_unparse(qpdf, procset),
```

PR #584, opened 2021-11-29
Fix segmentation fault in QPDF::getFilename

m-holger:
Code like
```c
    qpdf_data qpdf1 = qpdf_init();
    qpdf_oh_is_bool(qpdf1, 3);
```
causes  segmentation faults.

### DIFF
--- a/libqpdf/QPDF.cc
+++ b/libqpdf/QPDF.cc
@@ -2730,7 +2730,12 @@ QPDF::getUniqueId() const
 std::string
 QPDF::getFilename() const
 {
-    return this->m->file->getName();
+    std::string fn = "";
+    if (this->m->file.getPointer())
+    {
+        fn = this->m->file->getName();
+    }
+    return fn;
 }
 
 std::string


jberkenbilt:
Since this code path can only occur because of a programmer error, returning a dummy value is not the correct fix here. Instead, I'm initializing m->file to something so we don't get a segmentation fault. The current fix will, in most cases, give an unhelpful error message, but if/when I implement safe object handles and use them throughout the library, the situation should improve. The problems that can happen are quite pervasive if you don't properly call process methods. The ideal fix would be to check in a bunch of places, but this would probably have a negative performance impact. My fix will be zero overhead except in the case where the code is used incorrectly. I wonder if I need a way of attaching an error message to an invalid QPDFObjectHandle that would be included if it's dereferenced.

This is one of these "if I knew then what I know now" moments. For the most part, qpdf takes away the developer's opportunities to use the API incorrectly, e.g., not caring whether an object is direct or indirect, but this is not one of those cases. That said, I'm more worried about avoiding the segmentation fault than making this super obvious since basically nothing works if you don't call one of the read methods. My fix also updates the comment in qpdf-c.h. I might want to reread that file with a mind toward seeing whether a developer who starts there gets what they need. I'd have to first reread QPDF.hh and QPDFWriter.hh.

I think your efforts and going to result in increased usability of the C API all around.

When my fix goes through CI, I will merge it, which will close this in favor of that fix.

jberkenbilt:
btw, the failure of my commit in CI is because I canceled it before it was finished after pushing up a second commit that fixes some comments and changes the exception type.

PR #585, opened 2021-11-29
Ensure consistency QPDF::getRoot vs QPDF::getTrailer

m-holger:
`qpdf->getTrailer()->getKey("/Root")` and `qpdf->getRoot()` should display similar behaviour for non-opened `qpdf`'s re throwing exceptions.

### DIFF
--- a/libqpdf/QPDF.cc
+++ b/libqpdf/QPDF.cc
@@ -2769,7 +2769,15 @@ QPDF::getExtensionLevel()
 QPDFObjectHandle
 QPDF::getTrailer()
 {
-    return this->m->trailer;
+    QPDFObjectHandle trailer = this->m->trailer;
+    if (! trailer.isDictionary() 
+        || this->m->file->getName() == "closed input source")
+    {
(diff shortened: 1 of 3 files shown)


libqpdf/QPDF.cc
```diff
@@ -2769,7 +2769,15 @@ QPDF::getExtensionLevel()
 QPDFObjectHandle
 QPDF::getTrailer()
 {
-    return this->m->trailer;
+    QPDFObjectHandle trailer = this->m->trailer;
+    if (! trailer.isDictionary() 
+        || this->m->file->getName() == "closed input source")
```
jberkenbilt, line 2774: This is fragile. The clue here is that the object is uninitialized. This also wouldn't work the same in the after-close case as the never-opened case. But either way, I don't want to add checking against programmer errors in this code. getTrailer is called a lot, and anything extraneous here hurts performance. I am unlikely to accept any pull request that adds additional logic to getTrailer. The approach I would think about here might be to have some way to control what exception is generated when an uninitialized object is accessed, and I would do this as part of the implementation of a safer QPDFObjectHandle.

PR #586, opened 2021-12-02
C-API trap exceptions in qpdf_get_root

m-holger:
Calling qpdf_get_root when pdf file is not open raises an exception.

### DIFF
--- a/qpdf/qpdf-ctest.c
+++ b/qpdf/qpdf-ctest.c
@@ -510,6 +510,10 @@ static void test24(char const* infile,
     qpdf_oh root_from_trailer = qpdf_oh_get_key(qpdf, trailer, "/Root");
     assert(qpdf_oh_get_object_id(qpdf, root) ==
            qpdf_oh_get_object_id(qpdf, root_from_trailer));
+    qpdf_data pdf2 = qpdf_init();
+    qpdf_oh uninit_root = qpdf_get_root(pdf2);
+    assert(qpdf_oh_is_null(pdf2, uninit_root));
+    assert(qpdf_has_error(pdf2));
     qpdf_oh pages = qpdf_oh_get_key(qpdf, root, "/Pages");
     assert(qpdf_oh_is_dictionary(qpdf, pages));
     qpdf_oh kids = qpdf_oh_get_key(qpdf, pages, "/Kids");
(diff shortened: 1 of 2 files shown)


qpdf/qpdf-ctest.c
```diff
@@ -500,7 +500,7 @@ static void test24(char const* infile,
 		   char const* outfile,
 		   char const* outfile2)
 {
-    /* This test case is designed for minimal.pdf. */
+    /* This test case is designed for minimal.pdf. */
```
jberkenbilt, line 503: Please remove this whitespace change.

jberkenbilt (changes requested):
Interestingly, I am, at this moment, implementing almost this exact change in a similar but not identical way. I might accept this change, or I might decide to do it a little differently. Please fix the trailing whitespace issue, and then I can decide what to do. This is a little simpler than what I was going to do, but it might really be just as good. Thanks.

m-holger:
Sorry, not sure how I missed that.

jberkenbilt:
I'm thinking I may not literally take this implementation, but I will probably take the behavior into my implementation, which is a little more broadly encompassing. Basically, I've been struggling with a few decisions:
* Given that the C API object handle methods don't return error codes, how do we handle errors?
  * Convert the warnings as now (probably bad)
  * Allow them to cause the program to exit, sometimes also as now (definitely bad)
  * Have them write to stderr by default but allow user to define an error handler (okay, but cumbersome, particularly given that these errors are usually programmer errors)
  * Your solution: just have them be the current error and have them retrievable -- probably the best of these but with some problems

Problems with this solution:
* To check each time, you'd have to call qpdf_has_error and qpdf_more_warnings frequently, which would be even more cumbersome than now
* If you call multiple things that have errors with checking for intervening errors, what do we do? There's no stacking of errors, so we'd probably have to just overwrite a previous error. This is no worse than with other things that have errors.

Honestly I'm on the fence between the stderr/error handler solution and this solution, leaning slightly toward the stderr/error handler solution.

Rationale: while it's usually evil for libraries to write stuff to stderr, it should be okay if I give the caller a way to suppress it (by defining an error handler). Lazy programmers (or people who don't notice this change) won't do this, and then if they make the kind of mistake that causes an error, they'll notice because a message will get written to stderr. Then they can fix their code. Also, the only things that can cause errors here are either programmer errors (like dereferencing an uninitialized object or trying to combine object handles from different QPDF objects, which is pretty much impossible with the C API) or severe things like running out of memory of file I/O which are probably going to make the program crash anyway.

So....I'll probably go with the error handler solution. I'm already part way through implementing it. If I go that route, I will close this.

By the way, I have commits on my branch that are in the process of clarifying docs. I will request your review on those.

jberkenbilt:
Closed in favor of a different solution, now merged.

PR #587, opened 2021-12-02
C-API expose new_object as qpdf_oh_new_object

m-holger:
Allow users to clone qpdf_oh handles in order to take advantages of QPDF's reference counting.


jberkenbilt (approved):
Great change implemented perfectly. I will most likely accept this once CI is green.

PR #588, opened 2021-12-04
C-API expose QPDF::getObjectByID  and QPDF::replaceObject

m-holger:


jberkenbilt:
I'm just going to go ahead and do this one. It's just a few lines of code, particularly after the last round of modifications.

PR #597, opened 2021-12-17
C-API expose QPDFObject::getTypeCode and QPDFObject::getTypeName

m-holger:
Also fix minor typo

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,7 +1,7 @@
 2021-12-16  Jay Berkenbilt  <ejb@ql.org>
 
 	* Add several functions to the C API for working with pages. C
-	wrappers around sevearl of the "Legacy" page operations from
+	wrappers around several of the "Legacy" page operations from
 	QPDFObjectHandle.hh have been added. See "PAGE FUNCTIONS" in
 	qpdf-c.h for details. Fixes #594.
 
--- a/include/qpdf/qpdf-c.h
+++ b/include/qpdf/qpdf-c.h
@@ -587,6 +587,21 @@ extern "C" {
 
     typedef unsigned int qpdf_oh;
 
+    typedef int QPDF_OT_CODE;
+#   define QPDF_OT_UNINITIALIZED 0
+#   define QPDF_OT_RESERVED 1
+#   define QPDF_OT_NULL 2
+#   define QPDF_OT_BOOLEAN 3
+#   define QPDF_OT_INTEGER 4
+#   define QPDF_OT_REAL 5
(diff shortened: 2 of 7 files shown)


include/qpdf/qpdf-c.h
```diff
@@ -587,6 +587,21 @@ extern "C" {
 
     typedef unsigned int qpdf_oh;
 
+    typedef int QPDF_OT_CODE;
```
jberkenbilt, line 590: I won't accept this -- it duplicates information. To implement this, I would want to move object_type_e from QPDFObject.hh to Constants.h and just make QPDFObject::object_type_e an alias to it. I can just do that in my current work and then you can redo this PR to use that.

jberkenbilt (changes requested):

ChangeLog
```diff
@@ -1,7 +1,7 @@
 2021-12-16  Jay Berkenbilt  <ejb@ql.org>
 
 	* Add several functions to the C API for working with pages. C
-	wrappers around sevearl of the "Legacy" page operations from
```
jberkenbilt, line 4: I use a spell checker, which catches this. I will add spell checking to CI.

include/qpdf/qpdf-c.h
jberkenbilt: My solution seems like it is also going to duplicate information, but I'm going to use it anyway, so please stay tuned.

m-holger: I didn't like my solution either, but couldn't think of a better one.

Do we actually need to publish the constants in qpdf-c.h? We could have a simple comment that refers QPDFObject.h

I for one won't use the constants but define my own enum. I suspect most other users will do the same.

jberkenbilt: I think it's important to have the enumerated type exposed to the C API if we are going to return values from it. Anyway, I already pushed the change to main.

jberkenbilt:
@m-holger I think this looks good except where we commented. The spelling error is fixed on main already. If you rework this to get rid of the `#define` and new type and use the now exposed `qpdf_object_type_e` as your return type, I will most likely accept this. If you don't get to it, I will take this pull request in its current form and modify it to work so I can get this by the end of the day. I was originally thinking I'd release 10.5 this weekend, but I want to give it a few days for this stuff to sink in. I'll probably aim for early next week. If I don't get it out by around Tuesday, then I will probably want to wait until after the new year.

jberkenbilt:
I've incorporated this.

PR #603, opened 2022-01-05
Use DOC_DEST variable for installing documentation

dirkmueller:


jberkenbilt:
Hello! The correct fix to this is to make it conditional on the existence of the file, not whether it is being built. The qpdf distribution process itself drops the PDF in place after building it externally. I had that logic in before, but I guess I inadvertently lost it when I revamped the documentation. It's a one-liner that I'm happy to just fix, and I'll probably need a similar fix for the html. No need to resubmit the pull request. Thanks for bringing this to my attention!

dirkmueller:
@jberkenbilt ah okay, thats fine as well. I have removed the PDF testing bit from this PR, so it should be cleaner to merge. if you fix it differently thats good as well, feel free to merge or close as you wish.

jberkenbilt:
@dirkmueller Great, thanks! I'll keep it open until I make the fix, but since I will be making a different change to the same line, there's no point in merging this. I will tag this issue as closed in the fix and be sure to include the doc -> $(DOC_DEST) change. Thanks for catching that.

jberkenbilt:
I merged it and then added the additional code. Thanks!

PR #609, opened 2022-01-21
Add new convenience methods QPDFObjectHandle::isNameAndEquals, etc

m-holger:
Add methods isNameAndEquals, isDictionaryOfType, isStreamOfType


libqpdf/QPDFObjectHandle.cc
```diff
@@ -497,6 +497,35 @@ QPDFObjectHandle::isScalar()
...
+QPDFObjectHandle::isNameAndEquals(std::string const& name)
+{
+    return isName() && (getName() == name);
+}
+
+bool
+QPDFObjectHandle::isDictionaryOfType(std::string const& type,
+                                     std::string const& subtype)
+{
+    if(isDictionary() && hasKey("/Type") &&
```
jberkenbilt, line 510: Please put a space between `if` and `(`

jberkenbilt (changes requested):
This is an excellent contribution. Other than the small lexical issue, my only request would be to move most of the testing into test_driver.cc. I'd prefer to keep it so that all core functionality is tested without the C API and that qpdf-ctest.c primarily be used to test just that the C API is working. Also, though perhaps I'm overlooking it, I don't see any tests of isDictionaryOfType for a dictionary without a /Type key. I see tests for a non-dictionary and tests for a dictionary where the keys don't match, but not tests for a dictionary without /Type or a dictionary where /Type or /Subtype are not names. I believe the implementation correctly handles those cases. With those fixes, I would accept this pull request.

m-holger:
Thanks. Take 2.

jberkenbilt (approved):

jberkenbilt:
I made two commits after this: one to fix whitespace (with myself as committer and you as author to protect git annotate) and a second to add a ChangeLog entry. I don't use ChangeLog for everything since that's what version control history is for, but I use ChangeLog for any API enhancements or significant fixes since I use the ChangeLog rather than the history when I create release notes. Sometimes I keep the release notes as a running document when I'm updating, but with all the work on the manual, I'm not doing that for this round of changes and will do the release notes closer to when I'm ready to release.

PR #617, opened 2022-01-26
Remove some if statements and simplify some boolean expressions

m-holger:
Use QPDFObjectHandle::isNameAndEquals, isDictionaryOfType and
isStreamOfType.

### DIFF
--- a/libqpdf/QPDFWriter.cc
+++ b/libqpdf/QPDFWriter.cc
@@ -1241,9 +1241,7 @@ QPDFWriter::enqueueObject(QPDFObjectHandle object)
                 " another file.");
         }
 
-        if (this->m->qdf_mode &&
-            object.isStream() && object.getDict().getKey("/Type").isName() &&
-            (object.getDict().getKey("/Type").getName() == "/XRef"))
+        if (this->m->qdf_mode && object.isStreamOfType("/XRef"))
         {
             // As a special case, do not output any extraneous XRef
             // streams in QDF mode. Doing so will confuse fix-qdf,
@@ -1474,8 +1472,7 @@ QPDFWriter::willFilterStream(QPDFObjectHandle stream,
     QPDFObjGen old_og = stream.getObjGen();
     QPDFObjectHandle stream_dict = stream.getDict();
 
-    if (stream_dict.getKey("/Type").isName() &&
-        (stream_dict.getKey("/Type").getName() == "/Metadata"))
+    if (stream_dict.isDictionaryOfType("/Metadata"))
     {
         is_metadata = true;
     }
@@ -1691,11 +1688,8 @@ QPDFWriter::unparseObject(QPDFObjectHandle object, int level,
             QTC::TC("qpdf", "QPDFWriter preserve Extensions");
             QPDFObjectHandle adbe = extensions.getKey("/ADBE");
             if (adbe.isDictionary() &&
-                adbe.hasKey("/BaseVersion") &&
-                adbe.getKey("/BaseVersion").isName() &&
-                (adbe.getKey("/BaseVersion").getName() ==
-                 "/" + this->m->final_pdf_version) &&
-                adbe.hasKey("/ExtensionLevel") &&
+                adbe.getKey("/BaseVersion").isNameAndEquals(
+                            "/" + this->m->final_pdf_version) &&
                 adbe.getKey("/ExtensionLevel").isInteger() &&
                 (adbe.getKey("/ExtensionLevel").getIntValue() ==
                  this->m->final_extension_level))
@@ -1764,7 +1758,7 @@ QPDFWriter::unparseObject(QPDFObjectHandle object, int level,
                         for (int i = 0; i < filter.getArrayNItems(); ++i)
                         {
                             QPDFObjectHandle item = filter.getArrayItem(i);
-                            if (item.isName() && item.getName() == "/Crypt")
+                            if (item.isNameAndEquals("/Crypt"))
                             {
                                 idx = i;
                                 break;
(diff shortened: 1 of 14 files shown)


m-holger:
The formatting looks very odd. Any idea why in the vicinity of the changes tabs are expanded to 4 spaces while if one expands the files, in the normally hidden parts of the files tabs are expanded to 8 spaces?

jberkenbilt:
I'll come back to this after I'm done with my current refactoring work. Regarding the formatting, I don't remember seeing github highlight tabs before, so maybe this is new. I'd have to actually pull down the changes and look at them outside of github to be sure.

Your last PR had some weird indentation which I fixed in a separate commit. I guess you are using VS Code. I'm not sure what to do about code formatting. I could start using that ignore-revs thing you mentioned and do one big commit to remove all the tabs. Maybe that would help. (But I would want to do that myself.) I could look into some kind of automatic formatting.

I do all my code editing in emacs and have my "c-style" set to this (referring to myself as "Q" in this context):

```elisp
(defvar q-cc-style
  '("Q"
    (c-basic-offset . 4)
    (c-comment-only-line-offset . 0)
    (c-offsets-alist
     (defun-block-intro . +)
     (block-open . 0)
     (substatement-open . 0)
     (knr-argdecl-intro . 2)
     (statement-cont . +)
     (case-label . +2)
     (access-label . -2)
     (statement-case-intro . +2)
     (statement-case-open . +2)
     (arglist-intro . +)
     (arglist-close . +)
     (inline-open . 0)
     (inlambda . 0)
    )
  )
  "*cc-mode Style for Q's C++ indentation preference"
)
```

Most of this dates back to 1997 and has to do with style guidelines that were in place on a project I was working on at the time. It's not how I would do it today. Maybe I need to investigate a batch-mode automatic formatter (like prettier in javascript or go fmt or rust fmt) that would format all my code and just fix it once and for all, as long as I could make emacs get close to it. At this point, my bracing and indentation style has shifted in all my other work, and I do qpdf the way I used to do it only for maintaining consistency in the code base.

jberkenbilt:
clang-format looks very promising. I'm going to add it to my list of things to consider.

jberkenbilt:
Seems as though git blame has a -w option.

m-holger:
> I could start using that ignore-revs thing you mentioned and do one big commit to remove all the tabs.

I think that might be good. I looked  into the formatting a bit more after the white-space issues of my last PR. Up to then I used to think that the very odd formatting  I noticed in many part of qpdf was due to the reluctance to unnecessarily touch white space. It turns out that that it is due to VS Code interpreting the files as 4 space indentation, 1 indentation = 4 spaces per tab (same with Genie, the editor that I used until fairly recently). I have managed to persuade both editors to show indentations properly. Unfortunately while doing so, one auto-in/dedents by 8 spaces, and the other uses tabs.I am sure it can be sorted, but removing tabs feels easier.

jberkenbilt:
To be clear, even in the absence of a massive reformatting (which I am now strongly leaning toward), it has been my policy for several years to systematically remove tabs on lines that I change for any other reason. All of our many conversations have convinced me that I put too much emphasis on git blame over convenience and readability though so I am rethinking some things.

libqpdf/QPDFWriter.cc
```diff
@@ -1469,16 +1467,11 @@ QPDFWriter::willFilterStream(QPDFObjectHandle stream,
                              PointerHolder<Buffer>* stream_data)
 {
     compress_stream = false;
-    is_metadata = false;
```
jberkenbilt, line 1472: This is intentional. I like to initialize references right up front so that they have predictable values even if the function throws an exception early. For example, if stream.getObjGen or stream.getDict threw an exception, moving the initialization would change the meaning.

libqpdf/QPDFWriter.cc
```diff
@@ -1469,16 +1467,11 @@ QPDFWriter::willFilterStream(QPDFObjectHandle stream,
...
 
     QPDFObjGen old_og = stream.getObjGen();
     QPDFObjectHandle stream_dict = stream.getDict();
 
-    if (stream_dict.getKey("/Type").isName() &&
-        (stream_dict.getKey("/Type").getName() == "/Metadata"))
-    {
-        is_metadata = true;
-    }
+    is_metadata = stream_dict.isDictionaryOfType("/Metadata");
```
jberkenbilt, line 1474: You can replace the if, but don't assign to is_metadata. Set is_metadata inside the if.

Probably best not to replace `if (condition) { a = true; }` with `a = condition` unless it's blindingly obvious like there are no intervening statements between the initial declaration and the assignment, but even then, sometimes it's the way it is for consistency.

jberkenbilt (changes requested):
Except as noted, these are good changes. If you fix the comment, I'll accept this PR. However, I'd like to keep these kinds of code cleanup PRs to a minimum. They do help improve the code, but they also carry some risk and use up limited time/energy. That said, I very much appreciate the thought and effort that went into this and the resulting improvement in the code.

libqpdf/QPDFWriter.cc
m-holger: Makes sense

jberkenbilt (approved):
I'll merge once it passes CI. Thanks again.

libqpdf/QPDFWriter.cc
m-holger: Done the change, although I don't necessarily agree.

To my mind, within the method `is_metadata` is a const bool defined by the assignment, and immutable from this point on. To me, `if (condition) { a = true; }` implies `a` is mutable and may be changed later in the method.

jberkenbilt: That may be true in this case, but I frequently code in a way where I check many conditions and flip values of booleans as I go, and, in the absence of an `else`, you have to study the whole function to know whether the two are equivalent.

PR #619, opened 2022-01-27
Fix QPDFObjectHandle::isOrHasName

m-holger:
Ensure isOrHasName returns true if object is an array and the name is
present anywhere in the array.


jberkenbilt (approved):
Looks great -- correct fix and good tests. Thanks.

PR #620, opened 2022-01-29
Overload QTC::TC to allow bool discriminators

m-holger:


jberkenbilt:
I'm not interested in this change. I use QTC across multiple projects in multiple languages and have no desire to change its behavior or usage.

jberkenbilt:
I also feel that this obscures the meaning of TC. I like the fact that I can, at a glance, see that all of a range of numbers starting from 0 appear in different situations. I sometimes also use a -1 or other out of range number as a contract that a condition does not occur among other conditions that do.

m-holger:
Fair enough

m-holger:
Personally I feel that it is fairly obvious that the presence of flag implies n in {0, 1}, 2 flags n in {0, 1, 2, 3} which are the most common cases (other than n in {0}). For the remaining cases, the original form is still available.

That said, I don't feel strongly about it.

PR #621, opened 2022-01-30
Fix doc types and add some comments

m-holger:


jberkenbilt:
Thanks for the fixes.

PR #630, opened 2022-02-05
C-API expose QPDFObjectHandle::getValueAs... accessors

m-holger:


libqpdf/qpdf-c.cc
```diff
@@ -1238,6 +1238,22 @@ QPDF_BOOL qpdf_oh_get_bool_value(qpdf_data qpdf, qpdf_oh oh)
...
+    qpdf_data qpdf, qpdf_oh oh, QPDF_BOOL* value)
+{
+    return do_with_oh<QPDF_BOOL>(
+        qpdf, oh, return_false, [value](QPDFObjectHandle& o) {
+            QTC::TC("qpdf", "qpdf-c called qpdf_oh_get_value_as_bool");
+            bool v = *value;
+            QPDF_BOOL result = o.getValueAsBool(v);
+            if (result)
+            {
+                *value = QPDF_TRUE;
```
jberkenbilt, line 1251: This doesn't look right. Why not just `o.getValueAsBool(*value)`? I don't see how this would ever set *value to false.

jberkenbilt: or `*value = v` or `*value = v ? QPDF_TRUE : QPDF_FALSE`

libqpdf/qpdf-c.cc
```diff
@@ -1284,6 +1340,19 @@ char const* qpdf_oh_get_real_value(qpdf_data qpdf, qpdf_oh oh)
...
 }
 
+QPDF_BOOL qpdf_oh_get_value_as_real(
+    qpdf_data qpdf, qpdf_oh oh, char const** value, size_t* length)
+{
+    return do_with_oh<QPDF_BOOL>(
+        qpdf, oh, return_false, [qpdf, value, length](QPDFObjectHandle& o) {
+            QTC::TC("qpdf", "qpdf-c called qpdf_oh_get_value_as_real");
+            auto result = o.getValueAsReal(qpdf->tmp_string);
+            *value = qpdf->tmp_string.c_str();
```
jberkenbilt, line 1350: wouldn't it be better to only do this if result is true?

libqpdf/qpdf-c.cc
```diff
@@ -1303,6 +1382,19 @@ char const* qpdf_oh_get_name(qpdf_data qpdf, qpdf_oh oh)
...
 }
 
+QPDF_BOOL qpdf_oh_get_value_as_name(
+    qpdf_data qpdf, qpdf_oh oh, char const** value, size_t* length)
+{
+    return do_with_oh<QPDF_BOOL>(
+        qpdf, oh, return_false, [qpdf, value, length](QPDFObjectHandle& o) {
+            QTC::TC("qpdf", "qpdf-c called qpdf_oh_get_value_as_name");
+            auto result = o.getValueAsName(qpdf->tmp_string);
+            *value = qpdf->tmp_string.c_str();
```
jberkenbilt, line 1392: same as above -- only if result is true?

libqpdf/qpdf-c.cc
```diff
@@ -1313,6 +1405,19 @@ char const* qpdf_oh_get_string_value(qpdf_data qpdf, qpdf_oh oh)
...
 }
 
+QPDF_BOOL qpdf_oh_get_value_as_string(
+    qpdf_data qpdf, qpdf_oh oh, char const** value, size_t* length)
+{
+    return do_with_oh<QPDF_BOOL>(
+        qpdf, oh, return_false, [qpdf, value, length](QPDFObjectHandle& o) {
+            QTC::TC("qpdf", "qpdf-c called qpdf_oh_get_value_as_string");
+            auto result = o.getValueAsString(qpdf->tmp_string);
+            *value = qpdf->tmp_string.c_str();
```
jberkenbilt, line 1415: same

libqpdf/qpdf-c.cc
```diff
@@ -1323,6 +1428,19 @@ char const* qpdf_oh_get_utf8_value(qpdf_data qpdf, qpdf_oh oh)
...
 }
 
+QPDF_BOOL qpdf_oh_get_value_as_utf8(
+    qpdf_data qpdf, qpdf_oh oh, char const** value, size_t* length)
+{
+    return do_with_oh<QPDF_BOOL>(
+        qpdf, oh, return_false, [qpdf, value, length](QPDFObjectHandle& o) {
+            QTC::TC("qpdf", "qpdf-c called qpdf_oh_get_value_as_utf8");
+            auto result = o.getValueAsUTF8(qpdf->tmp_string);
+            *value = qpdf->tmp_string.c_str();
```
jberkenbilt, line 1438: same

jberkenbilt (changes requested):

libqpdf/qpdf-c.cc
m-holger: Changed . 
*value does not work as a bool&

libqpdf/qpdf-c.cc
m-holger: Not only better, but also correct. Changed

m-holger:
Sorry for the first attempt.

jberkenbilt (approved):
Thanks!

jberkenbilt:
> Sorry for the first attempt.

No problem. You've certainly seen enough of my slips. :-)

PR #631, opened 2022-02-06
Add method QPDFObjectHandle::getKeyIfDict

m-holger:


m-holger:
I am changing this to only silence warnings if called from a Null object.

jberkenbilt:
> I have changed the design of getKeyIfDict so as to only silence warnings if it is invoked by a Null object. ...

It seems as though our views on this topic have converged.

jberkenbilt (approved):
I am now 100% in agreement with your implementation. Two comments:
* is `getKeyIfDict()` still the right name for this? It might be. Or `getKeyIfNotNull()` might be more accurate. I'm leaning slightly toward changing it but am open to being convinced otherwise.
* There should be a comment explaining the behavior. Something like
```
// Return a null object if called on a null; otherwise call getKey().
// This makes it easier to query lower-level dictionaries with code like
// auto font = page.getKeyIfDict("/Resources").getKeyIfDict("/Font");
```
I'll await your response before merging in case you'd like to change the names and add the comments (rewording the comment as you fit including using the correct name).

m-holger:
I prefer `getKeyIfDict()` to `getKeyIfNotNull()`. To me, `getKeyIfNotNull()` gives the impression that we are merely guarding against null rather than that receiving a null is an expected possibility that has been fully anticipated.

How do you feel about `getKeyOrNull()`?

The suggested comment makes sense to me. For 10.7/11.0, perhaps have a more comprehensive section on how to deal with optional  items. Perhaps within an code example.

jberkenbilt:
> How do you feel about `getKeyOrNull()`?

I thought about that too, but it's confusing because it always returns Null if the key is not there. Let's just leave it as `getKeyIfDict` with a comment.

jberkenbilt:
> For 10.7/11.0, perhaps have a more comprehensive section on how to deal with optional items. Perhaps within a code example.

Good idea. Maybe I'll check over the existing examples and see if any of them can be changed to use this, or you can. I'd be okay with that for 10.6.

m-holger:
> I thought about that too, but it's confusing because it always returns Null if the key is not there. Let's just leave it as `getKeyIfDict` with a comment.

Works for me. Are you ok with me to add a 3rd commit or do you prefer me force pushing clean replacements for the 2 existing commits?

> I'd be okay with that for 10.6.

I assume this refers to updated example only. (For something more comprehensive e.g. manual / header file paragraphs I feel more thinking time would be good)

I'll have a look at examples if I get time. If I plan to touch any I'll let you know beforehand. Similarly, can you let me know if you start working on any of them.

jberkenbilt:
I'll just merge as is and add the comment. I am making another pass through examples for another purpose and can handle it if I think of it. Otherwise, not important. I agree we can wait on doc stuff for later.

jberkenbilt:
Thanks again. :-)

PR #632, opened 2022-02-07
Tidy example CLI usage

m-holger:
Change "-" to "--" for named parameters.
Remove spaces inside "[ option ]" for optional parameters.
Fix "pdf-mod-info --dump file" to match usage message.


jberkenbilt:
Interesting story about this...it was a actually a production tool used at my previous job, written by a different developer, which I was allowed to take when they let me open source qpdf. I guess I never really looked at it that closely. Thanks.

PR #635, opened 2022-02-08
Fix doc typos

m-holger:


include/qpdf/QPDFObjectHandle.hh
```diff
@@ -163,8 +163,8 @@ class QPDFObjectHandle
     // writeToken(). Example:
     // writeToken(QPDFTokenizer::Token(QPDFTokenizer::tt_name, "/text/plain"))
     // would write `/text#2fplain`, and
-    // writeToken(QPDFTokenizer::Token(QPDFTokenizer::tt_string, "a\\(b"))
-    // would write `(a\(b)`
+    // writeToken(QPDFTokenizer::Token(QPDFTokenizer::tt_string, "a\\(b)")
```
jberkenbilt, line 166: actually, this one is correct as written -- the ) belongs outside. This is an example of creating a string with an unbalanced parentheses.

jberkenbilt (approved):

jberkenbilt:
I'll take this and back out the one correction that's incorrect. Thanks again for all your editorial review.

include/qpdf/QPDFObjectHandle.hh
m-holger: Should have been obvious if I had thought about it.

PR #639, opened 2022-02-09
Remove /Type test in QPDFJob::shouldRemoveUnreferencedResources

m-holger:
Remove test for type == /XObject as value is optional
(as per spec 8.10.2).


m-holger:
For > 10.6?

m-holger:
This should actually be fixed in QPDFObjectHandle::isFormXObject.

jberkenbilt:
Looks good. Thanks. This doesn't add any API so I might do a 10.6.1 with this and some other tweaks in a week or so.

PR #642, opened 2022-02-11
Use std::abs() for floating-point tolerance tests

### LINKED ISSUE #641: test_driver.cc fails to build with Apple clang 10.0
qpdf/test_driver.cc:3298:12: error: call to 'abs' is ambiguous
    assert(abs(num - 42.0) < 1e-100);
    
Three instances of abs() should be changed to fabs(), and "#include <math.h>" added.

jaklein:
Fixes #641

### DIFF
--- a/qpdf/test_driver.cc
+++ b/qpdf/test_driver.cc
@@ -30,8 +30,8 @@
 #include <stdlib.h>
 #include <assert.h>
 #include <limits.h>
-#include <math.h>
 #include <map>
+#include <cmath>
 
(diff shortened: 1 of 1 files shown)


jberkenbilt:
I could rather `#include <cstddef>` and use `std::abs` instead. Could you try that fix?

jaklein:
std::abs is defined in <cmath>, but it works.

jberkenbilt:
I'm going to fix this a different way. Thank for the contribution.

PR #644, opened 2022-02-11
include math.h

pitwegner:
Otherwise, compilation fails with `error: call of overloaded 'abs(double)' is ambiguous`


jberkenbilt:
This has already been fixed on main (just a few minutes ago). I'll release 10.6.1 in a day or two.

PR #660, opened 2022-02-25
What is CI? ci.debian.net? Took me some time

ZjYwMj:
I hope I am not the only one wondering what CI is. Finding that out was not a quick search. Do other people find it more quickly? And I am still not sure it is what this patch refers to. Do give a hint about that, or use the full term, or apply the patch, or something. Thank you.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ QPDF is known to build and pass its test suite with mingw (latest version tested
 
 # Building Documentation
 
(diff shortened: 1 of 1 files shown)


jberkenbilt:
CI is a general term meaning continuous integration. ci.debian.net is just one, but that's not the one this refers to. This project has its CI on github actions. I guess I forgot the term CI wouldn't be familiar to some people. Anyway, it's not relevant, so I just removed the phrase "used in CI".

PR #661, opened 2022-02-26
Overload QPDF::get...Value to take a default value as parameter

m-holger:


jberkenbilt:
FYI I am not accepting any pull requests right now that change API in any way. Until final freeze for Ubuntu 22.04, I can still get bug-fix only changes in, but not changes that modify the API. My intention is to pause API changes until after I have code formatting and cmake done. So I won't be merging this any time soon, and probably there will be merge conflicts before I'm ready to take it.

At some point, I may decide to let main move on from the 10.6 branch so I can start taking pull requests and pushing changes into main but I'm not ready to do that. With working on formatting and build, I don't really want to have other changes coming in.

That said, I see your change is only going to the default if the object is null. I thought we were going to go for the default if the object was not of the correct type as well as if it was null. Was your understanding different?

jberkenbilt:
Also, I'm not planning any more 10.x releases. I'm going to treat cmake as ABI breaking because I don't want to hassle with making sure libraries and DLLs are created exactly the same way. In fact, I hope DLLs won't be created the same way, so cmake will definitely be ABI breaking on Windows. I want to get code formatting and cmake in early so I can live with them for a while before a release and make sure they are working the way I want them to.

jberkenbilt:
> I'll try to remember to merge it before I do the formatter changes so you don't have to redo it.

I put a note in TODO.

m-holger:
> I thought we were going to go for the default if the object was not of the correct type as well as if it was null. 

Your remark made me think about the behavior when the object is of an unexpected type. At the moment we have

- `"13"_qpdf.getIntValue(42)` --> 13, no warning
- `"null"_qpdf.getIntValue(42)` --> 42, no warning
- `"/Page"_qpdf.getIntValue(42)` --> 0, type warning

Perhaps it would be better to have 

- `"13"_qpdf.getIntValue(42)` --> 13, no warning
- `"null"_qpdf.getIntValue(42)` --> 42, no warning
- `"/Page"_qpdf.getIntValue(42)` --> 42, type warning

Therefore, perhaps a better implementation would be

```cpp
    QPDF_DLL
    long long getIntValue(long long default_value, bool required = false);

long long
QPDFObjectHandle::getIntValue(long long default_value, bool required)
{
    if (isInteger())
    {
        return dynamic_cast<QPDF_Integer*>(obj.get())->getVal();
    }
    else if (required || (! isNull()))
    {
        typeWarning("integer", "returning default value");
        QTC::TC("qpdf", "QPDFObjectHandle integer returning 0");
    }
    return default_value;
}

long long
QPDFObjectHandle::getIntValue()
{
    return getIntValue(0, true);
}

```

jberkenbilt:
I've been thinking about this too since yesterday, and I'm actually thinking that we shouldn't do this at all. It's starting to feel like interface sprawl to me. Perhaps it's my personal philosophy showing, but I prefer a somewhat more minimalistic approach, and I like to avoid ambiguity.

I'll review this history for context even though you were an active participant and clearly know this, but this is so you can see my thinking and for the benefit of others who read this ticket.

Before 10.6, there was only one way to get the value of something, and it always gave a value and sometimes gave a type warning. It evolved by mistake because, for a long time, its behavior was to either give a value or throw and exception, but that was causing qpdf and other qpdf-based tools to reject PDF files that were broken in ways that didn't really matter. Breaking the API was not an option because those methods are so core to qpdf's API, so I added the concept of the type warning. Things got a lot better, but we have the pesky type warnings. If you want to avoid them, you have to do an explicit type check first.

The new API, which I like, returns a bool indicating whether it was successful. If it was successful, you get a value. If not, you don't get a value. The reference isn't touched. That gives you a more convenient way to work with objects that is equally safe but avoids type warnings entirely. You initialize the variable that will receive the value to some sane value of your choice, then call getValueAsWhatever. If you are willing to accept your default value without caring whether the object was the correct type, just ignore the return value. If you care, you still have to write a conditional, but it's simpler, and the code flow is smoother. If I were starting from scratch and had to choose one interface over the other, I would choose this second one over the first one.

This third suggested interface now creates confusion. Does it give a type warning? Sometimes: it does if it's other than the type or null, otherwise it doesn't. Does it use your default value? Originally the answer was also sometimes, though your suggestion improves that somewhat at the cost of another optional parameter. To me, this throws a circuit breaker: the API is too complicated. It potentially saves you one line of code at the cost of having to understand subtleties about what its exact behavior is. As a library user, I would never want to use that interface, and as a library maintainer, I would not want to offer it.

Even something resembling rust's Option wouldn't really help very much because the code to access it would be as cumbersome as what qpdf has now. It's not cumbersome in rust because you have `if let`, `match`, and destructuring, but C++ doesn't have those.

So let's stop with what we have and not add a third way to get an object value. I would point out that in any specific application, one can easily do something like

```
auto asInt = [](QPDFObjectHandle oh){ // implement your desired behavior here };
int v = asInt(oh);
```

I tend to use that style when I want a local behavior. It's self-documenting in a way, and it avoids sprawl of the underlying API.

m-holger:
I agree my suggestion is unnecessarily complicated. That said, there are two facts to consider:

- The PDF spec has both optional and required entries  for dictionaries, and in fact optional entries appear to be three times as common as required entries.
-  While in many cases the defaults returned by `get...Value` is correct, there is also a significant number of cases where the PDF spec states different default values.

Given that optional entries are the most common case they should not require special handling, and in particular, they should not require guarding against missing (optional) values, i.e. we should be able to simply write
```cpp
auto page_layout = qpdf.getRoot().getKey("/PageLayout").getName(false, "/SinglePage");
``` 
While it is true that writing
```cpp
auto page_layout = "(SinglePage)"_qpdf;
(void) qpdf.getRoot().getKey("/PageLayout").getValueAsString(page_layout);
``` 
suppresses the warning on a missing optional value and only takes one extra line, it also suppresses all warnings about unexpected object types returned by getKey(), whether as a result of programmer error as in this case or whether as a result of a damaged/incorrect  file.

Anyway, below is my third attempt. The simple rules are

If the entry is of the right type, return it's value, otherwise return the default value.
If the entry is required and missing, warn.
If the entry is of the wrong type, warn.

```cpp
    QPDF_DLL
    long long getIntValue(bool required = true, long long default_value = 0);

long long
QPDFObjectHandle::getIntValue(bool required, long long default_value)
{
    if (isInteger())
    {
        return dynamic_cast<QPDF_Integer*>(obj.get())->getVal();
    }
    else if (required || (! isNull()))
    {
        typeWarning("integer", "returning 0");
        QTC::TC("qpdf", "QPDFObjectHandle integer returning 0");
    }
    return default_value;
}
```

jberkenbilt:
I can see your arguments, but I still feel like I'd rather not complicate the API with this. I have the general feeling of this problem in mind. Maybe when I work on the fluent interface or the json v2 support, some fresh ideas might come to mind.

jberkenbilt:
Let's not do this one.

PR #664, opened 2022-03-02
Fix QPDFPageObjectHelper::getCropBox and getTrimBox

m-holger:
Ensure the correct boxes are created/updated.

### DIFF
--- a/libqpdf/QPDFPageObjectHelper.cc
+++ b/libqpdf/QPDFPageObjectHelper.cc
@@ -367,7 +367,13 @@ QPDFPageObjectHelper::getTrimBox(bool copy_if_shared)
     QPDFObjectHandle result = getAttribute("/TrimBox", copy_if_shared);
     if (result.isNull())
     {
-        result = getCropBox(copy_if_shared);
+        QTC::TC("qpdf", "QPDFPageObjectHelper getTrimBox");
+        result = getCropBox();
+        if (copy_if_shared)
+        {
+            result = result.shallowCopy();
+            this->oh.replaceKey("/TrimBox", result);
+        }
     }
     return result;
 }
@@ -378,7 +384,13 @@ QPDFPageObjectHelper::getCropBox(bool copy_if_shared)
     QPDFObjectHandle result = getAttribute("/CropBox", copy_if_shared);
     if (result.isNull())
     {
+        QTC::TC("qpdf", "QPDFPageObjectHelper getCropBox");
         result = getMediaBox();
+        if (copy_if_shared)
+        {
+            result = result.shallowCopy();
+            this->oh.replaceKey("/CropBox", result);
+        }
     }
     return result;
 }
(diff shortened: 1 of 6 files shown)


jberkenbilt:
I believe the code is correct as written and that this fix is incorrect. Can you describe the intention here? The fallback values returned are per the PDF spec, section 14.11.2 Page boundaries, which describes the default value to return for each box if not present. Default value for /TrimBox is /CropBox, and default value for /CropBox is /MediaBox. (Also /BleedBox and /ArtBox default to /CropBox, but I don't have helpers for those.) The intention is not to add these boxes if they don't exist but to fallback to the default values.

m-holger:
The current code is correct for getCropBox() and getTrimBox(), but not for getCropBox(true) or getTrimBox(true).

In the case  where a page does not have a TrimBox, getTrimBox(true) does not create a TrimBox, and modifying the returned box will modify the CropBox or MediaBox that supplied the default value.

In the case where the page does not have a CropBox, getCropBox(true) does not create a CropBox but will create a MediaBox if it is not already present. Modifying the box returned by getCropBox(true) will modify the MediaBox for the page.

m-holger:
> The intention is not to add these boxes if they don't exist but to fallback to the default values.

This is what "copy_if_shared" suggests.

m-holger:
Why would you call `getTrimBox(true)` if you did not want to modify it? 

How would calling `getAttribute("/TrimBox", true)` help, (keeping in mind that  `getTrimBox(true)` calls  `getAttribute("/TrimBox", true)` anyway)?

Who would call `page.getTrimBox(true).setArrayItem(...)` to modify the MediaBox of the root node of the page tree?

More bizarrely, who would call `page.getTrimBox(true).setArrayItem(...)` to  create a CropBox (but not a TrimBox) on those pages whose ancestors have an explicit  CropBox?

What is the logic of calling `getCropBox(true)` from within `getTrimBox(true)` but calling  `getMediaBox(false)` from within  `getCropBox(true)`?

Sorry, this does not make sense to me.

jberkenbilt:
I've decided this needs a more radical fix. The current API is too dangerous, and I can't change the behavior without breaking the worst backward compatibility rule, so it calls for an _API-breaking_ change. I haven't worked out what it is, but it will probably be something like removing these two calls and adding two new ones: getEffectiveWhateverBox and getWhateverBoxToModify (TBD) that will have clear, unambiguous semantics for uses cases in which you want to get whatever box defines crip, trim, whatever or in the case where you want to explicitly set or modify one without affecting the others. I will have to think through what the API should be, but I will put it on the list for 11. This will force people who are using those methods to face this question and change their code. There is precedent for this kind of fix. I do it when an existing API is dangerously confusing and likely used very rarely, which is most likely the case here.

If you'd like, you can create a non-PR issue raising your concerns about the exiting API, or else I can do so making sure to reference and capture details from this. Then we can close the PR. I'll fix the issue early in the qpdf 11 cycle and get your comments on whatever API I end up coming up with. Of course, feel free to make suggestions. I don't want a PR for it right now though as I will want to do a deep dive into the code and try to reconstruct what I had in mind when I came up with the current API.

jberkenbilt (changes requested):
Marking this as "request changes" so I remember that this PR raises a bug but that I don't agree with the solution as presented.

jberkenbilt:
Closed in favor of #758

PR #665, opened 2022-03-02
Tidy example pdf-double-page-size

m-holger:
Also fix typo in pdf-attach-file example.


jberkenbilt:
I'm not going to take this until I have a chance to think through the bbox thing more closely.

jberkenbilt:
If you can remove everything from this PR that has to do with the get*Box problem and only include things like removal of the extraneous "", I'll take this. Otherwise, make sure that the issues with pdf-double-page-size are mentioned in the other issue (though we can't forget because this won't compile after that fix), and I'll grab the cosmetic fixes from this PR and close it.

m-holger:
I am not sure what you are saying. My interpretation is that you want me to revert back to using getKey instead of getAttribute. If that is the correct interpretation I can do that.

That said, the switch to getAttribute is not really related to the get*Box issue as it only deals with inheritance, not default values.

jberkenbilt:
Yeah, I read it too fast. The change is good.

PR #673, opened 2022-03-12
Minor manual fixes

m-holger:


jberkenbilt:
Thnask for fixing all my tpyos. I are apprciatte it.

PR #718, opened 2022-06-17
Code tidy: remove unnecessary parameter in QPDFObjectHandle::shallowC…

m-holger:
…opyInternal and rename shallowCopy

(change to  shallowCopy(bool first_level_only = false) for qpdf11?)

### DIFF
--- a/include/qpdf/QPDFObjectHandle.hh
+++ b/include/qpdf/QPDFObjectHandle.hh
@@ -1579,7 +1579,7 @@ class QPDFObjectHandle
         bool cross_indirect,
         bool first_level_only,
         bool stop_at_streams);
-    void shallowCopyInternal(QPDFObjectHandle& oh, bool first_level_only);
+    QPDFObjectHandle shallowCopy(bool first_level_only);
     void releaseResolved();
     static void setObjectDescriptionFromInput(
         QPDFObjectHandle,
--- a/libqpdf/QPDFObjectHandle.cc
+++ b/libqpdf/QPDFObjectHandle.cc
@@ -2804,25 +2804,20 @@ QPDFObjectHandle::hasObjectDescription()
 QPDFObjectHandle
 QPDFObjectHandle::shallowCopy()
 {
-    QPDFObjectHandle result;
-    shallowCopyInternal(result, false);
-    return result;
+    return shallowCopy(false);
 }
 
 QPDFObjectHandle
 QPDFObjectHandle::unsafeShallowCopy()
 {
-    QPDFObjectHandle result;
-    shallowCopyInternal(result, true);
-    return result;
+    return shallowCopy(true);
 }
 
-void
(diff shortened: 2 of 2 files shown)


jberkenbilt:
This change results in an extra copy of QPDFObjectHandle. The reason I did it the way I did was specifically to avoid the extra copy.

jberkenbilt:
Anyway, I'm inclined not to take this change. If you'd like to make a case for it, please do. Otherwise, if you accept my explanation, you can close this. I keep meaning to wrap my head around move in C++11. I've read about it a few times, but I haven't understood what it does from what I've read. I'd have to write some sample code to make sure I fully understand how it works. I don't think it can be used to prevent a copy from occurring when returning a plain, stack-allocated object.

m-holger:
When I looked at move in C++ I got the distinct impressions the the current compilers are pretty good at optimization. I am not convinced that the QPDFObjectHandle is actually returned on the stack. I am not claiming any deep knowledge, but from my reading this looks likely that a good compiler will optimize this away.

The problem will certainly disappear if first_level_only is made an optional parameter.

m-holger:
As far as move in general is concerned, I am not sure that this is likely to be of much benefit to the current QPDFObjectHandle, given that the only "expensive" part is the shared_ptr, which doesn't seem overly expensive to me.

Where it would become useful is if one wanted to start treating object handles properly as values rather than (mostly) as references, i.e. where every assignment / object creation involves a call to shallowCopy. I tried playing around with that, but came to the conclusion that this is too drastic change - even for my taste!

jberkenbilt:
Yeah, I was thinking about maybe just making first_level_only an optional parameter, but I'm not sure. Good point about the compiler optimizing it. I'm undecided. I'll think about it with the other changes.

jberkenbilt:
I've decided not to take this change. I like the way it's currently implemented better. If I were not thinking that I'm going to completely disallow unsafe shallow copies, I'd just make first_level_only an optional parameter. What I'll probably do instead is, if I make unsafe shallow copies completely invalid, would be to deprecate unsafeShallowCopy and, when it's time to remove it, just remove the indirection entirely. I know I have done this all over the code, but my evolving sense of coding style leads me away from having private and public overloads of the same method. The "internal" marker on the name sends a signal, at least to me (and hopefully other people who are learning my coding practices) that this is a method you can only call if you really know what you're doing and maybe is unsafe in some way. It's not a hard rule, and as you've heard me say over and over, this code has evolved over 20+ years, and different parts of the code reflect my thinking at different times. Oh well.

As you see, I've accepted most of your other tidying pull requests.

m-holger:
I accept your points.

That said, I still think it makes sense to drop the first parameter both for code clarity and in order to give the compiler the best chance to optimize the extra call out. 

I force pushed an amended commit.

jberkenbilt:
I think we can find a way to improve this with some of the other changes we are discussing.

PR #723, opened 2022-06-25
Code tidy: remove redundant calls to QPDFObjectHandle::isInitialized

m-holger:


jberkenbilt:
Will merge after it goes through CI.

PR #724, opened 2022-06-25
Refactor QPDFObjectTypeAccessor and QPDFObjectHandle::dereference

m-holger:


libqpdf/QPDFObjectHandle.cc
```diff
@@ -3187,12 +3116,15 @@ QPDFObjectHandle::assertPageObject()
     }
 }
 
-void
-QPDFObjectHandle::dereference()
+bool
+QPDFObjectHandle::dereference(bool unguarded)
```
jberkenbilt, line 3120: Please change `unguarded` to `allow_uninitialized` and get rid of the default for the method. Since it's a private method, I'd rather have an explicit value. It looks like `allow_uninitialized` might have the opposite sense of `unguarded`. To me, `unguarded` doesn't mean anything, but if anything, I would think `unguarded == true` would skip the check rather than perform it.

include/qpdf/QPDFObjectHandle.hh
```diff
@@ -1573,7 +1573,7 @@ class QPDFObjectHandle
     void typeWarning(char const* expected_type, std::string const& warning);
     void objectWarning(std::string const& warning);
     void assertType(char const* type_name, bool istype);
-    void dereference();
+    bool dereference(bool unguarded = false);
```
jberkenbilt, line 1576: See impl for comment -- change `unguarded` to `allow_uninitialized` and remove default.

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2868,7 +2798,6 @@ QPDFObjectHandle::copyObject(
                                " reserved object handle direct");
     }
 
-    dereference();
```
jberkenbilt, line 2871: I'd leave a comment in place of this call.
```c++
    // Type checks above guarantee that the object has been deferenced.
```

jberkenbilt (changes requested):
Good change pending my review comments.

libqpdf/QPDFObjectHandle.cc
m-holger: By unguarded I tried to express that is not called in an
```cpp
if (dereference()) {
    ...
}
```
construct. At least some of the existing calls of  the form
```cpp
dereference(true);
```
are actually called with uninitialized QPDFObjectHandles  and do actually throw exception. My preferred solution would  be to change those calls to to be guarded by an if statement and once this is done loose the `unguarded` parameter completely.

m-holger: Or loose the calls completely and replace with a comment as per your comment below.

m-holger:
I have force pushed an updated version that hopefully addresses all requested changes

PR #725, opened 2022-06-29
Rationalise interface between QPDFObject, QPDFObjectHandle and QPDF

m-holger:
Centralise casting of QPDFObjects and reduce repeated dereferencing.

### DIFF
--- a/include/qpdf/Constants.h
+++ b/include/qpdf/Constants.h
@@ -82,6 +82,8 @@ enum qpdf_object_type_e {
     /* Additional object types that can occur in content streams */
     ot_operator,
     ot_inlineimage,
+    /* Object types internal to qpdf - to be moved */
+    ot_unresolved,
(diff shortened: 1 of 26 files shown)


m-holger:
I don't like the name asStreamWithAssert. Any sugggestions?

jberkenbilt:
There are too many changes here for me to accept this without a discussion. I have scanned through the code, but I will have to do a more thorough review so I can try to figure out what you're trying to accomplish here. There are things that look like temporary code too but I have to inspect more carefully.

m-holger:
These are bits and pieces on the way to #726 which I felt would make sense even if you decided against splitting QPDFObject.

I put them into a single PR because of merge conflicts. Once you decide which ones you like I will do individual PRs in sequence.

We can discuss the individual commits later.  (Dinner time over here)

m-holger:
Not worth considering at this time.

PR #727, opened 2022-07-07
Inline QPDFObjGen methods

m-holger:
ABI breaking change


jberkenbilt:
A 3% to 4% improvement is a pretty big deal, so this is definitely worth it. I had some slight hesitation over QUtil.hh being included since this basically means it's going to be included everywhere, but a quick inspection of the code revealed that nothing that includes QPDFObjGen.hh doesn't also include QUtil.hh.

Issue #442, opened 2020-05-20
ParserCallbacks - Timings


drcdr:
I'm using ParserCallbacks to do stream parsing, and I'm looking at trying to improve the performance timings.  It appears that a lot of time is spent in QPDF outside of handleObject(), and I'm trying to figure out what to do next.  (I'm using QPDF libraries, not building from source.)

On one PDF file I've been using, times using std::chrono::steady_clock are:
- parsePageContents(cb): 273 ms total
- cb::handleObject(): 60 ms total

As part of that, I've separately timed just the stream extraction of all of the relevant page /Contents streams (using pipeContentStreams) and that's 1.8 ms.

So, about 273-60-1.8 = 210 ms is in QPDFObjectHandle::parseContentStream_data(), outside of cb::handleObject(). 

I'm not sure how to proceed with potential optimization.  Are there any "simple" things that you can think of that could be done?

jberkenbilt:
@drcdr It's funny you should bring this up. In my TODO file on my work branch (not yet in master), I have notes about optimizing the lexical layer of qpdf. I have not done detailed performance measurements, but I have a strong suspicion that a lot of the performance issues with the qpdf library stem from the lexer (QPDFTokenizer) being inefficient. It's not super bad....but there's a lot of function call and string copying overhead that could be improved. It is my intention for an upcoming release to put some energy into optimizing it. However, that's going to be tricky, and it's unlikely I would accept a pull request because I am going to want to think hard about the API, and it will dig deep into the library internals.

As far as simple things, if you are on Linux or another UNIX variant (possibly including macos), you might consider using mmap on your input file and using processMemoryFile to read the file. I have never tried this myself, by I think it should work fine. Alternatively, could read the whole file into memory and use processMemoryFile. qpdf can in some cases read the same parts of a file multiple times, especially where streams are involved, because its default behavior is to keep memory usage low even for very large files.

If you have any performance benchmarking you'd like to share, I can keep it in mind when doing optimization. With the release of 10.0.0, I added a bare bones performance checking script, checked in as performance_check, that runs the qpdf command line with various flags and times how long it takes. It's not very granular, but I would be expanding it somewhat to be able to get timings separately on the lexer. If you have any sample code you could share, I could possibly use it as I measure and try to improve performance.

[...]

Issue #311, opened 2019-03-29
qpdf consumes all my memory and crashes for a specific file


balaprassanna:
I am using `qpdf version 8.4.0`
While running `qpdf --check-linearization qpdfhike.pdf`, qpdf consumes all the machine memory and crashes.

https://drive.google.com/file/d/15Fhh-Vk37YvDO7yh9tfFqHukHssVxTEr/view?usp=sharing

balaprassanna:
Is there a way to limit maximum memory used by qpdf?

jberkenbilt:
Sorry for the delay here. I see no reason that qpdf shouldn't be able to handle this file, but it is having trouble with it. This is similar to #311 I think. This file has 1364 pages and 168,523 objects, so it is similar in size to the other file. I will find some time to study why qpdf is taking so long. There's obviously some kind of performance problem here. Thanks for the report.

NicolasGoeddel:
We got a similar issue here. On bigger PDFs (400+ pages, 10-20 MB) the qpdf process consumes more than 20 GB of RAM and then many other processes crash because of too less memory. In the end it is even hard to SSH into the machine and kill qpdf.

jberkenbilt:
Addressing this is my top priority on qpdf right now.

jberkenbilt:
There seem to be at least two separate problems here, but the two files (this one and #305) share a common attribute: object streams containing objects that are arrays containing a very large number of elements (like around 100,000). One thing is that qpdf's memory usage, which is intended to be nearly constant throughout file processing, actually grows with the number of object stream objects. This is the result of a design flaw that I should be able to correct. The other problem is that the parser seems to be intolerably slow for these very long arrays. I will have to study that a bit more. I'm hopeful that I can get the performance reasonable for these files though.

jberkenbilt:
[...]

PR #729, opened 2022-07-08
Tune QPDFObjectHandle::parseInternal

m-holger:
Summary of changes:

To facilitate refactoring, the static parseInternal method together with the related setObjectDescriptionFromInput method were moved to a new QPDFParser class.

To reduce code duplication and improve readability new private warn methods were added to the QPDFParser class.

To improve readability and improve performance, the olist, offset, contents_string and contents_offset stacks were combined into a single stack. To improve performance, both the new stack as well as the existing state stack were implemented as std::vectors rather than as SparseOHArrays.

Treatment of null objects was tuned by recognizing that QPDF_Arrays and QPDF_Dictionarys do not store direct null objects. They discard the null object and "store" them as a missing object instead. Creation of new distinct null objects for each direct null encountered, or setting descriptions/offsets for them, was eliminated. With the change, the impact of parsing a null object is the addition of one object handle to olist, a vector of object handles. No new null objects are created.

Once parsed, arrays continue to be stored as SparseOHArrays. This maintains the overwhelming majority of the benefits created by SparseOHArray.


jberkenbilt:
The reason for using `SparseOHArray` is that there have been files in the wild that have arrays with millions of `null`s in them, and older versions of qpdf were extremely slow on those files. I believe there is a test for it somewhere, or at least there was. Also this is extremely critical code, so I will have to give it a lot of scrutiny.

m-holger:
In its current incarnation, runtime for the pdf 1.7 is reduced by around 3.5 to 4% against main.

Parsing strings containing either an array of 2 arrays of 5 million nulls each or an array of 5 arrays of 2 million nulls each, the current version reduces the runtime by 1% to 2% against main.

I also tried a version optimized for handling arrays with large numbers of nulls, achieving a 33.5% to 36% improvement compared to main, at a cost of reducing the improvement for the pdf 1.7 spec to around 2.5%. Given that (hopefully) there are far more files without a large number of nulls, on balance this does not feel worthwhile to me. However, I may try a hybrid version that switches to the null optimized approach if it encounters a large number of nulls.

m-holger:
Removing QPDFTokenizer::Members ups the improvement to over 4.5%.

This raises the question: how likely is it that QPDFTokenizer will require updates between major versions? If it is highly unlikely then there is no real downside to removing Members. If there is a possibility that changes will be required, adding an empty Members class would provide the necessary flexibility.

jberkenbilt:
It is highly unlikely that I would modify the tokenizer between major versions. Other than performance optimization, there is probably nothing I would change in the tokenizer. There is an open ticket about string copying in the tokenizer, which is something I have been aware of for a long time. I may take a look at that but it's a good candidate for a PR. I don't want to solve it by adding a third party dependency. There's #442 and I feel like there was another one where someone suggested use of a third-party library that has string builders that are more efficient, but I am not able to find that one at this moment.

jberkenbilt:
In other words, I think removing Members from tokenizer is probably a good enhancement and one I probably would have made in an effort to tune performance, so thanks for taking a look. I'd probably have to look at a profiler to decide if there are other strong candidates for removal of Members.

m-holger:
Do you want me to push the last commit as a separate PR as its probably a lot easier to review than the other commits?

m-holger:
In its current incarnation the PR save a good 4% against main for the PDF 1.7 spec.

I haven't had the patience to run the full performance check with the # 307 and # 311 sample files, but for the first "no arguments" test the PR saves 19% to 20% against main.

Summary of changes:

To facilitate refactoring, the static parseInternal method together with the related setObjectDescriptionFromInput method were moved to a new Parser class.

To reduce code duplication and improve readability new private warn methods were added to the Parser class.

To improve readability (and possibly improve performance), the olist, offset, contents_string and contents_offset stacks were combined into a single stack. To improve performance, both the new stack as well as the existing state stack were implemented as std::vectors rather than  as SparseOHArrays.

Treatment of null objects was tuned by recognizing that direct null objects are not be stored in QPDF_Arrays or QPDF_Dictionarys. Creation of such null objects, or setting descriptions/offsets for them, was eliminated.

Points to consider:

Should the new class be QPDFObjectHandle::Parser or QPDFParser? (my view: QPDFParser)

If QPDFObjectHandle::Parser, should the implementation be moved to QPDFObjectHandle_parser.cc? (my view: yes)

jberkenbilt:
> Treatment of null objects was tuned by recognizing that direct null objects are not be stored in QPDF_Arrays or QPDF_Dictionarys. Creation of such null objects, or setting descriptions/offsets for them, was eliminated.

There's a wording error above and I can't figure out what you mean. "are not be stored".

Looking at the code, it seems like you are still creating and storing the null objects, just without descriptions. Is this sufficient? The issue with #311 is both about time and memory usage. I don't have any way to measure memory usage other than to run something like top or htop and watch it. Sometimes I have artificially put a sleep somewhere at the end so I can catch how much memory is in use before exiting, but I don't have any automated benchmark for it.

Even if this doesn't fully address the millions of nulls problem, I think all the changes here are good. If it doesn't address the millions of nulls problem, maybe there is a way to use `std::vector<QPDFObjectHandle>` as olist but detect either when it starts to get too big or when we start to see too many consecutive nulls and then switch. As long as the overhead of detecting and switching (and managing which one is in use) is lower than the overhead of just using SparseOHArray off the bat, it could still be a win, and we'd only have to pay for SparseOHArray when we need it. For that matter, QPDF_Array could potentially do that. I haven't thought it through all the way.

I am in favor of a QPDFParser class. The fact that the parser is embedded in the QPDFObjectHandle class is like an itch...it's always bothered me but I've never gotten around to fixing it.

jberkenbilt:
Subject to my concerns stated above, I would accept this change.

jberkenbilt:
I think the QPDFParser class does not have to be in the public API, so its header would live in libqpdf/qpdf rather than include/qpdf. Do you agree?

fredemmott:
> There's #442 and I feel like there was another one where someone suggested use of a third-party library that has string builders that are more efficient, but I am not able to find that one at this moment.

Possibly https://github.com/qpdf/qpdf/discussions/715 , where I suggested https://github.com/fmtlib/fmt, or `std::format` when/if c++20 is permitted

jberkenbilt:
> Possibly #715 , where I suggested https://github.com/fmtlib/fmt, or `std::format` when/if c++20 is permitted

@fredemmott Yes, that's the one. Thanks. I forgot to check discussions. :-)

m-holger:
> There's a wording error above and I can't figure out what you mean.

Sorry about that. I have rewritten that and the following paragraph to something hopefully more comprehensible.

> I think the QPDFParser class does not have to be in the public API, so its header would live in libqpdf/qpdf rather than include/qpdf. Do you agree?

Agreed

I will finish of my work on parser for normal files before looking at how to best handle the 311 problem. At an earlier stage I dealt with it by changing olist to <object, count> tuples and simply incrementing the count for consecutive nulls (rather than pushing additional nulls onto the olist). That improved performance for arrays of nulls by ~ 35% compared with my "normal" version at the time, but wiped out 1/3rd of the improvement for "normal" files.

Thanks for your comments.

jberkenbilt:
> Sorry about that. I have rewritten that and the following paragraph to something hopefully more comprehensible.
> 
> >

I get it, so you're thinking that if a file has arrays with millions of nulls, an actual vector of that length will only exist transiently during parsing, and the vector will be smaller because of your changes? That may very well be adequate, in which case this is fine as is. I'll probably do a little of my own testing with it to convince myself. If I'm satisfied that the millions of nulls cases is handled well enough, is this ready to merge? Or do you want me to wait until you've had a chance to study the 311 issue yourself? I won't get to it today, but at this time, it seems like I will have time this week.

m-holger:
I'm doing the change QPDFObjectHandle::Parser -> QPDFParser at the moment. I'll let you know when I'm done. You can then judge whether the performance for 311 files is good enough.

I am working on two assumptions:
- if there was a serious memory problem it would be reflected in a deterioration of run-times, and
-  the proportion of files that have the 311 problem is very small and therefore, as long as there is some improvement for those files, there is no justification for degrading the performance for normal files in order to tune 311 performance. In the words of our pointy-eared friend, the needs of the many outweigh the needs of the few.

m-holger:
I did a very basic check on the version I just pushed up. I watched in system monitor while the first test of the performance check was running. For both the pdf 1.7 spec and the # 305 memory usage stayed in the same range of between 7.2 and 7.5GB. That's about as sophisticated a check as I can do without doing further research. As far as it goes, it looks ok to me.

If you are happy with the performance, the commit is ready for a full review and as far as I am concerned good to go.

If you find performance/memory problems then I'll do some more tuning.

fredemmott:
For linux, `mallinfo2()` may be useful; one way to get more detailed information than that would be to optionally use jemalloc instead, which offers [much more detailed](https://github.com/jemalloc/jemalloc/wiki/Use-Case%3A-Basic-Allocator-Statistics) statistics

jberkenbilt:
I think 7+GB is going to be too big. Thanks @fredemmott for the suggestions -- I'll look at those of my first impulse to roll something quick and dirty with /proc/self/statm doesn't pan out. @m-holger Would it help if I snuck a few minutes on this in front of #726? I landed my work project today and anticipate being able to dive into #726 (or this first) tomorrow or Thursday. Wednesday I'm in meetings all morning, so looking at this might be something I could fit into a day when it's harder to get focused time.

m-holger:
7.2GB was the starting position before running the test. The addition due  to running the test was bringing it up to 7.5GB, so the delta was up to say 0.4GB, which was similar to what I observed while running the test on the pdf 1.7 spec. But again, I don't have much expertise  in this area and it would be a good idea for you to take an independent look.

jberkenbilt:
@fredemmott Your suggestions led me eventually to malloc_info(3) which seemed to provide enough information for me to come up with a heuristic. Thanks for chiming in.

@m-holger Good news -- this branch as it currently stands does not adversely impact memory usage in my tests. On my system for the file from 305, there is an 8% reduction in time and a 0.1% increase in memory from your changes. I will do a final review and merge assuming no issues.

I will push up my change that adds a `--report-mem-usage` flag to qpdf and will incorporate this into performance testing before the release.

jberkenbilt (approved):
This is great. I can merge it immediately and then remove the no-longer-used include in a separate commit, or you can fix it. I'll give you a little time to respond.

jberkenbilt:
I'm just going to merge it -- it will make it easier for me to work on #726.

jberkenbilt:
@m-holger FYI, no action required. The following code creates a file that effectively exercises performance on the millions of nulls case. The resulting file uses 35 MB RAM and processes in about 1.4 seconds on my work laptop on current main. With qpdf 8.4.2, it takes 11 seconds and allocates 575 MB RAM. With 10.6.3, it uses about 31 MB RAM and takes about 1.7 seconds. I will work this into the regular test suite and into the performance test suite.

```c++
#include <qpdf/QPDF.hh>
#include <qpdf/QPDFWriter.hh>
#include <qpdf/QPDFObjectHandle.hh>

int main()
{
    QPDF q;
    q.emptyPDF();
    auto null = QPDFObjectHandle::newNull();
    auto top = "[]"_qpdf;
    for (int i = 0; i < 10; ++i) {
        auto inner = "[]"_qpdf;
        for (int j = 0; j < 100000; ++j) {
            inner.appendItem(null);
        }
        top.appendItem(inner);
    }
    q.getTrailer().replaceKey("/Trash", top);
    QPDFWriter w(q, "/tmp/a.pdf");
    w.setObjectStreamMode(qpdf_o_generate);
    w.write();
    return 0;
}
```

jberkenbilt:
The real one will be a little different but this is the gist. You'll see it when I push.

PR #731, opened 2022-07-16
Tidy QPDFObjGen related code

m-holger:
Change occurrences of e.g. "8,0" or "8/0" to "8 0".

If this PR is accepted #727 would need to be amended.

### DIFF
--- a/include/qpdf/QPDFObjGen.hh
+++ b/include/qpdf/QPDFObjGen.hh
@@ -47,13 +47,19 @@ class QPDFObjGen
     bool
     operator<(QPDFObjGen const& rhs) const
     {
-        return ((obj < rhs.obj) || ((obj == rhs.obj) && (gen < rhs.gen)));
+        return (obj < rhs.obj) || ((obj == rhs.obj) && (gen < rhs.gen));
     }
     QPDF_DLL
     bool
     operator==(QPDFObjGen const& rhs) const
     {
-        return ((obj == rhs.obj) && (gen == rhs.gen));
+        return (obj == rhs.obj) && (gen == rhs.gen);
+    }
+    QPDF_DLL
+    bool
+    operator!=(QPDFObjGen const& rhs) const
+    {
+        return (obj != rhs.obj) || (gen != rhs.gen);
     }
     QPDF_DLL
     int
--- a/include/qpdf/QPDFObjGen.hh
+++ b/include/qpdf/QPDFObjGen.hh
@@ -38,7 +38,7 @@ class QPDFObjGen
     {
     }
     QPDF_DLL
-    explicit QPDFObjGen(int obj, int gen = 0) :
+    explicit QPDFObjGen(int obj, int gen) :
         obj(obj),
         gen(gen)
     {
@@ -77,7 +77,7 @@ class QPDFObjGen
     bool
     isIndirect() const
     {
-        return obj > 0;
+        return obj != 0;
     }
     QPDF_DLL
     std::string unparse(char separator = ',') const;
--- a/include/qpdf/QPDFObjGen.hh
+++ b/include/qpdf/QPDFObjGen.hh
@@ -68,6 +68,12 @@ class QPDFObjGen
         return gen;
     }
     QPDF_DLL
+    bool
+    isIndirect() const
+    {
+        return obj > 0;
+    }
+    QPDF_DLL
     std::string unparse(char separator = ',') const;
     QPDF_DLL
     friend std::ostream& operator<<(std::ostream& os, const QPDFObjGen& og);
--- a/include/qpdf/QPDFObjGen.hh
+++ b/include/qpdf/QPDFObjGen.hh
@@ -38,7 +38,7 @@ class QPDFObjGen
     {
     }
     QPDF_DLL
-    QPDFObjGen(int obj, int gen) :
+    explicit QPDFObjGen(int obj, int gen = 0) :
         obj(obj),
         gen(gen)
     {
--- a/include/qpdf/QPDFObjGen.hh
+++ b/include/qpdf/QPDFObjGen.hh
@@ -23,7 +23,6 @@
 #define QPDFOBJGEN_HH
 
 #include <qpdf/DLL.h>
-#include <qpdf/QUtil.hh>
 #include <iostream>
 
 // This class represents an object ID and generation pair.  It is
@@ -69,18 +68,9 @@ class QPDFObjGen
         return gen;
     }
     QPDF_DLL
-    std::string
-    unparse() const
-    {
-        return QUtil::int_to_string(obj) + "," + QUtil::int_to_string(gen);
-    }
+    std::string unparse(char separator = ',') const;
     QPDF_DLL
-    friend std::ostream&
-    operator<<(std::ostream& os, const QPDFObjGen& og)
-    {
-        os << og.obj << "," << og.gen;
-        return os;
-    }
+    friend std::ostream& operator<<(std::ostream& os, const QPDFObjGen& og);
 
   private:
     // This class does not use the Members pattern to avoid a memory
--- a/include/qpdf/QPDFObjectHandle.hh
+++ b/include/qpdf/QPDFObjectHandle.hh
@@ -116,9 +116,17 @@ class QPDFObjectHandle
         // indicating whether it ran without errors.
         QPDF_DLL
         virtual void
-        provideStreamData(int objid, int generation, Pipeline* pipeline);
+        provideStreamData(QPDFObjGen const& og, Pipeline* pipeline);
         QPDF_DLL
         virtual bool provideStreamData(
+            QPDFObjGen const& og,
+            Pipeline* pipeline,
+            bool suppress_warnings,
+            bool will_retry);
+        // The following two overloads are deprecated.
+        QPDF_DLL virtual void
+        provideStreamData(int objid, int generation, Pipeline* pipeline);
+        QPDF_DLL virtual bool provideStreamData(
             int objid,
             int generation,
             Pipeline* pipeline,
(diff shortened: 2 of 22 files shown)


jberkenbilt:
I'm going to try catch up today on all the suggestions from the last few weeks.

jberkenbilt:
sorry, already merged #727. Let me look at this one.

jberkenbilt:
I will probably take this. Can you go ahead and fix the merge conflict? I almost suggested this idea when you added `getObjGenAsStr` but thought it would be a lot of hassle and a lot of changes. The inconsistency has been bothering me for some time, but I didn't want to break the output on `--check` and similar methods because people might rely on them in scripts. That said, qpdf 11 is full of big changes, and I think people who are relying on the output of `--check` would be better to switch to --json. One issue though is that the argument to `--show-object` is `o[,g]` which was the original reason that I had `o,g` in a bunch of places. In practice, I don't think this is a big deal because generation != 0 is virtually never seen, especially in modern PDF files. Object streams don't even allow generation != 0. I'll have to give it a more thorough review, but I'm inclined to feel that the benefits outweigh the disadvantages at this point.

m-holger:
I have not added any tests for the additions to QPDFObjGen as the code is indirectly tested to death.

jberkenbilt:
> I have not added any tests for the additions to QPDFObjGen as the code is indirectly tested to death.

concur

libqpdf/QPDF.cc
```diff
@@ -1837,17 +1834,14 @@ QPDF::readObjectAtOffset(
                 offset,
                 "object with ID 0");
         }
-
-        if ((exp_objid >= 0) &&
-            (!((objid == exp_objid) && (generation == exp_generation)))) {
+        if (exp_og.isIndirect() && (exp_og != og)) {
```
jberkenbilt, line 1818: There are a few changes here that I disagree with but that also expose a weakness of my original implementation. I'm going to push up a conflicting commit that cleans this up, then you can fix your commit. In particular, it is not correct to use isIndirect here -- see later comments plus the comment I have added to the code in my upcoming commit.

m-holger:
Is there a way to disable 'deprecated' warnings where I forward the method call to the legacy version of provideStreamData?

m-holger:
Is there a way to disable 'deprecated' warnings where I forward the method call to the legacy version of provideStreamData?

libqpdf/QPDF_linearization.cc
```diff
@@ -325,11 +325,10 @@ QPDF::readLinearizationData()
 QPDFObjectHandle
 QPDF::readHintStream(Pipeline& pl, qpdf_offset_t offset, size_t length)
 {
-    int obj;
-    int gen;
+    QPDFObjGen og;
     QPDFObjectHandle H = readObjectAtOffset(
-        false, offset, "linearization hint stream", -1, 0, obj, gen);
-    ObjCache& oc = this->m->obj_cache[QPDFObjGen(obj, gen)];
+        false, offset, "linearization hint stream", QPDFObjGen(-1), og);
```
jberkenbilt, line 330: After my upcoming commit, this will need to be an explicit `QPDFObjGen(0, 0)`. I'd rather have the explicit `0` than rely on the default -- see comments in `QPDF.cc` after my commit.

libqpdf/QPDF.cc
```diff
@@ -1055,12 +1047,11 @@ QPDF::read_xrefStream(qpdf_offset_t xref_offset)
...
     bool found = false;
     if (!this->m->ignore_xref_streams) {
-        int xobj;
-        int xgen;
+        QPDFObjGen x_og;
         QPDFObjectHandle xref_obj;
         try {
             xref_obj = readObjectAtOffset(
-                false, xref_offset, "xref stream", -1, 0, xobj, xgen);
+                false, xref_offset, "xref stream", QPDFObjGen(-1), x_og);
```
jberkenbilt, line 1054: This will need to be `QPDFObjGen(0, 0)` after my upcoming commit.

jberkenbilt (changes requested):
Mostly good -- a few minor updates will be required.

jberkenbilt:
> Is there a way to disable 'deprecated' warnings where I forward the method call to the legacy version of provideStreamData?

There is, but don't do it that way. Let the deprecated version forward to the new version. It should be trivial to remove deprecated methods in a future update. The commit that you are asking about came in after I started my review and I haven't reviewed it yet.

libqpdf/QPDFJob.cc
```diff
@@ -1010,7 +1011,7 @@ QPDFJob::doListAttachments(QPDF& pdf)
                 for (auto i2: efoh->getEmbeddedFileStreams().ditems()) {
                     auto efs = QPDFEFStreamObjectHelper(i2.second);
                     v << "    " << i2.first << " -> "
-                      << efs.getObjectHandle().getObjGen().unparse() << "\n";
+                      << efs.getObjectHandle().getObjGen().unparse(',') << "\n";
```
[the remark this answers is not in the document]
m-holger: Thought it was useful for readers of my age.

m-holger:
I don't think I can do that. qpdf is calling a user provided method and would need to forward to a user provide legacy method if the user hasn't updated his code (yet).

jberkenbilt:
> I don't think I can do that. qpdf is calling a user provided method and would need to forward to a user provide legacy method if the user hasn't updated his code (yet).

I guess I haven't seen the code you're talking about, but I'd rather not make users tidy their code. I'm fine with tidying interfaces on private methods and so forth, but I usually only make users change their code if it's either unavoidable or there's something harmful about the old way.

libqpdf/QPDF.cc
```diff
@@ -1858,21 +1852,15 @@ QPDF::readObjectAtOffset(
             }
         }
     } catch (QPDFExc& e) {
-        if ((exp_objid >= 0) && try_recovery) {
+        if (exp_og.isIndirect() && try_recovery) {
```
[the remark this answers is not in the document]
m-holger: I didn't think exp_objid was possible for reading from a file and so > 0 and >= 0 were effectively the same.

jberkenbilt:
@m-holger My tweak is up, so feel free to rebase and resolve comments. I can review the 11th commit before merge unless you tell me that there are still more coming.

libqpdf/QPDF.cc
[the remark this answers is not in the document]
jberkenbilt: Yes, that's correct. That's why I got rid of the `-1` sentinel.

include/qpdf/QPDFObjGen.hh
```diff
@@ -38,7 +38,7 @@ class QPDFObjGen
     {
     }
     QPDF_DLL
-    QPDFObjGen(int obj, int gen) :
+    explicit QPDFObjGen(int obj, int gen = 0) :
```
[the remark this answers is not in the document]
m-holger: Done

include/qpdf/QPDFObjGen.hh
```diff
@@ -68,6 +68,12 @@ class QPDFObjGen
         return gen;
     }
     QPDF_DLL
+    bool
+    isIndirect() const
+    {
+        return obj > 0;
```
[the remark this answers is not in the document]
m-holger: Done

libqpdf/QPDF.cc
m-holger: I used QPDFObjGen() throughout for the default case. Is that ok?

include/qpdf/QPDFObjectHandle.hh
```diff
@@ -116,9 +116,17 @@ class QPDFObjectHandle
...
         virtual void
-        provideStreamData(int objid, int generation, Pipeline* pipeline);
+        provideStreamData(QPDFObjGen const& og, Pipeline* pipeline);
         QPDF_DLL
         virtual bool provideStreamData(
+            QPDFObjGen const& og,
+            Pipeline* pipeline,
+            bool suppress_warnings,
+            bool will_retry);
+        // The following two overloads are deprecated.
```
jberkenbilt, line 126: Please remove comment -- I have no intention to deprecate these.

jberkenbilt (changes requested):

jberkenbilt:
I will do another full review when you give me the signal.

libqpdf/QPDF.cc
jberkenbilt: Yes, I have no objection to it for any other use than readObjectAtOffset because of the explicit use of 0 to mean something.

include/qpdf/QPDFObjectHandle.hh
m-holger: Done

m-holger:
This finishes this batch.

libqpdf/QPDF.cc
m-holger: Added explicit 0,0's for those calls.

m-holger:
I think I addressed all your comments. If I missed anything let me know. Otherwise, this finishes this PR.

include/qpdf/QPDFObjGen.hh
```diff
@@ -39,7 +38,7 @@ class QPDFObjGen
     {
     }
     QPDF_DLL
-    QPDFObjGen(int obj, int gen) :
+    explicit QPDFObjGen(int obj, int gen) :
```
jberkenbilt, line 41: Technically `explicit` is not useful here because the constructor takes two required args....

jberkenbilt (approved):

jberkenbilt:
Thanks for this contribution. I would have never gone to the trouble of cleaning up these interfaces, but I think this improves the quality of the code and lowers the barrier of entry for someone else who may want to understand or work with the library internals.

PR #734, opened 2022-07-26
Code tidy : replace 0 with nullptr or true

m-holger:


m-holger:
Note change in libqpdf/Pl_LZWDecoder.cc . I couldn't see a purpose in the existing construct but may be missing something.

This would be a candidate for a .gitignore entry.

jberkenbilt:
Thanks. I tend to replace 0 with nullptr when I'm modifying code, but I would have never taken the trouble to do this. I'm taking this now because I am about to make changes that will conflict with it.

m-holger:
No trouble thanks to clang-tidy --fix

PR #736, opened 2022-07-27
Code tidy: fix lgtm alert

m-holger:
…ntHelper

### DIFF
--- a/include/qpdf/QPDFAcroFormDocumentHelper.hh
+++ b/include/qpdf/QPDFAcroFormDocumentHelper.hh
@@ -159,7 +159,7 @@ class QPDFAcroFormDocumentHelper: public QPDFDocumentHelper
     // Return top-level form fields for a page.
     QPDF_DLL
     std::vector<QPDFFormFieldObjectHelper>
-        getFormFieldsForPage(QPDFPageObjectHelper);
+    getFormFieldsForPage(QPDFPageObjectHelper const&);
 
     // Return the terminal field that is associated with this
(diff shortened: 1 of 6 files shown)


m-holger:
I pushed this PR purely to run LGTM check.

m-holger:
The PR probably improves performance, but to a highly insignificant degree (of the order of less than 0.06%).

I have looked at a number of other instances where "good practice" suggests that parameters should be passed as const& rather than by value. A number of them showed a more significant impact - however, in none of the cases was the impact in the desired direction!

Contrary to what the lgtm analysis check run suggests, this commit fixes the existing alert.

jberkenbilt:
I appreciate the analysis. I'd rather not take this -- I don't want to add code that replaces default behavior with manual memory management (using a pointer for the iterator). I'll just drop the lgtm banner. Performance is important, but maintainability is more important except in highly critical areas of code or when performance is particularly bad.

jberkenbilt:
I'd consider a PR that only changes copies to const& in function calls without the changes to the iterator.

m-holger:
I do not think this PR has any merit other than managing lgtm analyses.

For performance it seems to me that the existing compilers do a very decent job and on balance are better at optimizing code than I am.

I'd suggest you simply close this PR.

Hopefully QPDFObjectHandle will shrink in the future and the problem will go away by itself.

As an aside, the more I learn about lgtm the less impressed I am, and I do not think their ratings are meaningful.

jberkenbilt:
Thanks again for your analysis.

PR #738, opened 2022-08-01
Inline QPDFObjectHandle::getObjGen etc

m-holger:
Also, make QPDFObjectHandle::isIndirect const.

In-lining the methods reduces runtime by somewhere between 1.5% and 2%.


jberkenbilt:
I didn't know you could put inline function definitions in a header outside of the class that declares them.

PR #739, opened 2022-08-01
Add QPDF::getObject to replace getObjectByObjGen and getObjectByID

m-holger:
For consistency with similar methods, e.g. replaceObject.


m-holger:
It might worth adding a comment why there are four versions of the method.

jberkenbilt:
Meh, I don't see a big problem with the old way, but I guess this is a cleaner interface, and it's a low-impact change.

PR #743, opened 2022-08-05
Code tidy: make conversion to bool explicit

m-holger:
Remove redundant '==' from conditions as suggested by C++ core guideline ES.87.

### DIFF
--- a/examples/pdf-attach-file.cc
+++ b/examples/pdf-attach-file.cc
@@ -75,7 +75,7 @@ process(
               << std::endl;
     auto fs = QPDFFileSpecObjectHelper::createFileSpec(q, key, attachment);
 
-    if (mimetype) {
+    if (mimetype != nullptr) {
         // Get an embedded file stream and set mimetype
         auto ef = QPDFEFStreamObjectHelper(fs.getEmbeddedFileStream());
         ef.setSubtype(mimetype);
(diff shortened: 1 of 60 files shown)


jberkenbilt:
I don't want to take this change. I prefer explicit `ptr == nullptr` as defense against C/C++'s coercion of non-boolean values to boolean. I believe this was actually part of a style guide I used earlier in my career. So, while I think the 1 -> true and 0 -> nullptr adds value, I think this actually removes value.

m-holger:
This is a valid point. However, to my mind it has two corollaries:

1. Given that the style is different from the c++ core guidelines, the deviation should be documented in the "contributing to qpdf" section of the manual.

2. The approach should be consistently applied  throughout the code, i.e. the test for an initialized  pointer should always be ptr != nullptr rather than just pointer.

m-holger:
Also, the same argument applies to other implicit conversions to bool

jberkenbilt:
I'm not taking this. I don't care about ES.87. I think this is a matter of personal preference, and I like the way I've done it. It's a bit of a mental burden for me to examine these and decide whether they help or hurt when there are so many things of higher value I can use my qpdf time for. If I eventually pass this on, the next person (or people) can decide if they want to do this.

PR #744, opened 2022-08-05
Code tidy: simplify boolean expressions

m-holger:
To me, the simplified expression seem more readable.


jberkenbilt:
I carefully write my boolean expressions in a way that helps me understand their intent. I don't really feel like taking any more of these big tidying changes. I like the nullptr one but even that one I was on the fence about. I think these cross over into the territory of gratuitous change to fix something that's not really a problem

PR #746, opened 2022-08-07
Code tidy: remove redundant calls to smart_ptrs get() method

m-holger:
Demo of discussion point in #726 only.

### DIFF
--- a/libqpdf/ClosedFileInputSource.cc
+++ b/libqpdf/ClosedFileInputSource.cc
@@ -23,7 +23,7 @@ ClosedFileInputSource::~ClosedFileInputSource()
 void
 ClosedFileInputSource::before()
 {
-    if (nullptr == this->m->fis.get()) {
+    if (nullptr == this->m->fis) {
         this->m->fis =
             std::make_shared<FileInputSource>(this->m->filename.c_str());
         this->m->fis->seek(this->m->offset, SEEK_SET);
--- a/libqpdf/JSON.cc
+++ b/libqpdf/JSON.cc
@@ -204,7 +204,7 @@ JSON::JSON_blob::write(Pipeline* p, size_t) const
 void
 JSON::write(Pipeline* p, size_t depth) const
 {
-    if (nullptr == this->m->value.get()) {
+    if (nullptr == this->m->value) {
         *p << "null";
     } else {
         this->m->value->write(p, depth);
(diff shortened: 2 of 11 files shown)


jberkenbilt:
I would actually accept this one. I'm sure this is a remnant of changing getPointer() to get(), and it's only a handful of places. Sorry, my brain is not governed by strict, predictable rules either.

jberkenbilt:
It also doesn't change the comparison against `nullptr` -- it does exactly one thing, which is to remove the `get` call.

m-holger:
That is why I prefer to let clang-tidy do the typing. It does one change, for a single purpose, using a consistent rule.
I find it much easier and faster checking clang-tidy output than my own changes, or any other code subject to human idiosyncrasies.

For the record, #734, #743 and #744 and #746 were all authored by clang-tidy and checked by me.

m-holger:
And each of them address a single readability rule.

Incidentally, I would suspect that lgtm apply similar checks as clang-tidy.

jberkenbilt:
I guess I agree with some but not all of what clang-tidy likes.

jberkenbilt:
In the future, if you create pull requests with clang-tidy, it would be helpful if you mention this and include the exact invocation in the commit comment. It's helpful when reviewing history. Knowing that these are automated makes me feel less guilty for rejecting some of them since they probably don't require that much effort.

m-holger:
Some of the readability suggestions of clang-tidy are to my mind counter-productive, and as mentioned in #736, some of their performance suggestions are also counter-productive. There is definitely a case for deciding which rule should and which rules shouldn't apply to qpdf.

However, this is an entirely different question to
- whether there should be rules, and
- whether they should be fixed for existing code.

My personal view would be that code consistency as good. Rules that can be fixed mechanically using tools should be fixed globally in a single dedicated commit. In other case, code should be locally consistent; e.g. a method (of reasonable size) should consistently either use or not use "this->".

jberkenbilt:
If there are going to be certain rules that we apply everywhere, then I would want to build them into `format-code` so that there isn't drift. If I were going to enforce a rule about `this->`, it would be to always use it on member variables and not to use it on method calls, but I'm not sure I want to enforce that rule. Sometimes I use `this->` with method calls for clarity, and sometimes use of `this->` is required to explicitly access a member variable/function that is shadowed by a local variable, particularly in constructors.

m-holger:
> In the future, if you create pull requests with clang-tidy

Will do



> makes me feel less guilty for rejecting some

No need to feel guilty. Some of my PRs primary purpose is to elicit qpdf conventions and standards, therefore accept or reject both produce the desired result.

jberkenbilt:
> Some of my PRs primary purpose is to elicit qpdf conventions and standards, therefore accept or reject both produce the desired result.

This is part of what makes you so good to collaborate with. A mutually egoless approach leads to a better eventual outcome. No other contributor has come anywhere close to this level of elevating qpdf's quality and usability.

jberkenbilt:
...or elevating my thinking about how qpdf should be maintained.

PR #747, opened 2022-08-08
Add optional parameter allow_nullptr to QPDFObjectHandle::getOwningQPDF

m-holger:
Also, inline method and add optional parameter error_msg.

### DIFF
--- a/libqpdf/QPDFObjectHandle.cc
+++ b/libqpdf/QPDFObjectHandle.cc
@@ -1207,14 +1207,6 @@ QPDFObjectHandle::getUniqueResourceName(
                            " QPDFObjectHandle::getUniqueResourceName");
 }
 
-// Indirect object accessors
-QPDF*
-QPDFObjectHandle::getOwningQPDF()
-{
-    // Will be null for direct objects
-    return this->qpdf;
-}
-
 // Dictionary mutators
 
 void
@@ -1634,16 +1626,15 @@ QPDFObjectHandle::coalesceContentStreams()
         // files may have pages that are invalid in other ways.
         return;
     }
-    QPDF* qpdf = getOwningQPDF();
-    if (qpdf == nullptr) {
-        // Should not be possible for a page object to not have an
-        // owning PDF unless it was manually constructed in some
-        // incorrect way. However, it can happen in a PDF file whose
-        // page structure is direct, which is against spec but still
-        // possible to hand construct, as in fuzz issue 27393.
-        throw std::runtime_error("coalesceContentStreams called on object"
-                                 " with no associated PDF file");
-    }
+    // Should not be possible for a page object to not have an
+    // owning PDF unless it was manually constructed in some
+    // incorrect way. However, it can happen in a PDF file whose
+    // page structure is direct, which is against spec but still
+    // possible to hand construct, as in fuzz issue 27393.
+    QPDF* qpdf = getOwningQPDF(
+        false,
+        "coalesceContentStreams called on object  with no associated PDF file");
+
     QPDFObjectHandle new_contents = newStream(qpdf);
     this->replaceKey("/Contents", new_contents);
 
--- a/include/qpdf/QPDFObjectHandle.hh
+++ b/include/qpdf/QPDFObjectHandle.hh
@@ -957,9 +957,11 @@ class QPDFObjectHandle
         std::set<std::string>* resource_names = nullptr);
 
     // Return the QPDF object that owns an indirect object.  Returns
-    // null for a direct object.
+    // null for a direct object if allow_nullptr is set to true or
+    // throws a runtime error otherwise.
     QPDF_DLL
-    QPDF* getOwningQPDF();
+    inline QPDF*
+    getOwningQPDF(bool allow_nullptr = true, std::string const& error_msg = "");
 
     // Create a shallow copy of an object as a direct object, but do not
     // traverse across indirect object boundaries. That means that,
@@ -1874,6 +1876,19 @@ inline bool
 QPDFObjectHandle::isInitialized() const
 {
     return initialized;
+}
+
+// Indirect object accessors
+inline QPDF*
+QPDFObjectHandle::getOwningQPDF(
+    bool allow_nullptr, std::string const& error_msg)
+{
+    // Will be null for direct objects
+    if (!allow_nullptr && (this->qpdf == nullptr)) {
+        throw std::runtime_error(
+            error_msg == "" ? "attempt to use a null qpdf object" : error_msg);
+    }
+    return this->qpdf;
 }
 
 #endif // QPDFOBJECTHANDLE_HH
--- a/libqpdf/QPDF.cc
+++ b/libqpdf/QPDF.cc
@@ -2266,7 +2266,7 @@ QPDF::copyForeignObject(QPDFObjectHandle foreign)
         throw std::logic_error(
             "QPDF::copyForeign called with direct object handle");
     }
-    QPDF* other = foreign.getOwningQPDF();
+    QPDF* other = foreign.getOwningQPDF(false);
     if (other == this) {
         QTC::TC("qpdf", "QPDF copyForeign not foreign");
         throw std::logic_error(
@@ -2456,11 +2456,9 @@ QPDF::copyStreamData(QPDFObjectHandle result, QPDFObjectHandle foreign)
     QPDFObjGen local_og(result.getObjGen());
     // Copy information from the foreign stream so we can pipe its
     // data later without keeping the original QPDF object around.
-    QPDF* foreign_stream_qpdf = foreign.getOwningQPDF();
-    if (!foreign_stream_qpdf) {
-        throw std::logic_error("unable to retrieve owning qpdf"
-                               " from foreign stream");
-    }
+    QPDF* foreign_stream_qpdf = foreign.getOwningQPDF(
+        false, "unable to retrieve owning qpdf from foreign stream");
+
     QPDF_Stream* stream = dynamic_cast<QPDF_Stream*>(
         QPDFObjectHandle::ObjAccessor::getObject(foreign).get());
     if (!stream) {
--- a/libqpdf/QPDFFormFieldObjectHelper.cc
+++ b/libqpdf/QPDFFormFieldObjectHelper.cc
@@ -362,13 +362,11 @@ QPDFFormFieldObjectHelper::setV(QPDFObjectHandle value, bool need_appearances)
         setFieldAttribute("/V", value);
     }
     if (need_appearances) {
-        QPDF* qpdf = this->oh.getOwningQPDF();
-        if (!qpdf) {
-            throw std::logic_error(
-                "QPDFFormFieldObjectHelper::setV called with"
-                " need_appearances = true on an object that is"
-                " not associated with an owning QPDF");
-        }
+        QPDF* qpdf = this->oh.getOwningQPDF(
+            false,
+            "QPDFFormFieldObjectHelper::setV called with need_appearances = "
+            "true on an object that is not associated with an owning QPDF");
+
         QPDFAcroFormDocumentHelper(*qpdf).setNeedAppearances(true);
     }
 }
@@ -883,7 +881,7 @@ QPDFFormFieldObjectHelper::generateTextAppearance(
         if (found_font_in_dr && resources.isDictionary()) {
             QTC::TC("qpdf", "QPDFFormFieldObjectHelper get font from /DR");
             if (resources.isIndirect()) {
-                resources = resources.getOwningQPDF()->makeIndirectObject(
+                resources = resources.getOwningQPDF(false)->makeIndirectObject(
                     resources.shallowCopy());
                 AS.getDict().replaceKey("/Resources", resources);
             }
--- a/libqpdf/QPDFJob.cc
+++ b/libqpdf/QPDFJob.cc
@@ -2161,7 +2161,7 @@ QPDFJob::doUnderOverlayForPage(
     std::map<unsigned long long, std::shared_ptr<QPDFAcroFormDocumentHelper>>
         afdh;
     auto make_afdh = [&](QPDFPageObjectHelper& ph) {
-        QPDF* q = ph.getObjectHandle().getOwningQPDF();
+        QPDF* q = ph.getObjectHandle().getOwningQPDF(false);
         return get_afdh_for_qpdf(afdh, q);
     };
     auto dest_afdh = make_afdh(dest_page);
@@ -2597,7 +2597,7 @@ static QPDFObjectHandle
 added_page(QPDF& pdf, QPDFObjectHandle page)
 {
     QPDFObjectHandle result = page;
-    if (page.getOwningQPDF() != &pdf) {
+    if (page.getOwningQPDF(false) != &pdf) {
         // Calling copyForeignObject on an object we already copied
         // will give us the already existing copy.
         result = pdf.copyForeignObject(page);
--- a/libqpdf/QPDFPageObjectHelper.cc
+++ b/libqpdf/QPDFPageObjectHelper.cc
@@ -432,7 +432,8 @@ QPDFPageObjectHelper::externalizeInlineImages(size_t min_size, bool shallow)
                 this->oh.replaceKey(
                     "/Contents",
                     QPDFObjectHandle::newStream(
-                        this->oh.getOwningQPDF(), b.getBufferSharedPointer()));
+                        this->oh.getOwningQPDF(false),
+                        b.getBufferSharedPointer()));
             }
         }
     } else {
@@ -683,11 +684,10 @@ QPDFPageObjectHelper::removeUnreferencedResources()
 QPDFPageObjectHelper
 QPDFPageObjectHelper::shallowCopyPage()
 {
-    QPDF* qpdf = this->oh.getOwningQPDF();
-    if (!qpdf) {
-        throw std::runtime_error("QPDFPageObjectHelper::shallowCopyPage"
-                                 " called with a direct object");
-    }
+    QPDF* qpdf = this->oh.getOwningQPDF(
+        false,
+        "QPDFPageObjectHelper::shallowCopyPage called with a direct object");
+
     QPDFObjectHandle new_page = this->oh.shallowCopy();
     return QPDFPageObjectHelper(qpdf->makeIndirectObject(new_page));
 }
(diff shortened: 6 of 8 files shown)


m-holger:
Inspired by a fuzz test failure.

FYI, reformat-code reordered QOC includes (excluded from commit.)

jbarlow83:
I think getOwningQPDF should ultimately be converted to return a shared_ptr which it would obtain by locking a weak_ptr to the owning QPDF. This would protect against dangling pointers for libqpdf users. pikepdf currently avoids this by inspecting every QPDFObjectHandle for pointers and manipulating the Python reference count on the corresponding pikepdf.Pdf (i.e. QPDF). weak_ptr would be a cleaner solution.

If so this is a helpful refactoring step in that direction.

I was going to propose this earlier but given the extensive refactoring m-holger is doing I decided to stay out to avoid conflicting changes.

m-holger:
@jbarlow83 
I am not sure it is possible to provide a shared_ptr given that the QPDF object itself is not dynamically allocated.
Would it help if  getOwningQPDF returned a nullptr if the QPDF object has gone out of scope / been destroyed?

jbarlow83:
Good point. (There might be a case for holding QPDF shared_ptr anyway - that's probably it would be done for a greenfield project. But it would be disruptive.)

It would suffice to return nullptr if the owner has gone out of scope although not I'm sure how that would work. Global list of QPDFs?

m-holger:
@jbarlow83 
I think there are easier solutions than a global list.

> I was going to propose this earlier but given the extensive refactoring m-holger is doing I decided to stay out to avoid conflicting changes.

On the other hand, this might make it a good time to share your thoughts.

jbarlow83:
Thinking about this some more, removing most raw pointers from the public API is eventually worth doing. Raw pointers have ambiguous ownership semantics and would be clarified by their removal. (For example in BufferInputSource's constructor, there's an a Buffer* and bool to signal ownership; changing this to a shared_ptr would make the ownership automatically correct.)

For getOwningQPDF specifically, I think in most cases it's not actually necessary to return the QPDF. A lot of the tests are e.g. "is this a direct object?", or "do these two objects have the same owner?". Most of the time we don't need to expose the actual owner. Maybe a proxy class could be used here instead (getOwningQPDFProxy?). I'd suggest deprecation first.

(My original concern was not the existence of getOwningQPDF, but that QPDFObjectHandles might attempt to access their owner even if the owner is invalid.)

The object cache and QPDFObject are probably the areas of QPDF I understand least. For the QPDFObject/Value change, it looks to me like it's heading in a positive direction, but you've both invested a lot more time in studying this. I'm a little unsure about the extra level of indirection (QPDFObjectHandle->Object->Value).

jberkenbilt:
I was actually thinking just this morning that we should remove all raw pointers from the public API. Maybe it's time to start planning this for qpdf 12 along with fixing the behavior of QPDFObjectHandle. I'll be able to focus on this in a week or two and will try to distill my thoughts.

m-holger:
> My original concern was not the existence of getOwningQPDF, but that QPDFObjectHandles might attempt to access their owner even if the owner is invalid.

This is one of the motivations behind the Object/Value split.

The fundamental problem is that a QPDF object cannot keep track of the object handles it owns, and therefore it cannot notify handles of changes. However, some of the information about objects is held at the object handle level, including the owning QPDF object.

As a result, there is no way for the QPDF object to notify object handles that it is being destroyed, or for object handles to know whether their reference to the QPDF object is still valid.

> I'm a little unsure about the extra level of indirection (QPDFObjectHandle->Object->Value).

The problem is not confined to the `qpdf` member of the object handle but also affects the `obj` member. The only way for the object handle to know whether the `obj` member still point to the correct object is to check with the QPDF object (keeping it's fingers crossed that it still exists). The extra level of indirection is the price to be paid for being able to do away with (the cost of) those checks. On balance, having the extra indirection produces a saving.

m-holger:
> I think getOwningQPDF should ultimately be converted to return a shared_ptr

> I am not sure it is possible to provide a shared_ptr given that the QPDF object itself is not dynamically allocated.

> There might be a case for holding QPDF shared_ptr anyway

When I worked on # 749 it occurred to me that in reality QPDF is a shared_ptr already. The "true" QPDF object is the QPDF::Members that it points to.  It should therefore be easy (and cheap) to return the owning QPDF object itself rather than a pointer to it (throwing an exception if the method is called on a direct object).

(I haven't fully thought this through - there may be a show-stopper that I've missed)

@jbarlow83 , would that make life easier for pikepdf?

jberkenbilt:
@m-holger That's an interesting idea. I had never thought of QPDF as something that would be used that way. As an experiment, I did a test build with this addition to QPDF.hh

```c++
    QPDF(QPDF const&) = delete;
    QPDF& operator=(QPDF const&) = delete;
```

and qpdf builds and passes its test suite, which proves that there is no code in the qpdf library itself that ever makes copies of QPDF objects. I also did this test (with an unmodified libqpdf):

```c++
#include <qpdf/QPDF.hh>
#include <qpdf/QPDFWriter.hh>

int main()
{
    QPDF q;
    q.emptyPDF();
    QPDF q2(q);
    QPDFWriter w(q2, "/tmp/a.pdf");
    w.write();
    return 0;
}
```

and this results in

```
terminate called after throwing an instance of 'std::logic_error'
  what():  QPDFObjectHandle from different QPDF found while writing.  Use QPDF::copyForeignObject to add objects from another file.
```

So that probably means people aren't copying QPDF objects around now even though they could. If we made getOwningPDF return a QPDF object and opened up the pattern of officially treating QPDF objects like shared pointers, this problem would probably go away since it results from the comparison of getOwningQPDF with the address of the writer's own QPDF. This could be resolved by implementing operator== that compares the address of the Members variable.

I have a few concerns, which are not necessarily show-stoppers:
* After seeing the results of some of your other optimizations, I had actually considered whether we could get a little more performance out of QPDF by moving some (but not all) of the contents of Members into the object itself. We'd keep Members for ABI flexibility but optimize by moving some stable and frequently accessed members out. But this discussion makes me realize that that changes the meaning of copying QPDF and wouldn't only be safe if I blocked copying of QPDF or only moved things out of Members that were themselves pointer types. "delete"ing the default copy constructor and operator= would seem like a good idea in this case.
* I could delete the copy constructor and operator= now, but this would be a non-compatible change and could break code. I had never considered whether it would work to make a copy of a QPDF object to keep it in scope for copyForeignObjects (though there are other mechanisms for this), so that change could break working code.
* While it is definitely true that QPDF is like a shared pointer, it's kind of by mistake, and I'm not sure it's really idiomatic C++. I have sometimes wondered whether my decision to do things this way very early on was a mistake. At the time, I was playing around with Python and Java and had wished for more of a pass-by-reference behavior in C++. There were no shared pointers in the standard library at the time. I had PointerHolder, but it didn't have weak references. I did that as kind of an experiment, of course never imaging that qpdf would ever succeed as an open source project. Anyway, it's too late to change that.

QPDFObjectHandle has always been intentionally like a shared pointer, as implied by its name. I suppose making it possible to treat QPDF that way may not be a big stretch. That said, I can't promise that there aren't places in the library that assume two QPDF objects are the same if and only if they have the same address. There definitely used to be assumptions like that, but at some point, I introduced a unique ID for all QPDF objects. QPDF has a good test suite, but there may be code that does this that isn't exercised in a code path that would involve working with copies of QPDF objects. Also, if anyone wrote code that treats QPDF address comparison as equality, that code would break in subtle ways, and there's no easy way to detect that.

Having written all this, I wonder whether it might not be a bad idea to actually disallow copying of (and assignment to) QPDF objects to go down the shared pointer route for getOwningQPDF, even knowing that it would cause a ripple effect since we would have to eliminate all regular pointers from the API. That could be an item for qpdf 12 as previously stated.

libqpdf/QPDFObjectHandle.cc
```diff
@@ -1634,16 +1626,15 @@ QPDFObjectHandle::coalesceContentStreams()
...
-        // incorrect way. However, it can happen in a PDF file whose
-        // page structure is direct, which is against spec but still
-        // possible to hand construct, as in fuzz issue 27393.
-        throw std::runtime_error("coalesceContentStreams called on object"
-                                 " with no associated PDF file");
-    }
+    QPDF* qpdf = getOwningQPDF(
+        false,
+        "coalesceContentStreams called on object  with no associated PDF file");
+    // Should not be possible for a page object to not have an
```
jberkenbilt, line 1632: Can you move the comment up above the call to getOwningQPDF?

jberkenbilt (changes requested):
Subject to my trivial change about moving the comment, I think this is a good change. I've made a note in my TODO about this ticket containing a discussion of removing raw pointers so I won't lose it after this is closed.

If `format-code` is reordering headers, no need to exclude that from the commit, unless you think it's doing it erroneously. I'm not sure why it would do it though. If you exclude it and it happens next time I run format-code, it'll still get in.

m-holger:
> which proves that there is no code in the qpdf library itself that ever makes copies of QPDF objects

Interesting. I actually wondered on how much QPDF::Members cost and removed it to find out. It turned out that  removing QPDF::Members actually increased the run time. I assumed that this was due to some copying going on, probably due to QPDF being passed by value somewhere.

> I had actually considered whether we could get a little more performance out of QPDF by moving some (but not all) of the contents of Members into the object itself. We'd keep Members for ABI flexibility but optimize by moving some stable and frequently accessed members out.

I had the same thought, but my experiment with QPDF::Members made me rather less enthusiastic. I suspect it may not be all that beneficial except for code that gets very heavily excercised.

Another thought that occurred to me and which I haven't had time to investigate or fully think through: are there situations where it would make sense to implement methods at the Members level where the data is  rather than directly at the owning object level?

libqpdf/QPDFObjectHandle.cc
m-holger: Just pushed a corrected version up

jbarlow83:
>@jbarlow83 , would that make life easier for pikepdf?

If getOwningQPDF became:
```
QPDF getOwningQPDF(QPDFObjectHandle h) { return QPDF(h->etc()->get_qpdf_members()); }
```

I'd end up with is two Python pikepdf.Pdf that relate to the same QPDF. The QPDF object would have the appropriate reference count. Currently the pikepdf assumes a 1:1 correspondence from pikepdf.Pdf to QPDF, and stores some Python-side data in the Python object that matches a C++ object, for example, a temporary input file if one is needed. That's an assumption I made for pikepdf, and it could be changed. I could use a custom InputSource for the temporary file case for example.

There could be other issues but that's the only one I can think of that I know I'd have to change. There's no corresponding issue for QPDFObjectHandle because I don't keep side channel information and there's always been an assumption that mulitple QPDFObjectHandles which are actually the same object can be floating around.

In terms of the design question, I think it's worth considering that in C++ a copy constructor can implement either a deep copy or a shallow copy, while shared_ptr<T> has explicit shallow copy semantics. STL also tends to use deep copies (e.g. std::string since C++11, vector). Not saying you can't have a shallow copy constructor, but it's a bit less explicit what it will do.

https://en.cppreference.com/w/cpp/memory/enable_shared_from_this might be relevant to this situation, and would allow recovery of the owning shared_ptr from parts of the code that only a this-> pointer.

> QPDF::Members performance

Considering cache, this result makes some sense. The memory footprint of a QPDF isn't that large and pimpl's are usually sequential allocated, so close to the rest of object. The penalty comes in when we are potentially making millions of an object type like QPDFObjectHandle.

jberkenbilt:
A few other notes.

`QPDF` objects have a unique ID. This is used when copying foreign objects to keep track of which QPDF an object came from without having to cache pointers, which can be reused. This was to solve the problem of two unrelated QPDF objects that live at different times having the same address, not the problem of two co-existing QPDF objects that are the same, but it might be useful in here somewhere.

I did a quick read through `enable_shared_from_this`. I don't think I want to use it because I don't want QPDF to have to have a base class, especially a public one.

One idea that comes to mind is to have something like a QPDF::Handle, which would essentially be a custom shared pointer specifically for QPDF. I write this fully aware that I could be just repeating old mistakes, but I'm pretty sure it will work and have negligible cost since getOwningQPDF is not used that often and there aren't that many QPDF objects.

With this change, we would deprecate getOwningQPDF and introduce QPDFObjectHandle::getQPDFHandle instead. See #751 for a possible implementation. I didn't include the QPDFObjectHandle changes yet, but they would be trivial.

If we go with that, I would add the allow_nullptr semantics to the new method in place of this PR.

jberkenbilt:
A flaw/limitation of my QPDF::Handle solution is that a std::shared_ptr<QPDF::Handle> created from a std::weak_ptr<QPDF::Handle> will contain an invalid pointer once the QPDF is destroyed, but I still think this is much better than where we are now and offers a reasonable solution. It's already not safe to use a single QPDF object simultaneously from multiple threads, so there shouldn't be any race conditions where one thread locks a handle and another thread destroys the underlying QPDF object. Anyway, if we go with #751, I would make sure all this is documented.

jberkenbilt:
Another thing that might be worth doing would be to explicitly deprecate copying/assigning to QPDF, which I think I could do by just declaring a `[[deprecated("...")]]` default copy constructor and assignment operator, but I haven't tried it and have to run to an appointment. This comment will ensure I don't forget about it.

jberkenbilt:
It also occurs to me that getOwningQPDF would trivially be able to detect if the owning QPDF had disappeared and could throw a logic_error if that happened. That way, old code would continue to work but be safer, throwing an exception rather than returning a pad pointer in the event of a QPDF that's gone bad. New code that called getQPDFHandle() instead could explicitly check. There should be no valid cases of `QPDFObjectHandle`s associated with a QPDF that's gone bad. If you want to keep it around after the QPDF is gone, then `makeDirect()` should be called.

jberkenbilt:
> Another thought that occurred to me and which I haven't had time to investigate or fully think through: are there situations where it would make sense to implement methods at the Members level where the data is rather than directly at the owning object level?

@m-holger I think I'd rather not add member functions to Members. I like the pattern of just having it be a container for member variables. My hunch is that if we had a situation where we'd gain performance by adding a method to Members, we'd do better to pull that field out of Members and stick it in the object. So far, there are no cases that I recall of objects with Members and any other member variables, but I'd be more open to that than adding member functions to Members.

jberkenbilt:
@m-holger I ran format-code and pushed the fixup of the header ordering in QTC.hh to main. I guess I just missed that earlier.

m-holger:
I had a quick look at #751, but at the moment I am getting stuck at working out the rationale for getQPDF returning a pointer rather than (a reference to) the actual QPDF object itself.

I don't think this change affects #726. Sticking with an ordinary pointer is the most straight forward and cheapest option (https://github.com/qpdf/qpdf/pull/726/commits/1e2437ddf848e23681ac02d180a50c7d00351b76 ensures the pointer is invalidated when the owning QPDF gets destroyed).

jberkenbilt:
> I had a quick look at #751, but at the moment I am getting stuck at working out the rationale for getQPDF returning a pointer rather than (a reference to) the actual QPDF object itself.

I don't like interfaces that return references to things that might be null pointers.

jberkenbilt:
> > I had a quick look at #751, but at the moment I am getting stuck at working out the rationale for getQPDF returning a pointer rather than (a reference to) the actual QPDF object itself.
> 
> I don't like interfaces that return references to things that might be null pointers.

But maybe I'm misunderstanding the concern.

jberkenbilt:
oops, clicked close by mistake.

m-holger:
To me, calling getQPDF when the QPDF has been destroyed is a logic error and should throw an exception.

I would have expected something like
```cpp
    QPDF q;
    // ...
    auto wp = q.getHandle();
    // ...  
     {
         if (wp->isValid()) {
             auto qpdf = wp->getQPDF();
         }
     }
    if (wp->isValid()) {
         // The original QPDF is still valid
    // }
```
Too me, the item returned by getQPDF should keep the QPDF  object alive until the user is finished with it. Having the returned item quietly turn into a nullptr does not feel safe.

What is the reason for disallowing copying of QPDF objects?

m-holger:
The more I think about it, the more confused I get. I think that when I observed that QPDF is already kind of a pointer I may have thrown a complete red herring into the ring.

jberkenbilt:
> To me, calling getQPDF when the QPDF has been destroyed is a logic error and should throw an exception.

The problem is that QPDF objects can be allocated on the stack. There's no way to control when they get destroyed. All we can do is notice it and not hand people an invalid pointer.

> What is the reason for disallowing copying of QPDF objects?

There is code in the library that assumes that there's a one-to-one match between the address of a QPDF object and whatever source that QPDF object was created from. That assumption could exist in other people's code as well since it's been true about qpdf since the beginning. Copying QPDF objects basically doesn't work right now, so there's not much chance that anyone's relying on being able to copy QPDF objects. Now that I've noticed this, I'd rather explicitly block copying QPDF objects than make it work and potentially introduce deep, hard-to-detect bugs in people's code because the previously valid assumption that you can meaningfully do an equality test on two `QPDF*`s is suddenly no longer true.

As I see it, the problem @jbarlow83 raised is that getOwningQPDF may return a bad pointer and that there's no way to ever know this. We can't fix with `std::shared_ptr` because it's common to allocate QPDF on the stack or for library users to allocate QPDF objects in whatever way they want. I don't think we can return actual QPDF objects without creating extremely subtle bugs in code that's out of our control. So we have to have a way to detect when a QPDF pointer goes bad, and this provides a fairly simple way to do it.

If I were starting from scratch today, I'd probably have a factory method that returns std::shared_ptr<QPDF> and have that be the thing people passed around. Then the semantics would be clear, and the qpdf library would be completely free of this objects-are-really-references pattern, but sadly, it's too late for that now.

My brain might be stuck in old thinking patterns though, so if I'm making a mistake in my thinking, please call me on it.

jberkenbilt:
> Too me, the item returned by getQPDF should keep the QPDF object alive until the user is finished with it. Having the returned item quietly turn into a nullptr does not feel safe.

This is actually the line I meant to quote at the beginning of my last comment. I was trying to say why there's no way to "keep the QPDF object alive".

jbarlow83:
Maybe it would be better to, for v11, deprecate stack-allocation of QPDF, introduce a shared_ptr factory to get everyone to migrate; and then for v12: drop stack allocation interface and use that time to fix getOwningQPDF and other issues?

Changing all `QPDF q;` to `auto qptr = QPDF::create(); auto &q = qptr.get();` is greppable and avoids changing `.` to `->`.

jberkenbilt:
> Maybe it would be better to, for v11, deprecate stack-allocation of QPDF, introduce a shared_ptr factory to get everyone to migrate; and then for v12: drop stack allocation interface and use that time to fix getOwningQPDF and other issues?
> 
> Changing all `QPDF q;` to `auto qptr = QPDF::create(); auto &q = qptr.get();` is greppable and avoids changing `.` to `->`.

That seems like a pretty big change to me. I'd want to think this through a little more, and I'm eager to get qpdf 11 out. I'd have to think through a good migration path and a way to make it possible for people to write code that works with older versions and newer versions. I'm not sure whether I'm prepared to deprecate stack allocation of QPDF objects.

jberkenbilt (approved):
I think this is ready to go now.

jberkenbilt:
@m-holger I am ready to merge this change. Can you think of a reason not to? The other issue we're discussing is not resolved, but it's not really related. Your change here is a good incremental step regardless of where we land.

m-holger:
I think the PR is good to go.

> As I see it, the problem @jbarlow83 raised is that getOwningQPDF may return a bad pointer and that there's no way to ever know this.

#726 solves that problem. 

> The problem is that QPDF objects can be allocated on the stack. There's no way to control when they get destroyed. 

That is why the user needs to test whether the QPDF is still  valid prior to calling getQPDF. Calling getQPDF without making this check beforehand to me is a logic error

> There is code in the library that assumes that there's a one-to-one match between the address of a QPDF object and whatever source that QPDF object was created from.

That to me makes the case for returning a QPDF& rather than QPDF* stronger.  Once copy/assignment has been outlawed it will allow callling  QPDF methods directly, but will not allow to easily store a pointer/ref to QPDF.

PR #749, opened 2022-08-18
Tune QPDFTokenizer

m-holger:
Inlining QUtil functions has approx half the effect as Removing Members. Combined improvement is around 1%

### DIFF
--- a/libqpdf/QPDFTokenizer.cc
+++ b/libqpdf/QPDFTokenizer.cc
@@ -134,13 +134,7 @@ QPDFTokenizer::isDelimiter(char ch)
 void
 QPDFTokenizer::resolveLiteral()
 {
-    if (QUtil::is_number(this->val.c_str())) {
-        if (this->val.find('.') != std::string::npos) {
-            this->type = tt_real;
-        } else {
-            this->type = tt_integer;
-        }
-    } else if ((this->val == "true") || (this->val == "false")) {
+    if ((this->val == "true") || (this->val == "false")) {
         this->type = tt_bool;
     } else if (this->val == "null") {
         this->type = tt_null;
@@ -205,6 +199,14 @@ QPDFTokenizer::handleCharacter(char ch)
         inName(ch);
         return;
 
+    case st_number:
+        inNumber(ch);
+        return;
+
+    case st_real:
+        inReal(ch);
+        return;
+
     case st_string_after_cr:
         inStringAfterCR(ch);
         return;
@@ -224,7 +226,6 @@ QPDFTokenizer::handleCharacter(char ch)
     case st_inline_image:
         inInlineImage(ch);
         return;
-        this->val += ch;
 
     case st_in_hexstring:
         inHexstring(ch);
@@ -240,6 +241,14 @@ QPDFTokenizer::handleCharacter(char ch)
 
     case st_name_hex2:
         inNameHex2(ch);
+        return;
+
+    case st_sign:
+        inSign(ch);
+        return;
+
+    case st_decimal:
+        inDecimal(ch);
         return;
 
     case (st_token_ready):
@@ -327,6 +336,31 @@ QPDFTokenizer::inTop(char ch)
 
     case '/':
         this->state = st_name;
+        this->val += ch;
+        return;
+
+    case '0':
+    case '1':
+    case '2':
+    case '3':
+    case '4':
+    case '5':
+    case '6':
+    case '7':
+    case '8':
+    case '9':
+        this->state = st_number;
+        this->val += ch;
+        return;
+
+    case '+':
+    case '-':
+        this->state = st_sign;
+        this->val += ch;
+        return;
+
+    case '.':
+        this->state = st_decimal;
         this->val += ch;
         return;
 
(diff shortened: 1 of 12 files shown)


jberkenbilt:
I don't want to inline QUtil functions since it removes them from the binary interface.

m-holger:
I am getting

```
$ ./check_abi dump --lib build/libqpdf/libqpdf.so | grep "QUtil"
...
QUtil::is_digit(char)
...
QUtil::is_hex_digit(char)
...
QUtil::is_number(char const*)
...
```
with the in-lined functions.

I hope this means that the functions are part of the ABI otherwise some of the changes to QPDFObjGen and QPDFObjectHandle will need further thought.

jberkenbilt:
Okay, maybe inlining is fine anyway. I'm on the road and not really focused, so I'll save further comment. I remember previously deciding that the ABI impact if inlining is not a big deal for stable C++ stuff since people are not calling it though runtime loading of DLL and those tiny functions aren't going to need bug fixes that people should get for free without recompiling.

jberkenbilt:
So those are probably fine to inline.

m-holger:
Treating val and raw_val more like buffers rather than strings brings the combined saving to around 2.5%

m-holger:
I just pushed some refactoring of QPDFTokenizer. My main purpose was to work out what the code actually does. However, it turns out that my changes reduce runtime by some 2.5%, bringing the total savings compared to main to about 5%. 

My main approach was:

- Reduce the complexity of code with the aim to have one outer switch statement controlled by the current state, with inner switch statements or chained if statements controlled by the incoming char, followed by breaking the state machine up into individual methods for each state. 

- Reduce the complexity of each state by eliminating flags and introducing additional states instead. 

- Remove the need to unread chars within tokens by calling the new state from the old state where necessary.

- Given that tokens are processed one character at a time, eliminate the second pass for e.g names or numbers.

m-holger:
Tuning FileInputSource saves another ~7.5%

jberkenbilt:
How close is this to being ready to merge? I'll want to give it heavy scrutiny, but at a quick glance, the changes look good, and some of them are similar to ideas I've had floating around in my head for some time. This is a great contribution and not something I would have likely gotten around to even though I've wanted to do it. You probably also did a better job with it than I would have.

I may not have time again this weekend (a family wedding) but I expect I am not more than a week away from being able to put a focused chunk of time into all these open pull requests.

m-holger:
The last commit provides ~1.3% saving. It is probably more of a problem statement than a full solution.

As far as push mode is concerned, I am finished with QPDFTokenizer.

For InputSources, I have a couple of commits ready to move the data members BufferInputSource and CloseFileInputSource from the Member class. That completes the obvious and easy hits.

I am planning to look at the pull mode of QPDFTokenizer, but that does not affect the push mode code.

libqpdf/QPDFTokenizer.cc
```diff
@@ -138,7 +137,7 @@ QPDFTokenizer::presentCharacter(char ch)
 {
     handleCharacter(ch);
 
-    if (this->in_token && !this->unread_char) {
+    if (this->in_token) { //} && !this->unread_char) {
```
jberkenbilt, line 140: oops -- stray comment

libqpdf/QPDFTokenizer.cc
```diff
@@ -139,376 +133,719 @@ QPDFTokenizer::isDelimiter(char ch)
...
 {
-    if (this->m->state == st_token_ready) {
-        throw std::logic_error(
-            "INTERNAL ERROR: QPDF tokenizer presented character "
-            "while token is waiting");
-    }
-
-    char orig_ch = ch;
-
     // State machine is implemented such that some characters may be
```
jberkenbilt, line 148: I believe this comment is no longer true.

jberkenbilt (approved):
This is fabulous.

jberkenbilt:
The new implementation is not only more efficient, but easier to read and understand. Thanks for breaking it up commit by commit to make it easier to review. I found that reviewing side by side with whitespace disabled was quite doable. There were a few changes that I didn't completely see at a glance how they were safe, particularly around unreading characters. That said, I have high faith in the test suite for the tokenizer and in the care you've taken in this change.

Premature optimization is the root of all evil. This is probably one of the best examples of appropriate optimization I've seen. The tokenizer has been very stable for a long time, and it is not likely that it needs to be changed for any reason other than performance optimization. Certain things such as lots of return statements, long case statements, and special functions that have to be called in particular ways can be problematic in early implementations, but they are power tools for optimization. Many of the optimizations I had in my head, such as reducing string copies, are here, but yours go quite a bit farther. The astonishing thing is that the code is also simpler by the way you refactored things and broken them into individual functions, most of which are simple case statements. The code looks a little more like how a scanner generate would do it. A while ago, I did a full rewrite of the tokenizer using flex, but it turned out that using it to parse strings and so forth was quite complex (since flex isn't really designed for that) and ultimately resulted in more complexity rather than less. So this is really a great solution for simplifying and optimizing the tokenizer.

I'd be happy to take this as is and fix the two comment nits in a separate commit, but I'll pause a little before doing that in case you feel like fixing them or commenting on anything.

Thanks for your tremendous work on revitalizing some of this code.

jberkenbilt:
I see new commits, so I can pause a little more, or you can let me know what a mergeable increment is.

m-holger:
I addressed your comments in the final commit.

The previous two commits should be quite straightforward, but on the other hand are self contained. You can merge with or without them.

Done for now.

jberkenbilt:
Thanks again. I'll be offline in about 20 minutes for the rest of the day. I'll merge before I leave as long as it's far enough through CI. I pulled your changes locally, and they pass, so I'm not worried about waiting for CI to finish for the new commits.

PR #751, opened 2022-08-29
QPDF::Handle

jberkenbilt:
@m-holger @jbarlow83 Here's a crack at my QPDF::Handle proposal from #747.

### DIFF
--- a/include/qpdf/QPDF.hh
+++ b/include/qpdf/QPDF.hh
@@ -63,6 +63,16 @@ class QPDF
     QPDF_DLL
     ~QPDF();
 
+    // Copying QPDF objects has never been blocked by the API but has
+    // never been safe. If you have been copying QPDF objects around,
+    // please switch to using std::shared_ptr or use getHandle(),
+    // added in qpdf 11. Copying and assignment of QPDF objects will
+    // be removed in QPDF 12.
+    [[deprecated("use shared_ptr or getHandle instead")]] QPDF(QPDF const&) =
+        default;
+    [[deprecated("use shared_ptr or getHandle instead")]] QPDF&
+    operator=(QPDF const&) = default;
+
     // If you need to keep a raw (non-shared) pointer of a QPDF object
     // around and need to protect against its having become invalid,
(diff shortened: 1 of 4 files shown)


jberkenbilt:
There is a big discussion in #747 about the issue this change is attempting to address, but #726 may make the problem go away. I will revisit or close this after completing my study and hopefully acceptance of #726.

m-holger:
#726 addresses the immediate issue raised in #747.

I think this PR goes further and will remain relevant even if #726 is accepted, although I think this solution is more appropriate for the public API than for internal use, especially in performance critical code.

A related but separate issue that will need to be addressed is: what should happen to object handles / objects when the owning QPDF objects gets destroyed?

jberkenbilt:
I'm going to give serious thought to @jbarlow83 's suggestion to disallow stack-allocated QPDF objects and other QPDF objects that are not shared pointers. If I do that, I would introduce something similar to POINTERHOLDER_TRANSITION to make it easier to migrate code and create code that is backward-compatible. This would be for after qpdf 11. If I do that, then I think we can just use shared_ptr or weak_ptr in most locations where we currently use QPDF*, and for those cases that must use QPDF*, there can be a way to get a weak pointer to self without having to use public inheritence (by just having the factory method stash a weak pointer in the object). I'll get back to this after #726 is accepted (in current or modified form).

m-holger:
Just another random thought that I haven't fully thought through - in a sense, a pdf document is just another container type like QPDF_Array or QPDF_Dictionary.

Being able to unify the concepts might have benefits when dealing with the issues around ownership of direct objects. 

But that is definitely not for qpdf 11.

jberkenbilt:
My QPDF::Handle thought was kind of an analogy to QPDFObjectHandle with QPDF being more like QPDFObject. Lots of stuff to think through.

jberkenbilt:
I don't think this change is needed anymore after #726. I'm going to close this in favor of #726's solution of just clearing `qpdf` in QPDFValue. If I'm going to force QPDF objects to only exist in shared pointers, then I don't think I would want or need this.

PR #754, opened 2022-09-01
Make object private

jberkenbilt:
@m-holger This makes `QPDFObject` private. For your information and review.


jberkenbilt:
I've added an additional commit that disables copying/assigning to QPDF and adds the ``QPDF::create()`` factory method. @m-holger, @jbarlow83 I've looked through the QPDF public API as well as some internals at all uses of `QPDF*`, and I'm thinking it's probably not worth removing raw pointers from the API. There are quite a few of them, but the only places where they are saved (and not just used transiently by the method) are `QPDFValue` and `QPDF_Stream`, and the only place where they are returned that I can see is `getOwningQPDF()`. With #726, we are now clearing the pointer when the owning `QPDF` goes out of scope, which I think greatly improves the situation.

As an experiment, I temporarily blocked explicit creation of `QPDF` objects to see what would happen, and it's actually not that big a deal in the QPDF code itself, but I think it would probably cause headaches for people, and it would be very hard to do it in a way that would make it easy for people to create backward-compatible code. The reason is that if I make `QPDF::QPDF()` private, then you have to use `QPDF::create()` to create `QPDF` objects as you won't be able to call `std::make_shared<QPDF>()`. This means that I would want to deprecate the constructor for a while before disabling it, which means we wouldn't be able to start benefiting from always having a shared pointer until qpdf 12. I'd probably end up doing something similar to `POINTERHOLDER_TRANSITION`. It would be something like this:

* `QPDF_POINTER_TRANSITION = 0` -- no warnings around calling the QPDF constructor
* `QPDF_POINTER_TRANSITION = 1` -- calls to `QPDF()` are deprecated, but QPDF is still available so code can be backward compatible and use `std::make_shared<QPDF>`
* `QPDF_POINTER_TRANSITION = 2` -- the QPDF constructor is private; all calls to `std::make_shared<QPDF>` have to be replaced with `QPDF::create`

Then, in addition to having to define `POINTERHOLDER_TRANSITION`, people would have to define `QPDF_POINTER_TRANSITION`, which is a bit of a nuisance. Also, changing all calls that pass `QPDF*` around to pass a weak pointer around would be toil as it would require updating a lot of code.

So the question is whether it's worth it. If we were to do this, we'd have to look at each use of `QPDF*` in the interface and decide whether to use a `std::shared_ptr` or a `std::weak_ptr`. The answer would almost always be to use a `std::weak_ptr`, which means we'd have to take the extra step of calling `lock()`. There would be a temptation to use `std::shared_ptr`, but this opens us up to a new category of errors from creating circular references or otherwise preventing `QPDF` objects from getting destroyed. For `QPDFValue`, we have `releaseResolved`, but I don't think there ever _has_ to be a circular `QPDF` reference, so it would be better to just rely on the sanitizer tests in CI. After we do all this, the only benefit I really see is that `getOwningQPDF` could return a weak pointer, and we wouldn't actually have to have the one-line in `~QPDF` that calls `resetObjGen`. But we'd still need to call `releaseResolved`, so it has a very small impact on code quality. Alternatively, we could leave plain `QPDF*` in the interface but have `QPDF::create()` stash a `std::weak_ptr<QPDF>` pointing to itself in `QPDF::Members`. This would make it possible to get a valid weak pointer to self from a raw pointer without having to use public inheritance on whatever the thing was that let you get a weak reference to self. But then we still have raw pointers in the API.

So, while I do see some benefit and would probably do it this way if I were starting over again, I'm not sure we get enough benefit from removing raw pointers to be worth the headache, especially since the one place where the API was potentially returning an invalid pointer has now been fixed thanks to @m-holger's refactoring.

For now, I'm going to take removal of raw pointers off the qpdf 12 list, but I'm open to being convinced otherwise. Comments welcome.

jberkenbilt:
I may merge this as I'd like to push some changes that are on top of it. If you (@m-holger) find any issues, I can address afterwards. No rush -- I know you have other things to do besides work on qpdf. :-)

include/qpdf/QPDFObjectHandle.hh
```diff
@@ -1852,12 +1852,6 @@ class QPDFObjectHandle::QPDFArrayItems
     QPDFObjectHandle oh;
 };
 
-inline QPDFObjGen
-QPDFObjectHandle::getObjGen() const
-{
-    return isInitialized() ? obj->getObjGen() : QPDFObjGen();
-}
-
 inline int
```
m-holger, line 1855: Given that we can't inline getObjGen, it probably doesn't make sense to inline getObjID , getGeneration and isIndirect either. (If my understanding is correct, either way we end up with one function call, and if the methods are not inline, marginally fewer instructions at the point of call.)

jberkenbilt: If you want to try it and see, feel free, and I can merge when I'm back. Otherwise I'll try tomorrow or Tuesday.

jberkenbilt: On my system, removing these inlines actually results in a measurable performance decrease, so I think leaving them as they are is probably better. This may be dependent on the compiler or flags, but the way I build it is similar to the way it's built by most distributions, so I think I'll leave it alone.

PR #755, opened 2022-09-04
Fix commit  805c1ad : Reset QPDFValue::qpdf and QPDFValue::og when ...

m-holger:
On destruction of the QPDF object replace all indirect object references
with direct nulls.

Remove all existing code to release resolved references.

Fixes performance issue due to interaction of resetting QPDFValue::qpdf and
og members and prior code.


jberkenbilt:
Yes, of course. Make perfect sense but hadn't occurred to me.

PR #756, opened 2022-09-04
Use unique_ptr and move constructor for Buffer::Members

jbarlow83:
With the recent commit to privatize the copy constructor in QPDF, it's not longer possible for a QPDF::Members object to be shared, so the member can turned into a unique_ptr. Seems more clear than shared_ptr which implies there are other owners.

For Buffer has always implemented deep copies, so again the Buffer::Members can be converted to a unique_ptr. At the same time I set up move constructors, which cannot be automatically generated in this case due to the custom copy constructors. When compiler decides the move constructor can be used, it will now be able to avoid buffer copies in cases where it would previously be forced to duplicate the buffer and then delete the original.

I don't believe any other classes' ::Members can be converted to unique_ptr at the moment since their implicit copy constructors rely on shared_ptr, but it's quite possible there are some instances.

### DIFF
--- a/include/qpdf/Buffer.hh
+++ b/include/qpdf/Buffer.hh
@@ -49,6 +49,10 @@ class Buffer
     QPDF_DLL
     Buffer& operator=(Buffer const&);
     QPDF_DLL
+    Buffer(Buffer &&) noexcept;
+    QPDF_DLL
+    Buffer& operator=(Buffer &&) noexcept;
+    QPDF_DLL
     size_t getSize() const;
     QPDF_DLL
     unsigned char const* getBuffer() const;
@@ -75,7 +79,7 @@ class Buffer
 
     void copy(Buffer const&);
 
-    std::shared_ptr<Members> m;
+    std::unique_ptr<Members> m;
 };
 
 #endif // BUFFER_HH
--- a/libqpdf/Buffer.cc
+++ b/libqpdf/Buffer.cc
@@ -48,12 +48,24 @@ Buffer::operator=(Buffer const& rhs)
     return *this;
 }
 
+Buffer::Buffer(Buffer&& rhs) noexcept :
+    m(std::move(rhs.m))
+{
+}
+
+Buffer&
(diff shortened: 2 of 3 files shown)


jberkenbilt:
For QPDF, what would be the advantages to switching to `std::unique_ptr` other than a slight gain in efficiency and a little more clarity in intention? In the absence of a small reason, I'm inclined not to switch from `std::shared_ptr` to `std::unique_ptr` for two small reasons:
* If someone has a major problem with disabling QPDF copy and assignment and I decide to deprecate rather than removing that functionality, leaving it as shared would mean that I could just put those back in and increase the minor version. Otherwise, it's ABI-breaking to put those back in because I'd have to change the pointer type. That said, it's very unlikely that I would do this since copying QPDF objects has never been safe and throwing them into shared pointers has always worked.
* It's a little easier to reason about the code when you just follow the `std::shared_ptr` members pattern. That said, it's already the case, and even moreso in qpdf 11, that some classes don't use this pattern.

For Buffer, the ability to do a move seems pretty compelling. I'm not sure whether there are places in the code now that do this. I think I have generally tried to stick with pointers to buffers, but I'd have to dig through the code to be sure.

Would this change make a difference for pikepdf? If it solves some problem, I'd be more inclined to accept it. Right now, I'm on the fence, but overall, I think it seems like a positive change and may decide to take it even with my small misgivings.

jberkenbilt:
I think I want the Buffer change. I'm still undecided about the QPDF change, but I'll wait to decide until I hear your thoughts @jbarlow83 and will probably take the change unless you think that particular one really doesn't matter.

jbarlow83:
I dropped the QPDF change. I think it provides more clarity but it's not that big of a deal, perhaps not worth it compared to ABI issues if you're thinking of adjusting copying in the future.

I don't believe pikepdf currently tries to copy any Buffers.

jberkenbilt:
You're probably right about clarity. We can take a fresh look later. I'll accept this one for now. Thanks!

PR #757, opened 2022-09-05
Value proxy

jberkenbilt:
@m-holger Please review. This PR does two things:
* Rename `QPDFObject` to `QPDFValueProxy`. I'm open to suggestions if this name is awful.
* Update design.rst to reflect the object/value split.

The main rationale for renaming `QPDFObject` was to make it possible to remove it from the public API while making the transition from `QPDFObject::object_type_e` to `qpdf_object_type_e` smoother and also to avoid pitfalls from yanking a header file. The yanking a header file problem is not hypothetical. My first attempt at removing `QPDFObject.hh` actually passed for me locally but failed in CI because, on my local system, /usr/include/qpdf/QPDFObject.hh still existed, so even though I had my source tree first in the include path, everything still built with some stray references to `QPDFObject.hh` that I had overlooked. Unfortunately, this is very common. It happens with some regularity that people run into trouble because they have multiple versions of qpdf on their systems. C-based languages suffer a lot from the lack of modules and real dependencies, though I believe this is partially addressed in upcoming enhancements to the standard library. Oh well...nothing we can do about that.

I'm particularly interested in your input on whether `QPDFValueProxy` feels okay to you as a name for what used to be `QPDFObject` and, more importantly, whether you think my documentation updates are clear and accurately reflect your changes.


jberkenbilt:
You probably know this, but the docs as built from here can be found at https://qpdf--757.org.readthedocs.build/en/757/

The sections I rewrote were
* https://qpdf--757.org.readthedocs.build/en/757/design.html#design-goals
* https://qpdf--757.org.readthedocs.build/en/757/design.html#implementation-notes

m-holger:
While at the programming level object is a proxy for a value object, to my mind it is more than that. To me, the current object represents a location in the object hierarchy of a pdf document, while value represents the concrete object at that location. 

One possibility for the future would be that "object" carries information about what types of value is acceptable at that point in the object hierarchy.

I do not feel that value-proxy properly reflects that. At the same time I don't think that object is the right name - to me it conjures up association with low-level syntax objects like integer or array. 

Unfortunately I haven't found a good alternative name that clearly expresses that role.

m-holger:
The rationale for renaming QPDFObject makes sense to me.

I am reviewing the documentation updates.

jberkenbilt:
> Unfortunately I haven't found a good alternative name that clearly expresses that role.

Yes, I agree with all your points including, unfortunately, this last one. I guess we can keep QPDFValueProxy for now. I think the documentation explains its purpose.

manual/design.rst
```diff
@@ -230,43 +233,46 @@ could serve as a starting point to someone trying to understand the
...
-Clients do not interact with Objects directly but instead interact with
-``QPDFObjectHandle``.
+In a PDF file, objects may be direct or indirect. Direct objects are
+objects whose representations appear directly in PDF syntax. Indirect
+objects are references to objects by their ID. The qpdf library uses
+the ``QPDFObjectHandle`` type to hold onto objects and to abstract
+away in most cases whether the object is direct or indirect.
+
+Internally, ``QPDFObjectHandle`` holds onto a shared pointer to the
+underlying object value. When direct object is created, the
```
m-holger, line 243: When direct object -> When a direct object

m-holger (commented):
@jberkenbilt 
I don't have anything that I disagree with in the documentation.
I marked one typo that I spotted, but I did not proof-read.
Needs to be updated for the change back to QPDFObject and for QPDF_Destroyed.

manual/design.rst
jberkenbilt: Thanks -- fixed typo. I already made the other updates.

PR #759, opened 2022-09-07
Make BufferInputSource internals available to InputSource::fast... methods

m-holger:


jberkenbilt:
I would consider this for qpdf 11, which would enable us to potentially migrate other InputSource implementations to use it. Rather than `fast_enabled`, I'd probably call it something like `updates_position` and have it be an argument (possibly optional) to the InputSource constructor. The documentation would explain that an implementation of InputSource can set that to true if it ensures that the protected cur_offset and max_offset will always be set by the derived class to accurate values, thus enabling InputSource to use a faster implementation of seek and tell. Basically the PR as you have it with some slight tweaks.

m-holger:
@jberkenbilt 

I got the same CI failure  when I synced my fork:

Theme error:
An error happened in rendering the page acknowledgement.
Reason: ValueError('too many values to unpack (expected 3)')
make[3]: *** [manual/CMakeFiles/doc_html.dir/build.make:93: manual/html/index.html] Error 2

It looks like sphinx has been update from sphinx-5.2.0 to sphinx-5.2.0.post0 since you last committed to qpdf/qpdf.

There is also an error message:

ERROR: sphinx-rtd-theme 1.0.0 has requirement docutils<0.18, but you'll have docutils 0.19 which is incompatible.
 
This message was also present on the last successful commit.

m-holger:
File "/home/docs/checkouts/readthedocs.org/user_builds/qpdf/envs/759/lib/python3.9/site-packages/sphinx_rtd_theme/layout.html", line 13, in top-level template code
    {%- set (_ver_major, _ver_minor, _ver_bugfix) = sphinx_version.split('.') | map('int') -%}
ValueError: too many values to unpack (expected 3)

jberkenbilt:
The doc issue seems to have been already fixed. I triggered a rebuild of the failed actions job and will disregard the failure on the readthedocs, which didn't get rerun. Next time you push, I suspect that part will be fine.

m-holger:
With the switch to  C++17, this is probably not the right way to go.

PR #760, opened 2022-09-07
Fix issues around QPDF destruction

jberkenbilt:
@m-holger Does this work for you? I'm looking at the other issue of clearing owning_qpdf for nested objects.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2022-09-08  Jay Berkenbilt  <ejb@ql.org>
+
+        * Add QPDFObjectHandle::isDestroyed() to test whether an indirect
+        object was from a QPDF that has been destroyed.
+
 2022-09-07  Jay Berkenbilt  <ejb@ql.org>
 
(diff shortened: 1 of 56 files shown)


jberkenbilt:
@m-holger I believe this addresses the issues we have been discussing as far as we can go without breaking things that are relying on direct objects being able to have multiple owners.

jberkenbilt:
I'm still tweaking this. I gave it some more thought, and while I believe what is here is correct, I'm going to change some comments and change reset() to disconnect(). I have reached the conclusion that it should be fine to put a direct object from one file in another. The issue of objects having multiple parents is unrelated to this problem. I'm going to clean up my notes on this.

jberkenbilt:
Unless my head people think of something else, I'm done with this change.

jberkenbilt:
@m-holger I'm going to go ahead and merge this so I can push other changes behind it. Your comments are still welcome though feel no obligation. We've discussed all the changes here, so this is mainly for your information and review in case I missed something.

PR #764, opened 2022-09-11
Add default move constructor/assignment to QPDFObjectHandle

m-holger:


jberkenbilt:
Since the only member of QPDFObjectHandle is a single shared pointer, I'm not sure how much difference this would make. It would also require bumping to 11.1, which is not necessarily a bad thing, but I wouldn't want to do it just for this I think. We can hold this.

m-holger:
Just under 1%. It's passed a lot by value requiring updating the usage count when copying rather than moving.

m-holger:
No problem holding it in the pipeline.

jberkenbilt:
I had to revert this. It caused check_abi to fail because a symbol disappeared from the shared library (not sure whether it matters but still). Also, and more importantly, with this commit, there are many failures in pikepdf's test suite with errors about accessing a destroyed QPDF object. Reverting this commit solves both problems, so I think this will have to stay out for the foreseeable future.

jberkenbilt:
The symbol that disappeared was some LockCount symbol in std:: that didn't seem to be related directly to qpdf. At first I added a commit to exclude it from the ABI check but after I realized this commit was causing other problems, I reverted that since I didn't really understand the implications. Looking at the pikepdf failures, there were lots of failures where assertions on the use count of shared pointers were off by one. At first I thought this was an expected consequence because there would have been fewer references when the underlying object pointer was copied rather than moved. But then I noticed there were also lots of failures where qpdf was raising an exception indicating access to a destroyed QPDFObjectHandle. To me, this was suggesting that some objects were being moved even when the source object was still reachable. Given that this change, which I thought would just give the compiler permission to do certain optimizations, had this effect, I decided it would be best to revert the commit. Honestly I don't think having a move constructor here is important. Move constructors are really better when there's a potentially large amount of memory that's being copied, such as with Buffer or std::string or other collections. QPDFObjectHandle just has a single shared pointer, so the performance impact will not be significant. I wouldn't be comfortable adding this back in without really understanding how it works. I've read the C++ reference on this, so I understand what's supposed to happen, but I'm still a bit murky about it, and I would want to do some experiments. Anyway, given the thrashing around between qpdf, pikepdf, and distributions, I really didn't want to introduce a change with unknown risk that would also require another coordinated change between qpdf and pikepdf.

If you wanted to reproduce this, you could add the change locally and follow the instructions in README-maintainer for how I run the pikepdf test suite against a local clone of qpdf. If you figure it out, I'd be curious to know what you find, but don't spend time on it on my account. Some day I may dig into std::move and friends myself, but I kind of consider qpdf's heavy use of shared pointers to be a reason not to have to think about it that much.

jberkenbilt:
I think we are in 100% agreement on all of these points.

PR #765, opened 2022-09-11
Tidy object resolution following object / value split

m-holger:


jberkenbilt:
I feel like this change adds risk to the code. In the past, I have accidentally added code that caused resolve to be called recursively, which created hard-to-find bugs, so I added these checks as defensive coding. Unless there's a big reason to remove them, I'd rather leave them in place. The code feels more complex and error prone after this change. Can you explain why you think this is a good change? Sometimes I think we need to choose maintainability over performance, unless there's a really big gain.

jberkenbilt:
Also I may have overlooked it but I see you inserting into resolving but not removing from it. My ResolveRecorder is [RAII](https://en.cppreference.com/w/cpp/language/raii) to ensure that the protection against re-entrant resolve is safe across exceptions.

m-holger:
Sorry, I am missing something.

resolve must only be called on an unresolved object. It therefore can be called at most once, after which it is an error to call it again as it would overwrite the cache entry. Removing the `resolving.erase(og)` ensures that most of these cases will be trapped in addition to any recursive calling of resolve.

Without the  `resolving.erase(og)` in the destructor, there does not seem any benefit in having a separate class manage the `resolving` set.

If there is an exception, the unresolved object will be replaced by a null object in the obj cache and therefore resolve will not be  called a second time, even if the user decides to take some action to remediate. 

All that said, I do see a benefit in having a debug_assert for the `isUnresolved()`  precondition.

jberkenbilt:
> Sorry, I am missing something.

I'm probably the one who's missing something. I'll give it a close look before cutting 11.1.

include/qpdf/QPDF.hh
```diff
@@ -399,6 +399,11 @@ class QPDF
     QPDFObjectHandle getObjectByObjGen(QPDFObjGen const&);
     QPDF_DLL
     QPDFObjectHandle getObjectByID(int objid, int generation);
+    // The following overload is for internal use and should not
+    // be required by users of the library. Setting resolve to false will
+    // stop getObject from trying to resolve an unresolved object.
+    QPDF_DLL
+    QPDFObjectHandle getObject(QPDFObjGen const&, bool resolve);
```
jberkenbilt, line 406: If this is really only for `QPDFParser` to call, I'd rather see a private method called something like `getUnresolvedObject` and have it protected by one of those little wrapper classes I like to make that has `QPDFParser` as a friend. You know, something like `QPDF::GetUnresolved::getUnresolved(context, ref_og)`. See summary comments.

libqpdf/QPDF.cc
```diff
@@ -1942,18 +1942,14 @@ QPDF::readObjectAtOffset(
...
-    if (isCached(og) && !isUnresolved(og)) {
-        // We only need to resolve unresolved objects
-        return;
-    }
+    // resolve is (and must only be) called with unresolved objects.
 
-    // Check object cache before checking xref table.  This allows us
-    // to insert things into the object cache that don't actually
-    // exist in the file.
     if (this->m->resolving.count(og)) {
```
jberkenbilt, line 1947: I'd get rid of `m->resolving` entirely and just have throw a logic error if resolve is called on something that's resolved.

jberkenbilt (changes requested):
I see what's going on here. Rather than adding an overloaded `getObject()`, I'd suggest a private `getUnresolvedObject` and a wrapper that grants `QPDFParser` alone access to call it. Then you don't have to change any other existing calls to `getObject()`. The code in `getObjectUnresolved()` can be very short since it doesn't have to handle the direct object case. It can be a completely separate function that doesn't have to call `getObject()` though if you wanted to pull part of it out into a separate function, that would be okay. Let me know if my suggestion is not clear.

Rather than debug_assert for calling a resolved object, I'd rather see a logic error.

jberkenbilt:
I see what you're doing, but unless I'm still missing something, I think the change is more complex than necessary.

jberkenbilt:
I have a better fix, inspired by this one, that is a smaller code change with a better performance gain and also improves error reporting for an existing edge case. I will push it up soon.

jberkenbilt:
I have to step away for a few hours but will push it up when I return. The result is on my work branch but I intend to rework the commits and push up a separate branch with an explanation of what I was trying to get at with my comments here.

m-holger:
I haven't studied your solution in detail, but from what I can tell we came to the same conclusion on:

- it's not worth it making getObject try to resolve objects
- making calling resolve  with a resolved object a logic error would not guard in itself against reentry.
- and therefore , we might as well just return rather than throw a logic error, getting rid of the need to always check twice that the object is unresolved.

From what I can tell at first glance, you removed the check for direct  objects in getObject, creating "direct" unresolved objects.

jberkenbilt:
I plan on closing this in favor of #773 but want to give you @m-holger a chance to take a look.

jberkenbilt:
> From what I can tell at first glance, you removed the check for direct objects in getObject, creating "direct" unresolved objects.

Correct -- it never happens in the regular code path, and there's no reason to really have it as a special case as it actually hides error conditions in files that have `0 0 R` in them. The end result is the same -- 0,0 will never be in the xref table, and so resolve will end up creating a null. The only difference is that it will be cached, but that is harmless. If I wanted to block it, I'd probably block it in resolve itself, just to prevent actually caching it. All the existing test cases around indirect references to object 0 either passed or actually improved somewhat when I made that change.

jberkenbilt:
Your summary of conclusions matches mine. I kind of went around in a circle to get there.

jberkenbilt:
Removal of the explicit check makes this code cause `getAllObjects()` to return a direct null, but I haven't been able to find anything this breaks, and it turns out there has always been a code path through reserved objects to get this same result. I think I don't care. The advantages of removing the check seem to slightly outweigh the disadvantages.

```c++
#define POINTERHOLDER_TRANSITION 4
#include <qpdf/QPDF.hh>
#include <qpdf/QPDFWriter.hh>

int main()
{
    QPDF q;
    q.emptyPDF();
    auto oh = q.getObject(QPDFObjGen(0, 0));
    auto a = "[]"_qpdf;
    a.appendItem(oh);
    q.getTrailer().replaceKey("/Z", a);
    QPDFWriter w(q, "/tmp/a.pdf");
    w.setPreserveUnreferencedObjects(true);
    for (auto& o: q.getAllObjects()) {
        std::cout << o.unparse() << std::endl;
    }
    w.write();
    return 0;
}
```

jberkenbilt:
@m-holger I'd like to release 11.1 today so we can get an official pikepdf out and unblock distributions and so forth. I think we have reached the point of diminishing return on this and wouldn't recommend spending more time on it, though I'm still willing to look at suggested enhancements.

m-holger:
There is no reason to hold up 11.1. From a performance point of view, there is very little to be gained here.

jberkenbilt:
@m-holger Can we close this, or are you still planning to tweak it? I think we've gone far enough here, but I'll defer to your preferences.

m-holger:
Any additional tweaks are to facilitate other changes. Might as well include them with those changes.

PR #768, opened 2022-09-12
Fix CMake Error when cross-compiling

licy183:


jberkenbilt:
I may fix this a different way, but thanks for drawing my attention to it. Can you provide a sample cmake command you use for cross compiling? Maybe I can rewrite `qpdf_check_ll_fmt` to not need to run anything and/or provide a fallback based on toolchain. My old autoconf test for this probably goes back to the beginning (20 years ago) and may not even be needed anymore. I'll have to dig a little.

jberkenbilt:
It looks like inttypes.h actually has standard macros for this and that more recent versions of MSVC may also support lld. It seems that MSVC has supported %lld at least since 2017. I can probably drop this and just hard-coded lld.

jberkenbilt:
Also, this is actually used only by the test suite.

jberkenbilt:
See #769

PR #770, opened 2022-09-12
Missing symbols

jberkenbilt:
@jbarlow83 Can you check to see whether this solves your Windows build issue? I think it should.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@ cmake_minimum_required(VERSION 3.16)
 # the project line. When updating the version, check make_dist for all
 # the places it has to be updated.
 project(qpdf
-  VERSION 11.0.1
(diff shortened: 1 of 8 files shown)


jberkenbilt:
@jbarlow83 Sorry to post in multiple places. Once I get confirmation from you, I can merge this. I'll give it a short time (a few days at most) before re-releasing as 11.1. I want to see what other issues may trickle in.

jbarlow83:
I should be able to get this tested in the next 24 hours.

jberkenbilt:
> I should be able to get this tested in the next 24 hours.

Sounds good. I'll go ahead and merge this change as it is needed whether or not it provides a complete fix. I need an automated way to detect cases of things in the public API that are not exported properly. Maybe.

PR #781, opened 2022-09-21
Replace calls to QUtil::int_to_string with std::to_string

m-holger:
In these cases, this is what int_to_string eventually does anyway.


jberkenbilt:
I was going to say I didn't know about `std::to_string`, but apparently a previous version of me knew about it. Thanks. I guess this will help with performance because it won't pass through the layers of function calls and conditions, and obviously this is called a lot. Besides, no reason to use a qpdf function when a standard one will do. Good change.

m-holger:
On average, a bit over 2.5%. I expected a bit, but not quite that much.

PR #783, opened 2022-09-21
Inline Pipeline methods

m-holger:
For qpdf 12. I see approx 1.5% improvement.


jberkenbilt:
Unfortunately inlining exported methods is ABI-breaking, so this will have to wait.

jberkenbilt:
And you already said `for qpdf 12`, so you clearly know that already.

PR #786, opened 2022-09-23
Refactor Pl_Buffer

m-holger:
Base implementation of the buffer on std::basic_string<unsigned char>.

Performance gain for linearization test is over 2.5%.

### DIFF
--- a/include/qpdf/Pl_Buffer.hh
+++ b/include/qpdf/Pl_Buffer.hh
@@ -35,7 +35,6 @@
 
 #include <qpdf/Buffer.hh>
 #include <qpdf/Pipeline.hh>
-#include <qpdf/PointerHolder.hh>
 
 #include <memory>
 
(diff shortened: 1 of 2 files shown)


include/qpdf/Pl_Buffer.hh
```diff
@@ -35,7 +35,6 @@
 
 #include <qpdf/Buffer.hh>
 #include <qpdf/Pipeline.hh>
-#include <qpdf/PointerHolder.hh>
```
jberkenbilt, line 38: I had left all these `#include <qpdf/PointerHolder.hh>` statements in the public header files so that people who were using `PointerHolder` and getting it indirectly through other things would get the PointerHolder warnings instead of an error about `PointerHolder` not being defined, so in general, I don't want to pull them out until qpdf 12. I should have been explicit about that somewhere. I added a comment to #785 as a reminder.

jberkenbilt (changes requested):
This looks great other than the removal of the include statement that was left behind intentionally.

include/qpdf/Pl_Buffer.hh
jberkenbilt: I'm going to merge as is, then restore the include and add comments to all the includes.

m-holger: Noted for the future.

PR #787, opened 2022-09-23
Tune QPDF_String::useHexString()

m-holger:
Very small efficiency gain, but imo improves  readability.

### DIFF
--- a/libqpdf/QPDF_String.cc
+++ b/libqpdf/QPDF_String.cc
@@ -7,12 +7,6 @@
 // be used.
 #include <string.h>
 
-// See above about ctype.
-static bool
-is_ascii_printable(char ch)
-{
-    return ((ch >= 32) && (ch <= 126));
-}
 static bool
 is_iso_latin1_printable(char ch)
 {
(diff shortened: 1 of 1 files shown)


jberkenbilt:
I'll review this when I get a few minutes. Keep in mind that, on some systems, char is unsigned by default and an others its signed by default. I don't have CI running on any system that has unsigned char by default. I'm not sure whether arm64 is such a system. I am nearly certain that sparc is. I've tried to code things so that they work either way. I would like it if I had a way to have some tests that would fail if char were unsigned. I haven't looked into this. It's possible that there could be a compiler flag with gcc or clang or a way to run CI on arm64 if arm64 has unsigned char by default. Up to now, my approach has been to just always remember that char can be either way. For example, `ch >= 32 && ch <= 126` is the same whether char is signed or unsigned, but each of `ch <= 126` and `ch >= 32` by itself is different.

jberkenbilt:
When I upload the debian package, the test suite gets run on platforms with unsigned char by default, but I don't like to wait until that point to discover those bugs...

jberkenbilt:
gcc has `-fsigned-char` and `-funsigned-char`. Maybe I should have a CI build that builds with `-funsigned-char`. This program prints `-9` on Intel/linux compiled with `gcc a.c` but `247` when compiled with `gcc -funsigned-char a.c`.

```C
#include <stdio.h>
int main()
{
    char ch = '\xf7';
    printf("%d\n", ch);
    return 0;
}
```

jberkenbilt:
I added this in #788. If that's good, we can merge, and you can rebase to make sure this change passes in that setting.

m-holger:
I have to admit I wasn't aware of this and I guess it is another point to add to my list of why I hate C / C++.

However, by lucky happenstance, whether `\xf7` is `-9` or `247`, it will be `< 0` or `> 126`, so the result will be the same.

jberkenbilt:
#788 has gotten far enough in CI for me to know that it is effectively testing with both signed and unsigned characters by default. Once it gets through CI, I'll merge, we can rebase this on top of latest main, and then after CI passes, I'll review. I probably won't do it tonight. I have reached cerebral liquification for the day.

jberkenbilt:
Okay, I merged #788 so I suspect you'll probably be able to rebase this onto of the latest main before I next look at it, whether late tonight (your time) or tomorrow before I start my day. I am logging off for the day and don't intend to be on my computer for the rest of the day.

jberkenbilt (commented):
Looks right, thanks.

PR #788, opened 2022-09-23
Run tests in CI with char as unsigned char

jberkenbilt:
Add CI test with `-funsigned-char`


jberkenbilt:
I may want to add something to ensure that char is really unsigned here and really signed otherwise so I can be certain that the test is doing what it is supposed to do.

PR #793, opened 2022-09-27
Refactor stream creation

m-holger:


jberkenbilt:
Sometimes I look at your changes and feel embarrassed that I didn't do it that way to begin with.

m-holger:
I tend to write my own code the same way. It tends to take me longer refactoring my own code because I think I understand it and therefore switch of my brain. Its only after I have forgotten what I did that I start thinking and see the insanity in my code.

m-holger:
My other problem is that I tend to remember that there was a reason for doing things in an odd way for much longer than I remember what the reason actually was. As a result, odd code survives even when the reason for it has long disappeared.

jberkenbilt:
Yes, I think we have a lot in common in those ways.

I think you mentioned once that (my paraphrase) there wouldn't be much left of `QPDFObjectHandle` after you were done with it. If you intend to provide a replacement for QPDFObjectHandle that is more type-safe, etc., then I think it would be important to remove as much code from QPDFObjectHandle that would be duplicated as possible. Is this a step along that path? I would be in support of making the new object handle the main one and having QPDFObjectHandle exist for backward compatibility. We can discuss offline, but I just wanted to check with you whether this change is part of heading in that direction.

Regarding the timing, I think I'd like to make a quick 11.1.1 release with a small handful of bug fixes (besides, I can't resist having 11.1.1), then after we wait long enough to show that it's not a brown paper bag release (probably just 3 or 4 days), I would bump the C++ minimum version to 17 and move forward toward 11.2. This could be part of 11.2 since it adds new API. That would also enable you to use any C++-17 features you wanted, and it would hopefully give us a little more flexibility to add additional API as needed without having the minor version number climb too fast. Thoughts?

m-holger:
> I think it would be important to remove as much code from QPDFObjectHandle that would be duplicated as possible. Is this a step along that path?

Yes and no. I think QPDFObjectHandle has accumulated too many responsibilities over the years. Moving some of them to mind simplifies the code and makes the learning curve less steep. It will also minimize code duplication in revamp of QPDFObjectHandle.

> Regarding the timing

Works for me. Initial review of the PR before than would still be helpful if you have the time.

m-holger:
Any thoughts on when you will have some time to look at this?

jberkenbilt:
This weekend and next weekend are full, and work is pretty busy. I'm hoping to carve out some time in December to catch up and release 11.2. I really want to start on my page stuff. There's a good chance I will not be online at all tomorrow.

jberkenbilt:
This is good as is. I guess I must have added stream creation to QPDFObjectHandle because all the other objects were created there, but having it be inside QPDF makes more sense than passing a qpdf pointer through a factory class method.

PR #794, opened 2022-09-28
Some code tidying for class QPDF

m-holger:


jberkenbilt:
Generally I'm not a huge fan of lots of overloaded methods with different defaults and so forth, but I think it makes sense in this case. Most of the time I end up with that situation, it's either because of evolving the API and maintaining compatibility or when there are just two overloads: the "normal" way and the way with all the knobs. But I see that this does significantly improve the readability of these damaged PDF messages, and honestly, I have to look each time to remember which argument is which.

m-holger:
@jberkenbilt

I share your view. This was the least bad option that I could think of. I'll revisit when C++35 supports named parameters.

jberkenbilt:
> I share your view. This was the least bad option that I could think of. I'll revisit when C++35 supports named parameters.

I almost mentioned named parameters in my response. I'm not planning to require C++35 until qpdf version 43, but it's likely that the world will end on January 19, 2038 anyway, and I doubt all the compilers will have good support for C++35 yet.

PR #795, opened 2022-09-29
Add methods QPDFTokenizer::Token::isInteger and isWord

m-holger:


include/qpdf/QPDFTokenizer.hh
```diff
@@ -107,6 +107,27 @@ class QPDFTokenizer
...
+        {
+            return this->type == tt_word;
+        }
+        bool
+        isWord(std::string const& value) const
+        {
+            return this->type == tt_word && this->value == value;
+        }
+        bool
+        isWord(std::string const& value1, std::string const& value2) const
```
jberkenbilt, line 126: I like the change except not this overload of isWord. You can just say `isWord("A") || isWord("B")` and not worry about the slight amount of extra work.

libqpdf/QPDF_linearization.cc
```diff
@@ -121,13 +121,9 @@ QPDF::isLinearized()
         }
 
         QPDFTokenizer::Token t1 = readToken(this->m->file);
-        QPDFTokenizer::Token t2 = readToken(this->m->file);
```
jberkenbilt, line 124: This actually changes the meaning -- the old code unconditionally reads 4 tokens, while the new code may or may not read tokens depending on whether earlier tokens pass their checks. That means that this change changes where the file points to after the function is called. Have you determined whether this is safe?

jberkenbilt: There is a least one other example of that in this PR. Probably they are fine, but I haven't studied it.

jberkenbilt (changes requested):

libqpdf/QPDF_linearization.cc
m-holger: If the tokens don't pass the test either the loop is continued in which case another seek is called to explicitly set the file position before the next token is read, or the method returns false, in which case none of the callling code makes assumptions about the file position.

include/qpdf/QPDFTokenizer.hh
m-holger: I'll take this out

jberkenbilt (approved):
Looks good. Thanks again for the contribution.

PR #796, opened 2022-09-29
Overload QPDFTokenizer::readToken to take an InputSource&

m-holger:
shared_ptrs definitely do not come for free.

### DIFF
--- a/include/qpdf/QPDFTokenizer.hh
+++ b/include/qpdf/QPDFTokenizer.hh
@@ -177,7 +177,12 @@ class QPDFTokenizer
         std::string const& context,
         bool allow_bad = false,
         size_t max_len = 0);
-
+    QPDF_DLL
+    Token readToken(
+        InputSource& input,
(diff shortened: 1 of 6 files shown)


jberkenbilt:
I'll catch up on your pull requests soon. This would be a function that most people don't use though it is of course used heavily internally. If the performance gains are significant, it might be worth deprecating the old one, though I'd save adding the deprecation tag for 11.2 (rather than 11.1.1), so I'd add it later in a separate change. I plan on reviewing all your existing PRs before releasing 11.1.1 but if I have a chance, I will do 11.1.1 no later than Monday and maybe as early as tomorrow. Other than deciding which of your changes are going in for 11.1.1 and which for later, I don't have anything else. Let me know if you are working on anything that is low risk/high impact that should go in for 11.1.1.

jberkenbilt:
Actually, this adds to the public binary interface, so this will have to wait for 11.2.

PR #802, opened 2022-10-05
Refactor scanning and parsing

m-holger:


m-holger:
This is still work in progress.

As it stands at the moment, this is ABI breaking and therefore would be for qpdf 12.

jberkenbilt:
I added the `abi` label as you see. You can always remove it if it makes sense to do so in the future. I haven't scrutinized this yet. BTW, I think if you rebase, your change to enable C++17 will disappear since I pushed that up separately. It may be a couple of weeks before I have significant time for qpdf. This weekend and next weekend are quite full. If I have qpdf time, it will be just to take care of small issues. Thanks as always for your efforts.

m-holger:
I am not sure whether it is clear from the commits where I am going. If not, let me know and I'll summarize. In any case, this will need more documentation comments before it gets finalized.

I can see the possibility that there will be a case for API breaking changes, in which case this could be modified to go into 11.2.

m-holger:
Btw, is there a way to turn of sanitizer overflow checking for specific sections of code?

jberkenbilt:
> Btw, is there a way to turn off sanitizer overflow checking for specific sections of code?

I don't think so, but why would you want to?

m-holger:
Because whether or not the variable overflows makes no difference in this case. I (have to)  detect the risk of an overflow and handle it. Actually preventing the overflow adds overhead with no benefit

jberkenbilt:
The sanitizers only complain if there is an actual overflow, and if they complain, then there is an actual overflow. I'm not sure I understand exactly what situation there could be where it makes no difference if the variable overflows. Anyway, even if it really did make no difference, I would start getting bug reports all over the place and rejections from places. For example, both Red Hat and Google wanted at() to be used instead of operator[] for vector because of static analyzers. People are willing to pay a performance price for overflow checks.

PR #814, opened 2022-11-12
Warn if catalog type entry is invalid (fixes #810)

### LINKED ISSUE #810: "qpdf --check" should detect incorrect /Type
With qpdf 11.1.1, on a PDF file that has `/Type /Calalog` instead of `/Type /Catalog`, `qpdf --check` doesn't detect any error, while the `/Type` is incorrect: https://bugs.ghostscript.com/show_bug.cgi?id=706038

`qpdf --check` should detect such an error.

m-holger:


jberkenbilt:
Thanks for taking it on. These changes seem like a good start to something that's been often discussed, which is enabling a higher level of warnings that would be of concern to developers or people validating PDF files but not necessarily of concern to most users.

I agree with the general direction of the solution, which adds only the trivial cost of checking a boolean before issuing warnings. Before we commit, there a few things I'd like us to decide:
* Is `setCheckMode` the right name for this? We are enabling it now for check mode, but I think there should be a general command-line flag that enables additional warnings for people trying to troubleshoot PDF files, so this would include warnings for things like this that qpdf can check but doesn't care about or other cases where qpdf just deals with something.
* It would be nice if this could be implemented in a way that made other classes able to take advantage of this functionality but without adding runtime cost of checking for things needlessly. One idea that comes to mind is to build something into the logger where we could have flags that could be checked efficiently. Probably something that would be internally implemented with a method that resolves strings into bit values stored in a bit field internally. I feel like if we have more than a handful of these, we're way overdoing it, so a 64-bit unsigned integer type should be adequate.

I don't want to overdesign this, but before we add a method to the public API, I want to do something that we're not going to want to turn around and deprecate. Comments welcome.

m-holger:
setCheckMode works for me, as it enables warnings appropriate for checking pdf files.

Personally I can only see two other modes that make sense to me - UserMode and DeveloperMode, where DeveloperMode is a superset of UserMode and CheckMode a superset of DeveloperMode. On that assumption, boolean flags feel more appropriate than bit flags to me.

I agree that there should be a general flag to control the level of message, perhaps --verbosity, with possible values 0 (suppress all warnings) to 3 (enable CheckMode).

jberkenbilt:
I'll give it some thought. I don't want the perfect to be the enemy of the good, and this is definitely good. I'll address this before 11.2 since it has new ABI. I should really go ahead and get 11.2 out since it has the major performance fixes to openssl which affects my default Windows binaries.

m-holger:
I have removed the method to set check mode from the public API.

libqpdf/QPDF.cc
```diff
@@ -2461,6 +2461,10 @@ QPDF::getRoot()
     QPDFObjectHandle root = this->m->trailer.getKey("/Root");
     if (!root.isDictionary()) {
         throw damagedPDF("", 0, "unable to find /Root dictionary");
+    } else if (
+        m->check_mode && !root.getKey("/Type").isNameAndEquals("/Catalog")) {
```
jberkenbilt, line 2465: Can you add a comment here explaining that check_mode is an interim step toward a more comprehensive concept of warning levels? Or something like that? I just don't want to forget about it if we ever do go through the exercise of looking at all warnings. Feel free to reference this PR in the comment. If we were going to touch all warnings, there are several things people have asked for such as a table somewhere where someone could look up a warning and get a more detailed explanation. Whenever I think about this, I'm also painfully aware of the fact that the qpdf source code is not structured for internationalization either....but that's another whole issue and out of scope for this.

jberkenbilt (changes requested):
This is good to go as far as I'm concerned except for the addition of a comment as requested in a PR comment.

libqpdf/QPDF.cc
m-holger: Done.

jberkenbilt (approved):

PR #815, opened 2022-11-14
Fix stray formatting error

m-holger:


jberkenbilt (approved):
Yeah, I see this too if I upgrade clang-format. Looks like they fixed a bug with operator"". Now if emacs will fix it too....

PR #816, opened 2022-11-15
Refactor object copying

m-holger:


m-holger:
This was purely meant to be a code tidying exercise, but I'm not complaining.

jberkenbilt:
I don't think I agree with this change.

You'll need to explain the motivation of this change and what it achieves. It would take a fair amount of effort for me to review all these changes and convince myself that the functionality is unchanged and that the new implementation is better than the old one. Can you give me some hints? In particular, it's hard for me to see how this preserves the essence of what unsafeShallowCopy does, which is explicitly to copy only top-level keys. Also, I prefer shallowCopy() to copy(true). With copy(true), you have to look at the documentation to know what it does.

m-holger:
My initial motivation was twofold - (1) simplify the code and make it easier (for me) to understand, and (2) move functionality out of QPDFObjectHandle. For both of these purposes I moved the copy functionality to the QPDFValue subclasses, which know how to copy themselves.

I am not hung-up on the names. The existing QPDFValue::shallowCopy methods are true shallow copy methods (likeQPDFObjectHandle::unsafeShallowCopy). The new functionality added is to allow the whole object to be copied (like QPDFObjectHandle::shallowCopy).

I went for the option of having a parameter to control which of these behaviors is required because for all QPDFValue  types except arrays and dictionaries the code for a "unsafeShallowCopy" and a "shallowCopy" are identical. 

While not necessarily anticipated, I am seeing a performance improvement of around 4%.

jberkenbilt:
I had not run a performance test on this, but now I understand why you said "I'm not complaining."

I wonder whether we can better isolate the part that gives the performance gain. Can we also guarantee that we didn't lose the distinction between unsafe and regular shallowCopy?

I'll take another pass through and make specific suggestions. Thanks for the clarification. It would be helpful when you present a pull request with a bunch of changes and refactoring to provide an explanation so I have some context. Otherwise, I'm just looking at a bunch of changes out of context and having to go through the exercise of figuring out whether the changes are safe (which they usually are but I like to verify), whether I think they improve the code (which I usually think they do but not always), and what the benefit is.

m-holger:
> Can we also guarantee that we didn't lose the distinction between unsafe and regular shallowCopy?

I have pushed up a couple of intermediate commits that I hope demonstrate that the distinction has been maintained.

jberkenbilt (approved):
I think the key insight, which is actually explained in an existing comment in QPDFObjectHandle.hh, is that a "shallow copy" of a QPDFObjectHandle is really a deep copy that stops at indirect object boundaries, hence shallowCopy() does a full copy of the object. unsafeShallowCopy does a shallow copy of the underlying object.

Your intermediate commits are great. Thanks for including those. They definitely make it easier to validate the thought process.

I think this is ready to go now. Thanks again for helping me see what you were doing.

PR #817, opened 2022-11-19
Refactor QPDFObjectHandle::asArray, isArray and assertArray etc

m-holger:


m-holger:
CI produces the error messages below. This looks like a problem with the github action rather then the PR to me.

+ grep -E '^qpdf version 11.2.0'
dlopen(): error loading libfuse.so.2

AppImages require FUSE to run. 
You might still be able to extract the contents of this AppImage 
if you run it with the --appimage-extract option. 
See https://github.com/AppImage/AppImageKit/wiki/FUSE 
for more information
Error: Process completed with exit code 1.

jberkenbilt:
I'm aware of the appimage failure. I'm going to address it today. If the bug isn't fixed upstream, I'll work around it by installing the fuse library. I'm hoping to have some qpdf time this weekend.

jberkenbilt:
I don't agree with this change. I like the explicit isX and asX better than templates. This doesn't feel like a good use of templates since there is actually a very specific, fixed list of things these can be, and I don't think this improves performance or readability of the code.

As a rule, I don't care for this type of refactoring when not in explicit service of a larger goal, such as significant performance improvement or preparing the code for some major feature or rework that is made easier or safer by the refactoring. Otherwise, this type of change is disorienting to me as I go back into old code since it doesn't look the way I am used to having it look. Also, reviewing these types of changes requires time and effort that I could better spend doing other things with the code.

Of course, you are free to make these suggestions, and sometimes I may accept them, but it's helpful if they are presented in some context.

PR #822, opened 2022-11-25
Refactor QPDF::fixDanglingReferences

m-holger:
This feels too simple, but unless I am missing something, the rest has been taken care of by #726.

I see approx 10% performance improvement.

### DIFF
--- a/libqpdf/QPDF.cc
+++ b/libqpdf/QPDF.cc
@@ -577,6 +577,8 @@ QPDF::reconstruct_xref(QPDFExc& e)
     }
 
     this->m->reconstructed_xref = true;
+    // We may find more objects, which may contain dangling references.
+    this->m->fixed_dangling_refs = false;
 
     warn(damagedPDF("", 0, "file is damaged"));
     warn(e);
@@ -1290,65 +1292,48 @@ QPDF::showXRefTable()
     }
 }
 
+// Ensure all objects in the pdf file, including those in indirect references,
+// appear in the object cache.
 void
 QPDF::fixDanglingReferences(bool force)
 {
-    if (this->m->fixed_dangling_refs && (!force)) {
+    if (this->m->fixed_dangling_refs && !force) {
         return;
     }
-    this->m->fixed_dangling_refs = true;
 
-    // Create a set of all known indirect objects including those
-    // we've previously resolved and those that we have created.
-    std::set<QPDFObjGen> to_process;
-    for (auto const& iter: this->m->obj_cache) {
-        to_process.insert(iter.first);
-    }
-    for (auto const& iter: this->m->xref_table) {
-        to_process.insert(iter.first);
-    }
-
-    // For each non-scalar item to process, put it in the queue.
-    std::list<QPDFObjectHandle> queue;
-    queue.push_back(this->m->trailer);
-    for (auto const& og: to_process) {
-        auto obj = getObject(og);
-        if (obj.isDictionary() || obj.isArray()) {
-            queue.push_back(obj);
-        } else if (obj.isStream()) {
-            queue.push_back(obj.getDict());
-        }
-    }
-
-    // Process the queue by recursively resolving all object
-    // references. We don't need to do loop detection because we don't
-    // traverse known indirect objects when processing the queue.
-    while (!queue.empty()) {
-        QPDFObjectHandle obj = queue.front();
-        queue.pop_front();
-        std::list<QPDFObjectHandle> to_check;
-        if (obj.isDictionary()) {
-            std::map<std::string, QPDFObjectHandle> members =
-                obj.getDictAsMap();
-            for (auto const& iter: members) {
-                to_check.push_back(iter.second);
-            }
-        } else if (obj.isArray()) {
-            auto arr = QPDFObjectHandle::ObjAccessor::asArray(obj);
-            arr->addExplicitElementsToList(to_check);
-        }
-        for (auto sub: to_check) {
-            if (sub.isIndirect()) {
-                if ((sub.getOwningQPDF() == this) &&
-                    isUnresolved(sub.getObjGen())) {
-                    QTC::TC("qpdf", "QPDF detected dangling ref");
-                    queue.push_back(sub);
+    if (!this->m->fixed_dangling_refs) {
+        // First pass is only run if the the xref table has not been
+        // reconstructed. It will be terminated as soon as reconstruction is
+        // triggered.
+        if (!this->m->reconstructed_xref) {
+            for (auto const& iter: this->m->xref_table) {
+                auto og = iter.first;
+                if (!isCached(og)) {
+                    m->obj_cache[og] =
+                        ObjCache(QPDF_Unresolved::create(this, og), -1, -1);
+                    if (this->m->reconstructed_xref) {
+                        break;
+                    }
                 }
-            } else {
-                queue.push_back(sub);
+            }
+        }
+        // Second pass is skipped if the first pass did not trigger
+        // reconstruction of the xref table.
+        if (this->m->reconstructed_xref) {
+            for (auto const& iter: this->m->xref_table) {
+                auto og = iter.first;
+                if (!isCached(og)) {
+                    m->obj_cache[og] =
+                        ObjCache(QPDF_Unresolved::create(this, og), -1, -1);
+                }
             }
         }
     }
+    // Final pass adds all indirect references to the object cache.
+    for (auto const& iter: this->m->obj_cache) {
+        resolve(iter.first);
+    }
+    this->m->fixed_dangling_refs = true;
 }
 
 size_t
@@ -2082,6 +2067,8 @@ QPDF::reserveStream(QPDFObjGen const& og)
 QPDFObjectHandle
 QPDF::getObject(QPDFObjGen const& og)
 {
+    // This method is called by the parser and therefore must not
+    // resolve any objects.
     if (!isCached(og)) {
         m->obj_cache[og] = ObjCache(QPDF_Unresolved::create(this, og), -1, -1);
     }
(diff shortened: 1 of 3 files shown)


jberkenbilt (changes requested):
There needs to be a comment explaining _how_ the second pass fixes dangling references, and either there needs to be a coverage case or the comment has to make it obvious enough that we don't need a coverage case. For this to be right, something would have to cause resolving an object that contains a dangling ref to cause the dangling ref to appear in the object cache. It's not obvious to me what mechanism this is.

libqpdf/QPDF.cc
```diff
@@ -1293,62 +1293,28 @@ QPDF::showXRefTable()
...
+            if (!isCached(og)) {
+                m->obj_cache[og] =
+                    ObjCache(QPDF_Unresolved::create(this, og), -1, -1);
             }
         }
+        for (auto const& iter: this->m->obj_cache) {
+            resolve(iter.first);
+        }
+    }
+    // Final pass fixes dangling references reference.
```
jberkenbilt, line 1312: Here's where to put the explanation. Perhaps a comment in whatever part of the code is causing the dangling reference to appear in the object cache might be in order as well.

jberkenbilt:
I'm not saying I think the code is wrong -- the code is probably right, and there is an existing test. It's just no longer obvious to me reading the code that it is right. A few months out from #726, it's already hard for me to feel fully confident about it, so I'm sure that, years from now, it won't be obvious at all. The old code is obvious to me. I also get a little nervous about the loss of the coverage case for a tricky edge case, even though I know that we have a test for it. The lack of the coverage case also makes it harder to locate the test if we should ever need to do it. qtest creates an output file mapping each test to all the coverage cases it calls. I rely on this to find where in the test suite a condition may be exercised. It's particularly valuable for coverage cases that only appear once.

m-holger:
If this implementation is correct, there are a couple of conclusions.

It seems that fixDanglingReferences implements two tasks that now can be cleanly separated:

- populate the object cache with the (unresolved) objects from the xref table, and

- detect dangling references and replace them with null objects.

The first part could probably be done on the fly while constructing the xref table. Unless there is a case for parsing objects before the xref table has been read, the second part could probably done during parsing.

m-holger:
> I'm not saying I think the code is wrong -- the code is probably right,

I'm not saying the code is right, only that it passes the test suite.

What I am not sure about is what exactly a repair of the xref table does. Crucially,

- does it find objects that are in he file but are not in the xref table?, and
- does a single repair find all of them?  

I was rather hoping you know the answer to that as I find the relevant code heavy to read.

Assuming that a repair of the xref table does not add any new objects to the table, the new code works because during the first pass all existing objects are parsed, and as a result all indirect object references are added to the object cache as unresolved objects (unless they were already added earlier). During the second pass they will be replaced by nulls.

jberkenbilt:
> * populate the object cache with the (unresolved) objects from the xref table, and

I don't think we need to do this unless doing so helps solve some other problem. This can be done lazily as it is now. It would be needed for this, but you're already doing it here, which feels right. I think creating unresolved objects in the object cache while reading the xref table would have a negative performance and memory impact for cases where people are just opening a PDF file and accessing a few objects in it. This usage pattern is common but not covered in the performance testing benchmarks. I might want to think about adding some performance tests. Like if we had a performance case for just extracting the outlines tree or page labels or maybe even the first page...

> * detect dangling references and replace them with null objects.

Yeah, I'm trying to figure out whether that's actually happening. What I was trying to find in the code (I only spent a few minutes) was whether, whenever we encounter an indirect object reference during parsing, we add an unresolved object to the object cache if there isn't already something there. I feel like if we did this, then all that would be required to fix dangling references would be to resolve all objects in the xref table and all objects in the cache. This is exactly what your implementation does, so I think your implementation is correct as long as encountering an indirect object reference during parsing inserts an unresolved object into the object cache if something isn't already there. Is the code already doing that? It wasn't obvious to me that it was.

m-holger:
> What I was trying to find in the code (I only spent a few minutes) was whether, whenever we encounter an indirect object reference during parsing, we add an unresolved object to the object cache if there isn't already something there.

We do.

jberkenbilt:
Why is this related to xref table repair?

jberkenbilt:
> > What I was trying to find in the code (I only spent a few minutes) was whether, whenever we encounter an indirect object reference during parsing, we add an unresolved object to the object cache if there isn't already something there.
> 
> We do.

Can you point me to where in the code this happens? I'm sure I can find it, but apparently you already know where it is.

m-holger:
> Why is this related to xref table repair?

Because an underlying assumption in my solution is that all objects in the pdf file are listed in the xref table. If a xref repair can potentially find additional objects that would need to be handled.

jberkenbilt:
I don't even know if the second pass is needed. I think just resolving all objects in the xref table is sufficient. Actually converting dangling unresolved objects to nulls doesn't buy us anything at this point. The goal here is to make sure that code that determines what the next available object ID doesn't pick a number that appears in a dangling ref. Having the unresolved object in the cache should be sufficient for this.

I guess you're concerned about this action triggering xref repair. xref repair can definitely add new items to the xref table. It only ever happens one time. I think the existing code probably doesn't handle this correctly and may even crash if fixDanglingReferences triggers xref repair. That would be a good case to exercise.

jberkenbilt:
I posted my comment "I don't even know if the second pass is needed..." before I read your comment "Because an underlying assumption..." and my comment "I don't think we need to do this unless..." before I read "I'm not saying the code is right...". Clearly we're thinking in the same direction and I'm catching up with your thoughts.

m-holger:
> > > What I was trying to find in the code (I only spent a few minutes) was whether, whenever we encounter an indirect object reference during parsing, we add an unresolved object to the object cache if there isn't already something there.
> > 
> > 
> > We do.
> 
> Can you point me to where in the code this happens? I'm sure I can find it, but apparently you already know where it is.

QPDFParser line 193 calls QPDF::getObject, which adds an unresolved object to the cache if there is no existing entry.

m-holger:
> I don't even know if the second pass is needed. 

In some places it definitely is not needed. For e.g. getAllObjects its questionable. My feeling would be that the client code would resolve any unresolved objects anyway, thus replacing dangling references with nulls.

jberkenbilt:
Would you be willing to cherry-pick 5a9d6de4af88af165bd15e26e8f9d84dd8df1b56 (from my fork) into this PR?

jberkenbilt:
> In some places it definitely is not needed. For e.g. getAllObjects its questionable. My feeling would be that the client code would resolve any unresolved objects anyway, thus replacing dangling references with nulls.

I don't think the second pass is ever needed.

jberkenbilt:
This is a good find, by the way. I overlooked that fixDanglingRefences could be so simplified. I think we were focused on no longer needing to detect that an object had changed.

libqpdf/QPDF.cc
jberkenbilt: except now I think this pass is not needed at all.

m-holger:
I'll have a second go at this. Do you mind if I just steal your comments rather than adding them as a second commit?

jberkenbilt:
> I'll have a second go at this. Do you mind if I just steal your comments rather than adding them as a second commit?

Yes, that's fine. Thanks.

jberkenbilt:
Sometime in the next hour or two, I will attempt to create a test case that depends upon correct handling of reconstruction of the xref table for correct detection of dangling IDs. If the test fails, which I suspect it will, I will just mark it as an expected failure. You can see if you can make it pass with your fix. Let me know if you already have such a test file. I don't think it will take me more than a few minutes to create one.

m-holger:
I don't, and I think it would take me rather longer.

jberkenbilt:
I'm going to work on two quick changes, which I will push to main, then you can rebase on that:
1. Add something to the performance test benchmark that exercises just accessing a few objects
2. A file that only works right if xref recontruction is taken into consideration when fixing dangling references -- most likely this will be an XFAIL test case.

jberkenbilt:
You will need to update performance-test-files. There was a mistake in many-nulls.pdf.

jberkenbilt:
I merged #824 so you should be good to rebase off of that.

libqpdf/QPDF.cc
```diff
@@ -1290,65 +1292,48 @@ QPDF::showXRefTable()
     }
 }
 
+// Ensure all objects in the pdf file, including those in indirect references,
+// appear in the object cache.
 void
 QPDF::fixDanglingReferences(bool force)
 {
-    if (this->m->fixed_dangling_refs && (!force)) {
```
jberkenbilt, line 1296: Why did you remove `!force`? Even though you clear `fixed_dangling_refs`, there may be a reason to call this with `force`. For example, new ones may have come across when copying foreign objects or with other manual manipulation. `QPDFWriter` calls this with force, so we should still respect `force`.

jberkenbilt (changes requested):
Other than `force`, this looks right. I'm surprised that it didn't fix the XFAIL though. I'll have to look into that.

libqpdf/QPDF.cc
m-holger: My mistake. I took it out for testing and forgot to put it back in.

That said, there probably should be a test case that relies on force == true. Not sure on how to construct it

m-holger: Now fixed.

jberkenbilt: > That said, there probably should be a test case that relies on force == true. Not sure on how to construct it

There probably should be. I'm not sure it's actually possible to create such a case anymore. I can think of ways that would have worked to reproduce it before the object/value split such as inserting a new object with the version of parse that takes a QPDF context and including a dangling reference there, but that would no longer cause a dangling reference because the dangling reference would get inserted into the object cache.

This may just be defensive coding now.

jberkenbilt (approved):
I'm happy with this change, though I'm still a bit confused as to why it doesn't cause the xfail case to turn into a pass. Maybe there's something wrong with my test case. I may look into it further.

jberkenbilt:
I think I found the problem.

PR #824, opened 2022-11-25
Dangling xref prep

jberkenbilt:
@m-holger FYI. I don't usually do pure "test-driven development", but I believe the XFAIL test case I added will pass once the fix is made. I can verify that it passes if I insert an extra call to `fixDanglingReferences(true)` in test_driver.cc. Anyway, this confirms my suspicion that the existing code is wrong with respect to xref reconstruction. It was harder than I thought  to reproduce this. I had to create the dangling reference inside of an unreferenced object. Finding a file like this in the wild that would actually trigger this bug seems unlikely, but it's not impossible. In the current code, adding a new object actually does collide with an existing object.


jberkenbilt:
If you can get the test case to go to unexpected pass, you can do the inverse of 32251497c1c7185380dd7f7836613bfe024285b8 ("Temporary (revert after fix)...") not as a separate commit but inlined with the commit that fixes the bug.

jberkenbilt:
Ultimately all I did to create the test file was add a dangling object with an unreferenced object in it to the file, then break the xref table for an existing object.

m-holger:
I don't think #822 will address the dangling-bad-xref test. The problem is that the new object is inserted before reconstruct_xref is called.

I can achieve a pass if I call reconstruct_xref at the beginning of fixDanglingReferences but I don't think this is desirable. To me it would make more sense to either not preserve clashing dangling references or renumber them. Either case would be out of scope of #822

m-holger:
Another possible solution would be to if necessary renumber inserted objects as part of reconstruct_xref. I'll look into that possibility.

jberkenbilt:
> Another possible solution would be to if necessary renumber inserted objects as part of reconstruct_xref. I'll look into that possibility.

I don't think that would make sense. reconstruct_xref is basically right from an earlier version of the PDF spec, though I am not able to put my hands on the section. I don't think we should mess with the logic of reconstruct_xref.

m-holger:
Not sure we are talking about the same thing. What I am contemplating is:

damaged xref table shows first free object (10, 0)
we insert new objects (10 0) and (11 0).
reconstruct_xref finds object (10 0) 
I am suggesting renumbering the new objects from (10 0) and (11 0) to (11 0) and (12 0) in response to the newly discovered object (10 0).

jberkenbilt:
When you insert a new object, findDanglingRefs gets called first exactly to avoid this problem. Unless I'm misunderstanding. In other words, the fact that reconstruct_xref finds 10 means that when we insert a new object, it will be 11, not 10.

m-holger:
Except, as I understand it, reconstruct_xref gets only called when we try to load a damaged object. So, if we insert the new object before trying to resolve either (10 0) or (11 0), the new object will mask (10 0).

I think if we insert two new objects first, qpdf  will not call reconstruct_xref at all.

jberkenbilt:
nextObjGen calls fixDanglingReferences. The fixDangingReferences function was added specifically to fix the bug you are describing. There is no normal path to insert a new object that does not call fixDanglingReferences, so the situation you are describing can't happen. You can't insert a new object before trying to resolve either 10 or 11. The act of inserting a new object calls fixDanglingReferences before the object ID is assigned.

Do you see a code path for adding a new object that does not call fixDanglingReferences?

PR #825, opened 2022-11-26
Fix fix dangling

jberkenbilt:
@m-holger It is the calls to resolve that may trigger xref reconstruction.

### DIFF
--- a/libqpdf/QPDF.cc
+++ b/libqpdf/QPDF.cc
@@ -1292,48 +1292,48 @@ QPDF::showXRefTable()
     }
 }
 
-// Ensure all objects in the pdf file, including those in indirect references,
-// appear in the object cache.
 void
 QPDF::fixDanglingReferences(bool force)
 {
+    // Ensure all objects in the pdf file, including those in indirect
+    // references, appear in the object cache.
     if (this->m->fixed_dangling_refs && !force) {
         return;
     }
 
-    if (!this->m->fixed_dangling_refs) {
-        // First pass is only run if the the xref table has not been
-        // reconstructed. It will be terminated as soon as reconstruction is
-        // triggered.
-        if (!this->m->reconstructed_xref) {
-            for (auto const& iter: this->m->xref_table) {
-                auto og = iter.first;
-                if (!isCached(og)) {
-                    m->obj_cache[og] =
-                        ObjCache(QPDF_Unresolved::create(this, og), -1, -1);
-                    if (this->m->reconstructed_xref) {
-                        break;
-                    }
-                }
-            }
-        }
-        // Second pass is skipped if the first pass did not trigger
-        // reconstruction of the xref table.
-        if (this->m->reconstructed_xref) {
-            for (auto const& iter: this->m->xref_table) {
-                auto og = iter.first;
-                if (!isCached(og)) {
-                    m->obj_cache[og] =
-                        ObjCache(QPDF_Unresolved::create(this, og), -1, -1);
-                }
-            }
+    // Make sure everything in the xref table appears in the object
+    // cache.
+    for (auto const& iter: this->m->xref_table) {
+        auto og = iter.first;
+        if (!isCached(og)) {
+            m->obj_cache[og] =
+                ObjCache(QPDF_Unresolved::create(this, og), -1, -1);
         }
     }
-    // Final pass adds all indirect references to the object cache.
+
+    // Resolve all known objects. The parser inserts any indirect
+    // reference into the object cache, including dangling references.
+    bool orig_reconstructed_xref = this->m->reconstructed_xref;
+    bool triggered_xref_reconstruction = false;
     for (auto const& iter: this->m->obj_cache) {
         resolve(iter.first);
+        if (!orig_reconstructed_xref && this->m->reconstructed_xref) {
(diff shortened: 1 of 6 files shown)


jberkenbilt:
I'm done tweaking.

libqpdf/QPDF.cc
```diff
@@ -1292,48 +1292,48 @@ QPDF::showXRefTable()
     }
 }
 
-// Ensure all objects in the pdf file, including those in indirect references,
-// appear in the object cache.
 void
 QPDF::fixDanglingReferences(bool force)
 {
+    // Ensure all objects in the pdf file, including those in indirect
+    // references, appear in the object cache.
```
m-holger, line 1299: With the comment above the void, visual editors will display the comment (together with the method signature) as a tool tip wherever the method is called.

With the comment inside the method body, the comment will not be displayed.

libqpdf/QPDF.cc
```diff
@@ -1292,48 +1292,48 @@ QPDF::showXRefTable()
...
-            for (auto const& iter: this->m->xref_table) {
-                auto og = iter.first;
-                if (!isCached(og)) {
-                    m->obj_cache[og] =
-                        ObjCache(QPDF_Unresolved::create(this, og), -1, -1);
-                    if (this->m->reconstructed_xref) {
-                        break;
-                    }
-                }
-            }
```
m-holger, line 1318: How about having a resolve loop after the for loop if reconstructed_xref == false.

It after the the resolve loop reconstructed_xref is still false we can return, otherwise we continue with the second pass.

libqpdf/QPDF.cc
```diff
@@ -1292,48 +1292,48 @@ QPDF::showXRefTable()
...
-        if (!this->m->reconstructed_xref) {
-            for (auto const& iter: this->m->xref_table) {
-                auto og = iter.first;
-                if (!isCached(og)) {
-                    m->obj_cache[og] =
-                        ObjCache(QPDF_Unresolved::create(this, og), -1, -1);
-                    if (this->m->reconstructed_xref) {
-                        break;
-                    }
-                }
```
m-holger, line 1317: see next comment

m-holger:
This fixes the problem of an referenced object that is not in xref table being masked by a newly created object.

There is another (non?)-problem of a unreferenced object that is not in the xref table being omitted when the setPreserveUnreferencedObjects option is selected.

jberkenbilt:
> There is another (non?)-problem of a unreferenced object that is not in the xref table being omitted when the setPreserveUnreferencedObjects option is selected.

I don't care about that.

jberkenbilt:
> This fixes the problem of an referenced object that is not in xref table being masked by a newly created object.

That's the only way I could think of reproducing the problem, but I'm not completely convinced that it couldn't happen another way. Anyway, the previous logic was incorrect since just adding unresolved objects to the object cache will never trigger xref reconstruction. Only resolve can do that.

libqpdf/QPDF.cc
jberkenbilt: Okay, I didn't know why you had deviated from my convention of putting the comment inside the function. I keep the comment inside the function because emacs handles that case better.

libqpdf/QPDF.cc
jberkenbilt: To me, it's just an implementation detail. My preferred implementation is to just call it again instead of having the code replicated. Anyway, this path can happen at most once for any QPDF object, and it's a very rare case.

jberkenbilt:
I'm going to go ahead and merge this. If you feel strongly about a different implementation, feel free to do another PR, but at this point, the major improvement of making this much more efficient by not requiring a full traversal is in, and the long-standing (but probably unimportant) issue of this not playing properly with xref reconstruction is also fixed.

Unreferenced objects that are not in the xref table are normal and can happen when objects are reused. Trying to detect them for a non-broken file would be incorrect. The fact that they show up with a broken file is probably undesirable, except when a file is broken, you really never know why. By far the biggest reason I've seen is that people modify the contents of an object without updating the xref table.

jberkenbilt:
Thanks for finding this issue and drawing my attention to it and for providing the essential logic of the fix. I consider this to be primarily your contribution. I spent under 10 minutes tweaking to move the retry logic to the resolve call.

libqpdf/QPDF.cc
jberkenbilt: I'm going to explore lsp-mode in emacs and see if I can get some more modern IDE features. If there are certain conventions that we can follow uniformly that help out with various IDEs or visual editors, I'm open to it. This may be another one of those cases where we do a bulk change, but we should agree on it first. I want to make sure I don't break my emacs setup though or that I can rework it so that it handles changes properly. That said, I have some tolerance for issues here. I've gotten my emacs configuration for automatic indentation to work _almost_ like clang-format so that, most of the time, my code matches what the ultimate format will be. Then I just do a clang-format-buffer to fix it and don't mess around with it from inside emacs. The most destructive emacs operation is to indent a block, but I never do that anymore with clang-format-buffer. Anyway, you've mentioned the comment outside the function here, and I know you've mentioned other things other places but I don't recall the details. I'll open a discussion ticket.

PR #832, opened 2022-12-04
Remove redundant friend class statements

m-holger:


jberkenbilt:
Okay, obviously these are redundant because CI passed. Hmm. It seems that I have had a bug in my brain about this. It seems that a nested class has access to its parents' private interface. Somehow I did not think that was the case. I'm going to mark "dirty" all my brain storage about outer and inner classes' access to each other's private stuff across different languages.

PR #834, opened 2022-12-05
Pass exp_og by value in QPDF::readObjectAtOffset

m-holger:
Prevent calls to reconstruct_xref from accidentally changing an exp_og passed from the xref table.


jberkenbilt (approved):
Looks like a good change. I'll wait for CI.

m-holger:
Probably worthwhile reviewing other interfaces as well. I suspect most ought to be changed.

m-holger:
Included in #835

PR #835, opened 2022-12-06
Alternative fix logic for fixDanglingReferences

m-holger:
This implementation is slightly more efficient than #825 (~0.75%).

This implementation ignores the "force" parameter. All fixDanglingReferences (both #825 and this implementation) does is resolve all objects in the xref table. Once this has been done, forcing  fixDanglingReferences to run again has no effect as calling resolve on an already resolved object results in an immediate return.


jberkenbilt (approved):
I like it. If it passes CI, I will merge it.

PR #836, opened 2022-12-07
Update doc comment for QPDF::fixDanglingReferences

m-holger:


jberkenbilt:
On Wed, Dec 7, 2022, at 8:30 AM, m-holger wrote:
> 
> 
>> I didn't count, but do your comments stay within 80 columns?
>> 
> Assuming ./format-code works, yes.
> 

It doesn't touch comments. I have it configured not to do that. It broke too many of my comments when I tried to let it reflow.

>> I think it might be worth considering removing the optional argument for qpdf 12.
>> 
> To my mind, the options are to leave things as they are, or to deprecate the method and make it private in qpd13. Personally, I would just leave it as it is as it doesn't do any harm.
> 

Yeah, it's harmless to leave it alone.

m-holger:
That said, I have to make sure that the first comment line is >80 chars.

PR #841, opened 2022-12-13
Correct syntax of example shell test

AlisterH:
Here is the output showing that the current example code produces an error, and the corrected syntax doees not:
```
#  if [ qpdf --requires-password file.pdf ]; then
>     echo "testing"
>  fi
-bash: [: --requires-password: binary operator expected
#  if qpdf --requires-password file.pdf; then
>  echo "testing"
> else
>  echo "testing2"
> fi
testing2
```

### DIFF
--- a/manual/cli.rst
+++ b/manual/cli.rst
@@ -2917,7 +2917,7 @@ Related Options
 
    .. code-block:: bash
 
-      if [ qpdf --requires-password file.pdf ]; then
+      if qpdf --requires-password file.pdf; then
           # prompt for password
       fi
 


jberkenbilt:
Yes, you are correct. Not sure how I missed that. The perils of not automatically testing your documentation...

jberkenbilt:
Please don't create pull requests to `stable`. That's a documentation branch. Pull requests should be made to main, which is the default branch. I merged this into the wrong branch, but I fixed it manually.

jberkenbilt:
Did readthedocs help you create the PR? If so, this is something I guess I need to be aware of.

AlisterH:
> Did readthedocs help you create the PR? If so, this is something I guess I need to be aware of.

Yes, I used the "edit on Github" button.

PR #844, opened 2022-12-15
Refactor setting and getting of object descriptions

m-holger:
Set parsed offset at the same time as setting description.

### DIFF
--- a/libqpdf/QPDF.cc
+++ b/libqpdf/QPDF.cc
@@ -1848,9 +1848,7 @@ QPDF::resolve(QPDFObjGen og)
     }
 
     auto result(this->m->obj_cache[og].object);
-    if (!result->hasDescription()) {
-        result->setDescription(this, ("object " + og.unparse(' ')));
-    }
+    result->setDefaultDescription(this, og);
 }
 
 void
@@ -1946,10 +1944,7 @@ QPDF::resolveObjectsInStream(int obj_stream_number)
 QPDFObjectHandle
 QPDF::newIndirect(QPDFObjGen const& og, std::shared_ptr<QPDFObject> const& obj)
 {
-    obj->setObjGen(this, og);
-    if (!obj->hasDescription()) {
-        obj->setDescription(this, "object " + og.unparse(' '));
-    }
+    obj->setDefaultDescription(this, og);
     return QPDFObjectHandle::Factory::newIndirect(obj);
 }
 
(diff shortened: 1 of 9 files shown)


m-holger:
The only place I could find that actually uses parsed offsets is in test_parsedoffset.cc. 

I am seeing around 8% improvement.

jberkenbilt:
If you keep improving qpdf's performance, it will start taking negative time, and when you need extra time to do something, you can just run qpdf on a big file and get the savings back.

jberkenbilt:
I remember that the parsed offset thing was added to support a use case of someone doing processing on a PDF file by byte offset based on information gleaned from the parser callbacks. I don't remember the exact use case. I could track it down. I haven't looked at the PR yet, but I'd be less concerned about changing the numbers in error messages than the numbers reported by the parser. Anyway, I'll have a look soon. Things have been excessively busy for me.

m-holger:
> I don't remember the exact use case. I could track it down. 

I don't think it matters. It's 2 lines of code in either getDescription or getParsedOffset, and neither of them gets called all that much.

jberkenbilt:
At a glance, this looks great. I will study it carefully when I get a chance. Thanks.

m-holger:
Is there any particular reason why objects that are not owned by a QPDF object are deemed not to have a description?

jberkenbilt:
> Is there any particular reason why objects that are not owned by a QPDF object are deemed not to have a description?

Not that I can recall. I guess originally objects only got descriptions when being parsed and the only way to parse was when reading from a file, but I'm just guessing.

jberkenbilt:
Given your recent activity, would you like me to pause before reviewing this and the other more significant PR? I merged the little ones and then later noticed that one of the other ones incorporated those as well. github's automatic rebase will deal with it but I don't know whether the other ones being merged may change anything about these.

m-holger:
I would find it more helpful to get 844 and 858 reviewed as soon as you have a change. I started squashing PRs because I am getting to many merge conflicts I am working from main.

jberkenbilt (approved):
This all looks good. I see why you're asking about why qpdf has to be set for there to be a description. I doubt there's a reason. If that constraint is not needed, feel free to remove it. I'm sure you can study the code to make sure there aren't cases where the code is checking for that as a proxy for checking for a qpdf. I hope that doesn't happen.

PR #853, opened 2022-12-21
Fix QPDFObjectHandle::isScalar

m-holger:
Exclude uninitialized, destroyed and reserved objects.


jberkenbilt (approved):
Good change. We might add more non-scalars, but we'll almost surely never add another scalar.

PR #857, opened 2022-12-30
Refactor QPDFWriter::enqueueObject

m-holger:
Move test for m->linearized outside loop.


jberkenbilt (approved):
I wonder why I did it the other way. Probably happened when I added linearization (long before qpdf was open source).

PR #858, opened 2022-12-31
Refactor QPDFParser::parse

m-holger:
The main changes are:
- delay creation of QPDFObjectHandles until parse returns or objects are inserted into QPDF_Arrays/QPDF_Dictionaries.
- avoid unnecessary copying of maps and vectors.
- avoid the overhead of calling QPDFValue methods through QPDFObjectHandles methods.
- change to dictionary processing loop.

### DIFF
--- a/libqpdf/QPDFParser.cc
+++ b/libqpdf/QPDFParser.cc
@@ -404,7 +404,7 @@ QPDFParser::parse(bool& empty, bool content_stream)
                         QPDFObjectHandle::newString(frame.contents_string);
                     dict["/Contents"].setParsedOffset(frame.contents_offset);
                 }
-                object = QPDF_Dictionary::create(dict);
+                object = QPDF_Dictionary::create(std::move(dict));
                 setDescription(object, offset - 2);
                 // The `offset` points to the next of "<<". Set the rewind
                 // offset to point to the beginning of "<<". This has been
--- a/libqpdf/QPDFParser.cc
+++ b/libqpdf/QPDFParser.cc
@@ -441,14 +441,14 @@ QPDFParser::setDescription(
 }
 
 void
-QPDFParser::warn(QPDF* qpdf, QPDFExc const& e)
+QPDFParser::warn(QPDFExc const& e) const
 {
     // If parsing on behalf of a QPDF object and want to give a
     // warning, we can warn through the object. If parsing for some
     // other reason, such as an explicit creation of an object from a
     // string, then just throw the exception.
-    if (qpdf) {
-        qpdf->warn(e);
+    if (context) {
+        context->warn(e);
     } else {
         throw e;
     }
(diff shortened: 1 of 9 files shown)


jberkenbilt:
Feel free to ping me when merge conflicts are resolved and if you decide to make changes in light of the discussion in #844.

m-holger:
Rebased and good to go.

libqpdf/QPDFParser.cc
```diff
@@ -191,20 +191,20 @@ QPDFParser::parse(bool& empty, bool content_stream)
                 if (content_stream) {
                     object = QPDF_Operator::create(value);
                 } else if (
-                    (value == "R") && (state != st_top) && (size >= 2) &&
-                    (!olist.back().isIndirect()) &&
```
jberkenbilt, line 180: Why is it okay to remove the indirect check? I think this was to prevent bogus stuff like 10 0 R 0 R where 10 0 R resolves to an integer. By making olist take a list of object pointers, do we lose the ability to check that?

jberkenbilt: Looking through the code, I'm thinking the previous refactorings have actually made that case impossible or such that it won't show up that way anymore.

jberkenbilt (approved):
This seems fine. Can you confirm that you have considered by doubt? If you can demonstrate that the change to remove the indirect check is safe, them I'm good with this. I haven't scrutinized the memory use with move semantics carefully. I'm relying on this being a strictly internal API along with the sanitizer checks to make sure it's right. When you std::move something out of a shared pointer, the original shared pointer just turns into nullptr, right? I haven't personally made use of std::move though I understand what it does conceptually and know that the object you move from is considered invalid after the move, so I just want to make sure we're not moving out of things that might still be referenced.

jberkenbilt:
I'm prepared to merge this as is once you clear up a few of my questions, unless investigating my questions leads you to decide to make some changes.

libqpdf/QPDFParser.cc
m-holger: I thought the check was there to prevent the call to isInteger accidentally resolving an indirect object, which would have crashed the parser/qpdf.

10 0 R 0 R may or may not be interpreted as as an indirect reference depending on whether 10 0 R had been resolved prior to calling parse. I think check for indirect integers ought to go back in.

jberkenbilt: We should make sure there's actually a test for it. That code has probably always been there and was probably not added in response to a bug, so it may not have occurred to me to create a test for it.

jberkenbilt (changes requested):
Changing to "request changes" so I don't forget we have to resolve the isIndirect thing.

jberkenbilt (approved):

m-holger:
I have added the checks back in. I look into adding an extra test for "10 0 R 0 R" separately.

std::move(x) simply flags to the compiler that you will not use the variable x again. If x is a shared ptr, the compiler may copy it to the destination without increasing the ref count while simultaneously setting x to nullptr (without decreasing the ref count).

jberkenbilt:
> I have added the checks back in.

This could be as simple as using the contextual form of parse in some test in test_driver.cc. Maybe there's even an existing test where it would fit. I'm thinking adding a couple of lines to test 31 would be the quickest way.

jberkenbilt:
The comment in test 31 that says the input file is not used is lying.

m-holger:
I'll have a look. Dinner time.

jberkenbilt:
I added a test to test 31. I had to push up a change for the consistent download-external-libs failure, which was rate limiting from github (as I guessed it would be).

PR #859, opened 2022-12-31
Add QPDFObjectHandle methods to support the interaction with QPDFObject and QPDFValue

m-holger:


jberkenbilt:
Does removal of Factory and ObjAccessor achieve anything other than simplifying other code? Can you convince me of the value of having public methods that people aren't supposed to call? It's probably fine in that they are not marked with `QPDF_DLL` which means they are not usable from the Windows DLL or non-Windows shared library and reference incomplete types that are not part of the public API. However someone linking statically could still call these.

Adding operator bool requires bumping to 11.3. Either you can do it, or I can do it after merging. README-maintainer has instructions for doing it. I've considered scripting it but have never bothered. CI will fail if it's done wrong. Search README-maintainer for "version numbers are consistent" for the list of the three places that have to be updated.

m-holger:
> However someone linking statically could still call these.

But why would they?  What could they do after calling them?

jberkenbilt:
Meh, I can't really think of a good reason not to do it. There is value to cleaner code. There's a merge conflict now but I am good to go with this. I feel like the fact that you want this and that it helps reduce your friction in making helpful changes to the library is really adequate justification.

jberkenbilt (approved):

jberkenbilt (approved):

m-holger:
@jbarlow83 

The operator
```cpp
QPDFObjectHandle::operator bool() const
{
    return obj != nullptr;
}
``` 
introduced in this PR causes a pikepdf test failure and had to be reverted out for qpdf 11.3.0

The motivation for introducing the bool operator is that it effectively provides us with a zero-cost version of a `std::optional<QPDFObjectHandle>`, which in turn allows us to write code like 

```cpp
QPDFObjectHandle
QPDFObjectHandle::getArrayItem(int n)
{
    resolve();
    if (auto item = obj->at(n)) {
        return item;
    } else {
        // handle failure
    }
}
```
saving many costly range and type checks as a result. I would therefore like to reintroduce the operator for 10.4.0 if this can be done without causing problems for pikepdf.

Would you be able to look into the reason for failure and let me know whether this is easily fixable?

jbarlow83:
I did a little triage. pikepdf test failure is:
```
    def test_objectlist_repr(pal):
        cs = pikepdf.parse_content_stream(pal.pages[0].Contents)
        assert isinstance(cs[1][0], pikepdf._core._ObjectList)
        ol = cs[1][0]
>       assert (
            "[Decimal('144.0000'), 0, 0, Decimal('144.0000'), Decimal('0.0000'), Decimal('0.0000')]"
            in repr(ol)
        )
E       assert "[Decimal('144.0000'), 0, 0, Decimal('144.0000'), Decimal('0.0000'), Decimal('0.0000')]" in '_ObjectList[1, 1, 1, 1, 1, 1]'
E        +  where '_ObjectList[1, 1, 1, 1, 1, 1]' = repr(_ObjectList[1, 1, 1, 1, 1, 1])

tests/test_objectlist.py:21: AssertionError
```

This is checking for consistent parsing of QPDFObject/Value derived from content streams.

Probably fixable. I'd say there's a decent possibility that qpdf is affected too. Can't look into further at the moment I'm afraid.

m-holger:
@jbarlow83 , I am rather out of my depth here, but I have been able to solve the test failure although not the underlying problem.

If I duplicate the `_ObjectList.__repr__` method

```python
    py::bind_vector<ObjectList>(m, "_ObjectList") // Autoformat fix
        .def("__repr__", [](ObjectList &ol) {
            std::ostringstream ss;
            bool first = true;
            ss << "pikepdf._core._ObjectList([";
            for (auto &h : ol) {
                if (first) {
                    first = false;
                } else {
                    ss << ", ";
                }
                ss << objecthandle_repr(h);
            }
            ss << "])";
            return ss.str();
        })

        .def("repr", [](ObjectList &ol) {
            std::ostringstream ss;
            bool first = true;
            ss << "pikepdf._core._ObjectList([";
            for (auto &h : ol) {
                if (first) {
                    first = false;
                } else {
                    ss << ", ";
                }
                ss << objecthandle_repr(h);
            }
            ss << "])";
            return ss.str();
        });
```
and then set the `_ObjectList.__repr__`  method inside the test

```python
def test_objectlist_repr(pal):
    cs = pikepdf.parse_content_stream(pal.pages[0].Contents)
    assert isinstance(cs[1][0], pikepdf._core._ObjectList)
    ol = cs[1][0]

    pikepdf._core._ObjectList.__repr__ = pikepdf._core._ObjectList.repr

    assert (
        "[Decimal('144.0000'), 0, 0, Decimal('144.0000'), Decimal('0.0000'), Decimal('0.0000')]"
        in repr(ol)
    )
```
the test passes.

`cs` and `ol` definitely hold the correct values (and `repr(cs[1][0][0]` correctly returns `Decimal('144.0000')`), but during the call to `repr` pybind11 attempts to convert object handles inside the `_ObjectList`  into booleans and than convert the booleans into integers. 

I have no idea whether this is good news or bad news.

m-holger:
@jbarlow83 

Did you have a chance to look into this any further? I noticed you marked the test in question as 'expected fail' on 12 April.

m-holger:
@jbarlow83 

Furthermore, the test now passes with `QPDFObjectHandle::operator bool() ` enabled.

PR #862, opened 2023-01-01
Do double indirect test correctly

jberkenbilt:
I realized I had done it wrong before....parsing a string stops after the first successful object, so the error was from the exception in the assertion. That's why other tests wrap stuff in [].


jberkenbilt:
@m-holger FYI

PR #863, opened 2023-01-01
Refactor QPDF_Array

m-holger:
The main changes are:

- Merge SparseOHArray into QPDF_Array given that they are closely coupled and that there is no longer a reason to keep them separate.
- Change index type to int in order to avoid unnecessary conversions.
- Change underlying data structure to std::map as it turned out to be on balance more efficient here.
- Make most methods virtual in order to avoid costly dynamic casting.
- Eliminate repeated checks. Most method now return false or an uninitialized object handle to indicate failure rather than throwing an exception.
- Reduce copying by using move semantics and by shifting elements in place during insert/erase.
- Move some methods from QPDFObjectHandle to QPDFValue for efficiency.
- Add additional ownership checks and object warnings.

I have renamed most methods to be more in line with std::vector. I have also dropped the "this->". Given that there are only two data members and the methods are all short I do not feel that "this->" improves readability in this case.


m-holger:
Ignore for now

m-holger:
I have reviewed whether it is feasible to change SparseOHArray::elements from a map to vector<shared_ptr<QPDFObject>> with direct nulls represented by nullptr, given that this would reduce the overhead considerably compared to storing actual QPDF_Nulls and given that #844 reduces space requirement. I found that this would still involve an unacceptable space penalty in the "many-nulls" case (compared to 10.2, x 20 for the many-nulls test file, x 6 for an actual test file).

libqpdf/QPDFValue.cc
```diff
@@ -43,3 +44,15 @@ QPDFValue::objectWarning(std::string const& warning) const
...
 }
+
+// Array methods
+
+int
+QPDFValue::size(bool type_warn, bool allow_null) const
+{
+    if (type_warn) {
+        typeWarning("array", "treating as empty");
+        QTC::TC("qpdf", "QPDFObjectHandle array treating as empty");
```
jberkenbilt: This isn't QPDFObjectHandle anymore, so the coverage case text should be updated.

libqpdf/QPDFValue.cc
```diff
@@ -43,3 +44,15 @@ QPDFValue::objectWarning(std::string const& warning) const
 {
     warn(QPDFExc(qpdf_e_object, "", getDescription(), 0, warning));
 }
+
+// Array methods
+
+int
+QPDFValue::size(bool type_warn, bool allow_null) const
+{
+    if (type_warn) {
```
jberkenbilt, line 66: This makes me a little uncomfortable. Doing type checks inside QPDFValue makes me feel like we're slipping a little bit into QPDFObjectHandle land. If QPDFObject/QPDFValue is at least partly intended to be the basis for a more typesafe object handle, does it make more sense to just leave these checks in QPDFObjectHandle? That seems like the right layer to address type issues.

libqpdf/qpdf/QPDFObject_private.hh
```diff
@@ -165,6 +160,19 @@ class QPDFObject
...
 
+    // Array methods
+
+    int
+    size(bool type_warn = false, bool allow_null = false) const
+    {
+        return value->size(type_warn, allow_null);
+    }
+    QPDFObjectHandle
+    at(int n, bool warn = false, bool allow_null = false) const
```
jberkenbilt, line 176: Same as size -- I feel like this is heading a bit in the wrong direction regarding type safety. I'd rather see the clutter in QPDFObjectHandle of asArray(), isArray(), etc. -- it makes it more obvious inside the abstraction layer that's responsible for type checking.

jberkenbilt (changes requested):
I stopped reviewing part way through the commit history because there are parts of this that I feel are moving in the wrong direction and wanted to give you a chance to explain before I spend too much time on this.

I feel like moving type checks from QPDFObjectHandle into QPDFValue is moving in the wrong direction. I don't think that QPDFValue or its derived classes should really have to be concerned about when they are being accessed like arrays. I think it would be better to keep type checks in QPDFObjectHandle, let QPDFValue remain a minimal base class, and let each type of QPDFValue only be focused on what kind of type it actually is. Otherwise we run the risk of having QPDFValue turn into what QPDFObjectHandle is except having all the type checking code distributed throughout all the classes where it's harder to find, see, understand, fix, etc.

I think the part of this that merges SparseOHArray into QPDF_Array as well as renaming the methods of QPDF_Array to be more in line with the standard library are good.

If you disagree, let's discuss.

libqpdf/QPDFValue.cc
jberkenbilt: but I don't really think this should be here -- I think it should probably stay in QPDFObjectHandle.

jberkenbilt:
> I have reviewed whether it is feasible to change SparseOHArray::elements from a map to vector<shared_ptr> with direct nulls represented by nullptr, given that this would reduce the overhead considerably compared to storing actual QPDF_Nulls and given that #844 reduces space requirement. I found that this would still involve an unacceptable space penalty in the "many-nulls" case (compared to 10.2, x 20 for the many-nulls test file, x 6 for an actual test file).

It might be possible to do something like start with a regular std::vector and then switch over if the vector gets past a certain size. I have considered that implementation but never actually tried it. It would add complexity to the code, but if  if the size checks aren't too costly, maybe it would make it so you only had to pay the time penalty of the sparse array when it actually matters. We could experiment with what a good threshold is, but my first instinct is to use something like 500. I think arrays larger than this are pretty unusual, but arrays in the 200+ range appear in font maps, which are very common. This is definitely a case where we'd be adding a little complexity in the name of performance. This is something I have not historically done, but it seems consistent with your efforts to optimize the parts of the code whose functionality is extremely stable and tested.

jberkenbilt:
I have ideas about how one might implement my suggestion efficiently, but I'll wait and see if you agree/want to spend time on it, and if so, we'll see what what you come up with first since you're ideas on efficiency are often better than mine.

libqpdf/QPDFValue.cc
m-holger: All the QTC labels are tidied in the final commit.

jberkenbilt: Missed that, thanks.

m-holger:
I would not been keen on that. I looked into this approach when I worked on #749 and found that it was not beneficial. My main concern would be that the additional complexity would make it more difficult to achieve other savings.

libqpdf/QPDFValue.cc
jberkenbilt: To clarify, with this change, something like `asInteger()->size()` moves back from being a compile-time error to being a runtime error, and I thought one of the main thrusts here was to get rid of that kind of thing.

jberkenbilt:
That was my feeling as well. We can just leave it as a map. That's been my inclination all along.

m-holger:
I have pushed up a version that leaves the error-checking/handling in QPDFObjectHandle.

libqpdf/QPDFValue.cc
m-holger: It will always be a runtime error, asInteger will return a nullptr. Using virtual methods rather than dynamic casting is just a less expensive way of generating the runtime error .

jberkenbilt: if asInteger() returns a pointer to a type that doesn't have a size() method, it will be a compile error, won't it? 
```c++
#include <memory>

class Base {
};

class D1: public Base {
  public:
    void f()
    {
    }
};

class D2: public Base {
};

void f() {
    auto b1 = std::make_shared<D1>();
    dynamic_cast<D1*>(b1.get())->f();
    dynamic_cast<D2*>(b1.get())->f();  // <-- compile error
}
```
To me, this is worth the overhead of dynamic cast. To code it correctly, you'd have to know the thing is an array or call isArray().

I suppose another way to do it that avoids dynamic cast is to use a child struct with one of each type.Each derived class's constructor could assign into this protected member variable in the base class. The code would be a little ugly for this, and each individual `as` would probably need to be done with template specialization, but it would avoid the dyanmic cast. It depends on what you're willing to trade to remove the dynamic cast overhead.

#include <memory>
#include <stdexcept>
#include <iostream>
#include <cassert>

class D1;
class D2;

class Base {
  public:
    template <typename T>
    T* as()
    {
        // Provide a template specialization for each derived class
        // that returns a value from child
        throw std::logic_error("no specialization provided");
    }

  protected:
    struct {
        // Each derived class must have a field here and set it to
        // `this` in its constructor.
        D1* d1{nullptr};
        D2* d2{nullptr};
    } child;
};

class D1: public Base {
  public:
    D1() {
        this->child.d1 = this;
    }

    void f()
    {
        std::cout << "D1::f" << std::endl;
    }
};

template <>
D1* Base::as()
{
    return child.d1;
}

class D2: public Base {
};

template <>
D2* Base::as()
{
    return child.d2;
}

int main() {
    auto b1 = std::make_shared<D1>();
    b1->as<D1>()->f();
//    b1->as<D2>()->f(); // <- compile error
    assert(b1->as<D2>() == nullptr);
    return 0;
}
```
It's kind of a C-style roll-your-own inheritance.

So I guess the questions, which of these would we rather see?
1. Overhead of RTTI for dynamic cast
2. Virtual methods (loss of compile-time checking for type-specific methods, distribution of error checking)
3. Extra boiler-plate in each class to keep type-safety and avoid RTTI

I guess I'd go with 1 or 3 depending on how much overhead RTTI has. Looking at my sample implementation, I don't think 3 is really all that bad from a code ugliness angle.

Comment (jberkenbilt) [reply libqpdf/QPDFValue.cc]: (By `asInteger()` and `asArray()`, I mean `as<QPDF_Integer>()` and `as<QPDF_array>()`. I momentarily forgot they had been replaced by a templated function.)

Comment (jberkenbilt) [reply libqpdf/QPDFValue.cc]: With this implementation, templated `as` isn't really natural. Here's a simpler version that reverts to the specific `as` methods:
```c++
#include <memory>
#include <stdexcept>
#include <iostream>
#include <cassert>

class D1;
class D2;

class Base {
  public:
    // Implement for each derived class to return the right value from child
    D1* asD1() {
        return child.d1;
    }
    D2* asD2() {
        return child.d2;
    }

  protected:
    struct {
        // Each derived class must have a field here and set it to
        // `this` in its constructor.
        D1* d1{nullptr};
        D2* d2{nullptr};
    } child;
};

class D1: public Base {
  public:
    D1() {
        this->child.d1 = this;
    }

    void f()
    {
        std::cout << "D1::f" << std::endl;
    }
};

class D2: public Base {
};

int main() {
    auto b1 = std::make_shared<D1>();
    b1->asD1()->f();
//    b1->as<D2>()->f(); // <- compile error
    assert(b1->asD2() == nullptr);
    return 0;
}
```

jberkenbilt:
I won't have more time to look at this today.

m-holger:
> So I guess the questions, which of these would we rather see?
>
>    Overhead of RTTI for dynamic cast
    Virtual methods (loss of compile-time checking for type-specific methods, distribution of error checking)
    Extra boiler-plate in each class to keep type-safety and avoid RTTI

My answer in this case would be virtual methods. The reason for this would be two-fold

#### Compile time checks

Any compile time checks only apply to the implementation of QPDFObjectHandle itself and any other core components of qpdf that access QPDFValue directly such as QPDFParser. These components are both extensively tested and are performance critical.

Compile time checks/type safety for library users, including the non-core parts of qpdf itself, will need to be build on top of QPDFObjectHandle.

#### Higher level constructs

The trigger for creating this PR are was me drafting #866 , which will make the case for providing array methods that not only work with arrays but also  with other object types. This will require array methods to be virtual.

jberkenbilt:
okay, if you want to go back to your original implementation with virtual methods, I'm okay with it.

jberkenbilt:
actually, I see they are still virtual methods. I'll re-review when I get a chance. Not today, but I'll try not to put it off too long.

m-holger:
This is a major rewrite of my previous attempt. The major change is that I am using vectors as the underlying data structure except where it is absolutely necessary to use sparse arrays. This has a positive performance impact for normal files and a very significant impact for "many nulls" case.  

Sparse arrays are only used for arrays created by QPDFParser (for arrays with more than 100 nulls). The only other place where I see a need for the creation of sparse arrays is in QPDF_json.

I have eliminated duplicated range checks, including on conversion between int and size_t (keeping in mind that vector sizes can not exceed the max of int). Checks for uninitialized object handles take place on array creation and element insertion and is combined with the foreign qpdf check. For performance reasons, the code for the check is included in QPDF_Array. The checks are omitted for arrays created by QPDFParser for performance and since they are safe.

jberkenbilt:
Is this ready for review now and acceptance if I agree with it?

m-holger:
Yes. I will be out for the rest of the afternoon.

jberkenbilt (approved):
I'm happy with this implementation and with the changes to array's internal API without exposing anything or breaking QPDFObjectHandle's API. Someday when I want to make sure I am using std::move in the best way, I will back and study this further.

PR #866, opened 2023-01-03
Background to PR863

m-holger:
(very early draft of a proposal for a qpdf 12 enhancement)

### Add array style methods to QPDFObjectHandle that work for all object types

It is common for directory entries that allow for a list of objects to also allow a single object of the (same) type, as well as maybe a null  object. For example,  the `/Filter` entry in a stream dictionary is specified as an optional name, or an array of 0, 1, or more names.

Users of the qpdf library should not need to be concerned with the detail. Instead, they should be able to process the result of `getKey("/Filter") ` as if it was an array of names. 

...

As a demo of the concept I have implemented simple size, at, begin and end methods and used them with qpdf itself. This is to test whether the methods are practical and sufficiently efficient for general use (but not necessarily for performance critical parts of qpdf itself).


m-holger:
Closed as this is out of date.

PR #867, opened 2023-01-03
Add private namespace QUtil_12

m-holger:
This is an alternative solution to adding temporary functions to QUtil in #864. Instead, this proposes to create a complete private copy of QUtil that can be incrementally updated for qpdf  12 and used in its entirety to replace QUtil when it comes to the ABI update.

On balance this feels like a cleaner solution, both making it easier to add additional tweaks to QUtil that may be required to allow other parts of qpdf take advantage of new features of C++-17, and making it easier to finally switch to qpdf 12.

It is also about 1% faster than the equivalent commits in #864.


jberkenbilt:
My impulse here would be to add the new header file only and to commingle the QUtil_12 implementations and tests with the regular QUtil ones. I feel like that will make the history easier to follow and will avoid adding and removing entire files.

m-holger:
I had a play around. On balance it looks to me that replacing entire file maintains a better blame history.

I've added a temporary commit to see that state after migrating to qpdf 12.

jberkenbilt:
I've given some thought to this, though I'm still open to discussion. My position is this: I want `std::string_view` in the API, but I don't want to do it gradually. I would like to see, instead, a transition similar to the `PointerHolder` transition, though this one will be easier since it will not require code changes for most people. I would also like to avoid adding clutter to the internal API including temporary overloads or temporary namespaces. I'm fine with some replacement of `std::string const&` with `std::string_view` in the internal API where it doesn't break existing code and in using it for new things. I definitely don't want to add temporary overloads or temporary namespaces to the public API.

Thought it's been less time than usual for a major release, I am willing to release qpdf 12 specifically for a `std::string_view` transition since I think this will have enough benefit to make it worthwhile, and it will prevent us from having to perpetuate use of `std::string const&` when a better option is available. The way I would like to see this done is that we move the entire API, wherever it makes sense, from `std::string const&` to `std::string_view` in a side branch, and when we're confident that it's right and that we're not going to need to make more significant changes, we can open the window for qpdf, merge that, and make any other ABI changes that we have in the queue. It's okay with me if this happens before my `pages` work. If `std::string_view` goes in while my `pages` work is in progress, I'm happy to do a one-time pass through that work to bring it in sync.

As for timing, I'd say go ahead and work on the side branch, and when you think it's ready, we can discuss. We don't have to arbitrarily put it off.

What do you think of that proposal?

m-holger:
Works for me

m-holger:
Superseded by decision re qpdf 12.

PR #869, opened 2023-01-06
Add new method Pipeline::writeStr

m-holger:
For performance

### DIFF
--- a/libqpdf/QPDFWriter.cc
+++ b/libqpdf/QPDFWriter.cc
@@ -981,9 +981,9 @@ QPDFWriter::writeBinary(unsigned long long val, unsigned int bytes)
 }
 
 void
-QPDFWriter::writeString(std::string const& str)
+QPDFWriter::writeString(std::string_view str)
 {
-    this->m->pipeline->writeString(str);
+    this->m->pipeline->writeStr(str);
 }
 
 void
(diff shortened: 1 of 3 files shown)


libqpdf/QPDFWriter.cc
```diff
@@ -983,7 +983,7 @@ QPDFWriter::writeBinary(unsigned long long val, unsigned int bytes)
 void
 QPDFWriter::writeString(std::string const& str)
```
jberkenbilt, line 984: Does it make sense for this to take `std::string_view` too?

jberkenbilt (approved):
I'm fine with this change but will wait for your response before merging.

libqpdf/QPDFWriter.cc
jberkenbilt: There's a case to be made for having part of qpdf 12 be to replace `std::string const&` with `std::string_view` in as many places as possible and to live with the slight performance penalty in the meantime to avoid lots of API clutter.

m-holger: > Does it make sense for this to take std::string_view too?

Agree. Updated

> There's a case to be made for having part of qpdf 12 be to replace std::string const& with std::string_view 

Agree

m-holger:
Superseded by decision re qpdf 12.

PR #871, opened 2023-01-08
Tune QPDFWriter

m-holger:
A couple of minor tweaks for performance

### DIFF
--- a/include/qpdf/QPDFWriter.hh
+++ b/include/qpdf/QPDFWriter.hh
@@ -751,7 +751,8 @@ class QPDFWriter
         std::string cur_data_key;
         std::list<std::shared_ptr<Pipeline>> to_delete;
         Pl_Count* pipeline;
-        std::list<QPDFObjectHandle> object_queue;
+        std::vector<QPDFObjectHandle> object_queue;
+        size_t object_queue_front{0};
         std::map<QPDFObjGen, int> obj_renumber;
         std::map<int, QPDFXRefEntry> xref;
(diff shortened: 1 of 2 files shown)


jberkenbilt (approved):
Looks fine. Is there an advantage of using getArrayAsVector and getDictAsMap over using aitems() and ditems()? It seems that the latter would avoid an extra copy.

m-holger:
A 10 - 15% performance hit.

The chance of looking at aitems and ditems is one of the reasons why I'm reflecting on #863. They should be faster than getting maps / vectors.

jberkenbilt:
> The chance of looking at aitems and ditems is one of the reasons why I'm reflecting on #863. They should be faster than getting maps / vectors.

I would obviously welcome improvements to aitems and ditems. I'll go ahead and merge this as is, but if you fix aitems and ditems, then this can be switched. I guess you would probably search for calls to getArrayAsVector and getDictAsMap. I would like aitems and ditems to be as efficient as iterating over the underlying objects (noting that QPDF_Array is a really a map, but to the extent possible). I really only added the getAsArray and getAsMap methods to make it easier to iterate (way back in the early days) without regard for performance.

jberkenbilt (approved):

m-holger:
> ...  if you fix aitems and ditems ...

I think this will be ABI breaking

jberkenbilt:
> I think this will be ABI breaking

Then, if successful, feel free to add to #785.

PR #876, opened 2023-01-13
Refactor QPDFObjectHandle::ditems

m-holger:
This changes how QPDF_Dictionary handles insertion of QPDF_Nulls / erasure of elements in order to make safe iteration over dictionaries more efficient.

QPDFDictionaryItems has been refactored to access items through iterators of the underlying map in QPDF_Dictionary.

A replacement has been added which returns values rather than references to a copy of the underlying dictionary item. Returning references to a copy has a number of subtle problems, including references returned by the iterator being affected by subsequent iterator operation of this or any other iterator of the same QPDFDictionaryItems instance.  The change was not included in QPDFDictionaryItems since it is API breaking.

Use of dItems is more efficient than using getDictionaryAsMap.

Postfix in/decrement operators of QPDFDictionaryItems are broken. This has not been fixed since it has not created problems in practice and since dItems should be used in preference to ditems.

### DIFF
--- a/examples/pdf-mod-info.cc
+++ b/examples/pdf-mod-info.cc
@@ -31,7 +31,7 @@ dumpInfoDict(
 {
     QPDFObjectHandle trailer = pdf.getTrailer();
     if (trailer.hasKey("/Info")) {
-        for (auto& it: trailer.getKey("/Info").ditems()) {
+        for (auto it: trailer.getKey("/Info").dItems()) {
             std::string val;
             if (it.second.isString()) {
                 val = it.second.getStringValue();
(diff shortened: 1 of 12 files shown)


m-holger:
The refactored ditems is significantly more efficient than the original. With the original, the penalty for using ditems instead of getDictionaryAsMap in QPDFWriter is around 11% increase in runtime. The refactored version brings this down to around 2%.

jberkenbilt:
I think we should just fix ditems to return values rather than references. Returning references to items in the dictionary feels like a bug, and there's no valid use case for it. I'd document it as a behavior change, but I think it's sufficiently unlikely to cause a problem to tolerate it since the old behavior is potentially harmful. So I guess I'd like to see what you are proposing as dItems just be ditems and mark this as for qpdf 12. If there are small API breaking changes to replace broken behavior, that is tolerable.

I doubt I will have time to review it this weekend.

m-holger:
The change will mean that some existing programs will not compile, i.e `auto& item: oh.ditems()` will need to be changed to  `auto item: oh.ditems()` or  `auto&& item: oh.ditems()`.

jberkenbilt:
Hmm, okay. Maybe we shouldn't break that. I'll have to give this a closer look.

PR #877, opened 2023-01-17
Inline QIntC functions

m-holger:
For performance


m-holger:
This requires further consideration. In combination with #863 and #876 in-lining of QIntC functions appears to have a negative impact .

m-holger:
This is good to go as far as I'm concerned. #863 and #876 need more work.

jberkenbilt (approved):

PR #878, opened 2023-01-18
Tidy QPDFXRefEntry creation

m-holger:


jberkenbilt:
I'm not really a fan of having lots of overloaded constructors like this. To me, it actually reduces clarity/understandability of the code. I'd rather see a single comment near the single constructor explaining what the fields mean under various conditions if the code is hard to understand without it.

m-holger:
To my mind, one of the roles of the constructor is to enforce the invariants, ideally at compile time. Having separate constructors enforce that only type 1 records have an offset, only type 2 records have a stream number and index and prevent at compile time the creation pf type 3 records.

m-holger:
Also, it improves run times on average by around 0.5%.

jberkenbilt (approved):

jberkenbilt:
Upon reflection, I decided that what I didn't like was C++ constructors and overloaded functions, not your change. I'd rather see QPDFXrefEntry.newType1(), etc. but I'd have to pick a different language for that. Anyway, you've earned the right to have influence over the API even if you do it differently from how I would do it. I am grateful for your numerous contributions.

PR #882, opened 2023-01-27
Refactor handling of structural character in JSONParser

m-holger:
Key changes:
- introduce lexer states for each of the structural tokens
- use early return from handleToken to avoid unnecessary processing and make it easy to understand the code
- move validation checks into the two main processing switch statements
- combine the stacks


libqpdf/JSON.cc
```diff
@@ -868,6 +874,27 @@ JSONParser::getToken()
                 lex_state = ls_string;
             } else if (QUtil::is_space(*p)) {
                 action = ignore;
+            } else if (*p == ',') {
+                lex_state = ls_comma;
+                ready = true;
+            } else if (*p == ',') {
```
m-holger, line 894: Duplicate test for comma - test will disappear in #889

PR #883, opened 2023-01-27
Refactor lexing of strings in JSONParser

m-holger:
Process strings in a single pass


libqpdf/JSON.cc
```diff
@@ -623,12 +623,10 @@ namespace
         void getToken();
         void handleToken();
         void numberError();
-        static std::string
-        decode_string(std::string const& json, qpdf_offset_t offset);
```
m-holger, line 631: Stray comment needs remving

PR #890, opened 2023-01-29
Eliminate the use of shared pointers in JSONParser

m-holger:


jberkenbilt (approved):
I finally had a chance to scrutinize. Looks good. Error offsets are better than before. Sorry it took me so long. I wanted to give it the attention it deserved.

PR #891, opened 2023-01-30
Replace calls to strchr

m-holger:
For performance

### DIFF
--- a/include/qpdf/QUtil.hh
+++ b/include/qpdf/QUtil.hh
@@ -545,13 +545,15 @@ namespace QUtil
 inline bool
 QUtil::is_hex_digit(char ch)
 {
-    return (ch && (strchr("0123456789abcdefABCDEF", ch) != nullptr));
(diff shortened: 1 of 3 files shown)


jberkenbilt:
Just curious...is this better or worse than a switch statement?

m-holger:
I did not specifically investigate for this case.

I did look into 'if' vs 'switch' when I played with QPDFTokenizer last summer. Both my own experiments and comments on eg stackoverflow suggests that there is no real difference where '==' comparisons are concerned. Where '<' / '>' or 'a < x < b' comparisons are concerned if statements can have the advantage when they allow common cases (eg space or ctrl char '< 33') to be moved to the top rather that being handled  in the default part of a switch statement. (For JSON::encode_string this made the key difference and the other tweaks were just icing on the cake).

Returning a boolean expression theoretically should be more efficient than returning true / false from within an 'if' statement but in practice I would be very surprised if the compiler didn't optimize the if statement away.

jberkenbilt (approved):

PR #896, opened 2023-02-09
Refactor setting of object descriptions in QPDF::JSONReactor

m-holger:
This provides slightly less information in error messages, but improves run time for the JSON input test by around 13%.

### DIFF
--- a/include/qpdf/QPDF.hh
+++ b/include/qpdf/QPDF.hh
@@ -32,6 +32,7 @@
 #include <memory>
 #include <stdio.h>
 #include <string>
+#include <variant>
 #include <vector>
 
 #include <qpdf/Buffer.hh>
@@ -50,6 +51,7 @@ class BitStream;
 class BitWriter;
 class QPDFLogger;
 class QPDFParser;
+struct JSON_Descr;
 
 class QPDF
 {
@@ -1152,6 +1154,7 @@ class QPDF
         QPDF& pdf;
         std::shared_ptr<InputSource> is;
         bool must_be_complete;
+        std::shared_ptr<std::variant<std::string, JSON_Descr>> descr;
         bool errors;
         bool parse_error;
         bool saw_qpdf;
--- a/include/qpdf/QPDF.hh
+++ b/include/qpdf/QPDF.hh
@@ -32,7 +32,6 @@
 #include <memory>
 #include <stdio.h>
 #include <string>
-#include <variant>
 #include <vector>
 
 #include <qpdf/Buffer.hh>
@@ -51,7 +50,6 @@ class BitStream;
 class BitWriter;
 class QPDFLogger;
 class QPDFParser;
-struct JSON_Descr;
 
 class QPDF
 {
@@ -1110,72 +1108,7 @@ class QPDF
         std::set<QPDFObjGen>::const_iterator iter;
     };
 
-    class JSONReactor: public JSON::Reactor
-    {
-      public:
-        JSONReactor(
-            QPDF&, std::shared_ptr<InputSource> is, bool must_be_complete);
-        virtual ~JSONReactor() = default;
-        virtual void dictionaryStart() override;
-        virtual void arrayStart() override;
-        virtual void containerEnd(JSON const& value) override;
-        virtual void topLevelScalar() override;
-        virtual bool
-        dictionaryItem(std::string const& key, JSON const& value) override;
-        virtual bool arrayItem(JSON const& value) override;
-
-        bool anyErrors() const;
-
-      private:
-        enum state_e {
-            st_initial,
-            st_top,
-            st_qpdf,
-            st_qpdf_meta,
-            st_objects,
-            st_trailer,
-            st_object_top,
-            st_stream,
-            st_object,
-            st_ignore,
-        };
-
-        void containerStart();
-        void nestedState(std::string const& key, JSON const& value, state_e);
-        void setObjectDescription(QPDFObjectHandle& oh, JSON const& value);
-        QPDFObjectHandle makeObject(JSON const& value);
-        void error(qpdf_offset_t offset, std::string const& message);
-        QPDFObjectHandle reserveObject(int obj, int gen);
-        void replaceObject(
-            QPDFObjectHandle to_replace,
-            QPDFObjectHandle replacement,
-            JSON const& value);
-
-        QPDF& pdf;
-        std::shared_ptr<InputSource> is;
-        bool must_be_complete;
-        std::shared_ptr<std::variant<std::string, JSON_Descr>> descr;
-        bool errors;
-        bool parse_error;
-        bool saw_qpdf;
-        bool saw_qpdf_meta;
-        bool saw_objects;
-        bool saw_json_version;
-        bool saw_pdf_version;
-        bool saw_trailer;
-        state_e state;
-        state_e next_state;
-        std::string cur_object;
-        bool saw_value;
-        bool saw_stream;
-        bool saw_dict;
-        bool saw_data;
-        bool saw_datafile;
-        bool this_stream_needs_data;
-        std::vector<state_e> state_stack;
-        std::vector<QPDFObjectHandle> object_stack;
-        std::set<QPDFObjGen> reserved;
-    };
+    class JSONReactor;
 
     void parse(char const* password);
     void inParse(bool);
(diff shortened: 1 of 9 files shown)


qpdf/qtest/qpdf/test-89.out
```diff
@@ -1,5 +1,5 @@
-WARNING: manual-qpdf-json.json, trailer at offset 1801: operation for array attempted on object of type dictionary: ignoring attempt to append item
-WARNING: manual-qpdf-json.json, obj:1 0 R at offset 1119: operation for array attempted on object of type dictionary: ignoring attempt to append item
-WARNING: manual-qpdf-json.json, obj:5 0 R at offset 1444: operation for dictionary attempted on object of type array: ignoring key replacement request
-WARNING: manual-qpdf-json.json, obj:5 0 R at offset 1456: operation for dictionary attempted on object of type name: ignoring key replacement request
+WARNING: manual-qpdf-json.json, object 0 0 at offset 1801: operation for array attempted on object of type dictionary: ignoring attempt to append item
```
jberkenbilt, line 1: oops, trailer turned into `object 0 0`

qpdf/qtest/qpdf/test-89.out
```diff
@@ -1,5 +1,5 @@
-WARNING: manual-qpdf-json.json, trailer at offset 1801: operation for array attempted on object of type dictionary: ignoring attempt to append item
-WARNING: manual-qpdf-json.json, obj:1 0 R at offset 1119: operation for array attempted on object of type dictionary: ignoring attempt to append item
```
jberkenbilt, line 2: `obj:1 0 R` is the name of the key in the json file. `object 1 0` is what it represents, but I'd rather see the actual key in the error message.

jberkenbilt (changes requested):

qpdf/qtest/qpdf/test-89.out
m-holger: That is the "slightly less information in error messages" I mentioned.

The trade-offs are:

- Construct individual object descriptions and add 15% to the runtime (i.e. forgo the 13% potential saving on the status quo) for runs that use a JSON input file.
- Store the key with the object and use a template string to construct the warning. Adding an extra string to QPDFValue adds somewhere between 0.5% to 1% to the runtime of QPDF, whether it uses JSON input or not.
- Drop the key and rely on the offset (and maybe theObjGen which we get for free), the proposed solution.

In addition to the 3 simple options above there are 2 further options that don't add any real overhead unless warnings are issued:

- Instead of storing a shared pointer to a template string with QPDFValues, have a shared pointer to a struct containing (at least) the input source and the template string. Calculate the key from the input source and offset only when needed.
- Provide a cmdline option to provide additional info given the JSON input file and offset.

To me the last 3 options are the most attractive, and the proposed option is probably sufficient. This is based on:
- I prefer not to have to pay unless I use it.
- Given that JSON is a pure text file, the offset is probably sufficient to locate the problem . (Most users who receive warnings probably won't be interested)
- Offset and key are probably insufficient info for those users who are interested. Additional info that what be useful:
   - full path of the actual object that triggered the warning
   - line no  and col corresponding to the offset, and possibly:
   - offset, line no and col for each element of the path, and
   - the actual line, surrounding lines, and marker highlighting the point of problem.

m-holger: Its probably useful to be clear about the audience for the warnings - are they for the benefit of end users or of developers tracking down programming errors. The two groups have different needs. Arguably QPDF provides too much info for the first group and two little for the second.

qpdf/qtest/qpdf/test-89.out
m-holger: See comments below. It would be easy to surpress ```object 0 0``` for ```obj == 0```.

qpdf/qtest/qpdf/test-89.out
jberkenbilt: They are for end users who are trying to create or use qpdf json files. I don't think they contain too much information. I don't want to sacrifice error messages for performance. I looked at it a little too fast and was thinking these were json errors, not qpdf errors. I might be okay with `obj:1 0 R` becoming `object 1 0` if there were a really compelling advantage, but I'm not okay with `trailer` becoming `object 0` or `object 0 0`. `trailer` has to be `trailer`. `object 0` looks like a mistake. But I'd rather have it be `obj:1 0 R` because it will be more obvious to someone who created a json with errors in it that they are in the right place, and I don't think a 13% improvement in json parsing is going to move the needle for most qpdf users.

Adding a command line to make warnings more specific is out. The command-line is already too complicated, and this won't help when other tools consume qpdf json.

Adding context for errors is too much. A byte offset is fine. qpdf is not a compiler.

I don't think we can sacrifice usability in the name of performance. We're talking about a linear improvement on code path that is not run repeatedly. Basically you're creating these object descriptions when the file is parsed. We could clearly gain a lot of performance by dropping object descriptions entirely, but it makes it much harder to work with error messages.

Basically I don't see the performance of qpdf json as a problem that requires other sacrifices to solve. Things that can be improved through simple refactoring, like checking token type codes instead of using dynamic cast or other optimizations to the parser are great -- they have an upside with no real downside. Squeezing out extra performance at the cost of user confusion when interpreting error messages feels like a wrong trade to me.

jberkenbilt:
I'm just ignoring fuzz errors for now. I guess it takes time for their infrastructure to pick up the new docker image across all their runners or something.

qpdf/qtest/qpdf/test-89.out
```diff
@@ -1,5 +1,5 @@
 WARNING: manual-qpdf-json.json, trailer at offset 1801: operation for array attempted on object of type dictionary: ignoring attempt to append item
 WARNING: manual-qpdf-json.json, obj:1 0 R at offset 1119: operation for array attempted on object of type dictionary: ignoring attempt to append item
 WARNING: manual-qpdf-json.json, obj:5 0 R at offset 1444: operation for dictionary attempted on object of type array: ignoring key replacement request
-WARNING: manual-qpdf-json.json, obj:5 0 R at offset 1456: operation for dictionary attempted on object of type name: ignoring key replacement request
+WARNING: manual-qpdf-json.json, obj:0 0 R at offset 1456: operation for dictionary attempted on object of type name: ignoring key replacement request
```
jberkenbilt, line 4: How did we lose the object number here?

qpdf/qtest/qpdf/test-90.out
```diff
@@ -1,5 +1,5 @@
 WARNING: various-updates.json, trailer at offset 606: operation for array attempted on object of type dictionary: ignoring attempt to append item
 WARNING: various-updates.json, obj:7 0 R at offset 197: operation for array attempted on object of type dictionary: ignoring attempt to append item
-WARNING: various-updates.json, obj:7 0 R at offset 309: operation for integer attempted on object of type array: returning 0
+WARNING: various-updates.json, obj:0 0 R at offset 309: operation for integer attempted on object of type array: returning 0
```
jberkenbilt, line 3: and here?

jberkenbilt (changes requested):

qpdf/qtest/qpdf/test-89.out
m-holger: Whether the error message says ```obj:1 0 R``` or ```object 1 0``` is not an issue - as long as we can pick up the obj gen from the QPDFValue rather than from the json file - which is consistent with what we are doing for objects read from a pdf file.

Having a separate template for trailer is not an issue  either - I got myself confused with messages containing a json path and thought `trailer` was just one of many possible values, which would be a problem.

I've pushed up a second attempt. If that one is acceptable, I would suggest we drop the ```obj:0 0 R``` for direct objects and simply go for "object at offset ...".

qpdf/qtest/qpdf/test-89.out
m-holger: See my comment above from just now. The object at at 1456 is direct.

qpdf/qtest/qpdf/test-90.out
m-holger: The current commits reproduce the existing output.

m-holger:
> Adding context for errors is too much. A byte offset is fine.

I am not convinced. One of the main problems that I see that if I remember correctly JSON is meant to make it easier for the qpdflib to be called from other languages. In this scenario, the JSON file is likely to only exist in memory during program execution and therefore the offset is of little use for identifying the object that caused the warning. So all you are left with is the obj gen of the (? great great grand parent of the ?) object.

On a less serious note, as far as I know it is much easier to find a (line, col) position than a offset position in most editors.


> I don't think we can sacrifice usability in the name of performance.

I fully agree. At the same time, if usability comes at a significant cost it makes sense to me to look at options for providing the same level of usability at a lower cost.

jberkenbilt (approved):
I like the new initializers and have been updating some of my older code to use them too. Thanks.

PR #901, opened 2023-02-16
Refactor removal of reserved objects in QPDF::JSONReactor

m-holger:
At the end of importing a JSON file, scan the object cache for reserved objects and replace with null objects.

When we are updating an existing QPDF object from JSON, do we need to preserve any pre-existing reserved objects?

### DIFF
--- a/include/qpdf/JSON.hh
+++ b/include/qpdf/JSON.hh
@@ -425,7 +425,7 @@ class JSON
         std::function<void(Pipeline*)> fn;
     };
 
-    JSON(std::shared_ptr<JSON_value>);
+    JSON(std::unique_ptr<JSON_value>);
 
     static bool checkSchemaInternal(
         JSON_value* this_v,
@@ -443,13 +443,13 @@ class JSON
         ~Members() = default;
 
       private:
-        Members(std::shared_ptr<JSON_value>);
+        Members(std::unique_ptr<JSON_value>);
         Members(Members const&) = delete;
 
-        std::shared_ptr<JSON_value> value;
+        std::unique_ptr<JSON_value> value;
         // start and end are only populated for objects created by parse
-        qpdf_offset_t start;
-        qpdf_offset_t end;
+        qpdf_offset_t start{0};
+        qpdf_offset_t end{0};
     };
 
     std::shared_ptr<Members> m;
--- a/include/qpdf/JSON.hh
+++ b/include/qpdf/JSON.hh
@@ -54,6 +54,8 @@ class JSON
 {
   public:
     static int constexpr LATEST = 2;
+
+    QPDF_DLL
     JSON() = default;
 
     QPDF_DLL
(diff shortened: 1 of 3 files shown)


m-holger:
I don't think the CI failure is caused by my code.
All errors are similar to:

command: dct_compress rawdata a.jpg 400 256 gray

libc++abi: terminating with uncaught exception of type std::runtime_error: Wrong JPEG library version: library is 90, caller expects 80

No idea how to fix this

jberkenbilt:
The mac build failure is probably related to conflicting libjpeg versions on the github runner. It's hard for people to get ABI transitions right. This will probably get fixed, but if it hasn't disappeared by whenever I have time to look at it, I'll either work around it in the qpdf build, report an issue/offer a PR to github, or both. I imagine this is probably breaking various builds across the ecosystem....

jberkenbilt:
I pushed a fix to the mac CI problem to main. Assuming my latest main build succeeded (I'm not staying up to wait for it), you should be able to rebase off current main. When macos builds were first supported in github, there were no jpeg library development files on the runner, so I built and installed it. This is no longer necessary, so I just removed my old workaround.

jberkenbilt:
I don't understand this change.

jberkenbilt:
It's hard for me to convince myself that this doesn't leave us open to weird ordering problems around the order in which objects are first referenced and first appear, though there is the "non-trivial replacement" coverage case which obviously still passes. With #902 and #896 both merged, there's only one commit left here, but I don't understand the purpose of the change, and it's not clear to me how the behavior is the same.

I don't think it's correct to clear all reserved objects. Suppose we implemented something in the future (or someone did so in their own code) where we could copy foreign objects from a combination of a PDF file and JSON file? Seems contrived, and maybe it's not a good example, but to me, replacing reserved objects that don't have anything to do with the JSON file feels like a bug. At the very least, it could mask a bug in which objects are left reserved.

So based on my analysis, this doesn't seem like a good change, but maybe I'm missing something, so feel free to clarify.

m-holger:
> I don't think it's correct to clear all reserved objects.

I added code to scan the object cache for pre-existing reserved objects at the start and leave them untouched when I replace reserved objects with null objects at the end. 

>  I don't understand the purpose of the change

Purely performance. Identifying reserved objects that we created (and didn't replace) by scanning the object cache (even twice) saves about 2% on the JSON input test compared to keeping track of reserved objects as we create and replace them.

> It's hard for me to convince myself that this doesn't leave us open to weird ordering problems around the order in which objects are first referenced and first appear

I don't think there is a problem. The timing of reserved object creation and destruction does not change, only the method by which we identify left-over reserved objects.

I rebased  this PR on top of #900. The combined effect is around 2.5% saving on the JSON input test, around 4.5% on the JSON output test and a small but consistent saving on the non-JSON tests.

jberkenbilt (approved):

PR #902, opened 2023-02-17
Refactor creation of object descriptions

m-holger:
This PR is based on the observation that only a very small proportion of the object descriptions created are actually used. It therefore makes sense to minimize the cost of creating and storing object descriptions at the expense of making it more expensive to access object descriptions.

This is achieved by storing the minimal amount of information required to create object descriptions with each object and assembling that information into the object description string at the point of retrieval.

This process was started in #844, which replaced object description strings with shared pointers to template strings, with individual information that was already available such as the obj gen and offset being filled in at the point of retrieval. This PR extends this by allowing the shared pointer not only to point at template strings but also at structs with additional information such as an object's parent. This makes it possible for example to create the parent's object description if and only if it is required at the point a child object's object description is retrieved.

### DIFF
--- a/include/qpdf/QPDF.hh
+++ b/include/qpdf/QPDF.hh
@@ -32,6 +32,7 @@
 #include <memory>
 #include <stdio.h>
 #include <string>
+#include <variant>
 #include <vector>
 
 #include <qpdf/Buffer.hh>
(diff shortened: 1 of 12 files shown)


jberkenbilt (approved):
I didn't have `variant` in my head. Looking at the docs, it seems to be new in C++17. I think I might have known that at one time. This is a change I would have never thought to make, but I think it's a good change.

PR #908, opened 2023-02-21
Fix typo in manual

m-holger:


m-holger:
re line 538:

> The following will produce unpredictable and probably incorrect
results because ``out.pdf`` won't have the same object numbers as

Would it make sense to flag that this is the counter example  by starting the sentence with something like:

"By contrast, the following will produce ..."?

jberkenbilt:
> "By contrast, the following will produce ..."?

Yes, I think that would be a good change. I'll hold merging in case you want to push this up.

m-holger:
Updated

jberkenbilt (approved):

PR #916, opened 2023-03-05
Split QPDFJob::run into createQPDF and writeQPDF

m-holger:
Allow users to transform the QPDF object created by a QPDFJob::run using custom code prior to writing.

This allows users to avoid having to potentially have to run one QPDFJob to perform initial transformation, re-openning the file to perform the custom transformations, writing the files and then running a second QPDFJob to perform final output transformations.


jberkenbilt:
I haven't looked at this yet, but on my roadmap for QPDFJob is to allow QPDFJob objects to be chained in some way where you could have a QPDF object be the input or the output. This may be a different way to do what I had in mind. I will look at it when I get a chance.

m-holger:
I was thinking about the same thing, and what I am proposing here is not a substitute for that. What I am proposing here is less powerful but, in my opinion, more convenient for many simple cases.

jberkenbilt (commented):
I think this is very good, and I intend to accept it. I'm not focused on qpdf this weekend, and I want to think through whether I agree with the API, etc., but I think there's a very strong chance I will take it. One additional thing that will be useful and that I want before release would be to add corresponding functions to the C API.

jberkenbilt:
> I was thinking about the same thing, and what I am proposing here is not a substitute for that. What I am proposing here is less powerful but, in my opinion, more convenient for many simple cases.

It may be good enough that I wouldn't ever bother with the chaining. I have my thoughts about chaining documented in TODO if you're curious.

m-holger:
> One additional thing that will be useful and that I want before release would be to add corresponding functions to the C API.

That was understood. I intended to wait until the C++ is nailed down.

Do you think we need a separate  C example?

m-holger:
@jbarlow83 , could you have a look at this proposal.

I was hoping that `auto job = QPDFJob::initializeFromJson; job.getQPDF()` could be easily wrapped in a Python procedure, allowing pikepdf to be used to customize qpdf jobs, i.e. allow users to write something lke

```python
# some code to setup QPDFJob config JSON
with  wrapped_get_QPDF(json) as pdf:
    #custom processing implemented in pikepdf
# output automatically written on exiting the with statement 
```
Does that look feasible/sensible?

jbarlow83:
Yes it does look this could be done.

I do recommend instead doing
```
std::unique_ptr<QPDF>
QPDFJob::createQPDF()

void
QPDFJob::writeQPDF(QPDF &qpdf)
```

because this allows the caller of the API to decide how to store the pointer. It's only necessary to return a shared_ptr when the callee still has a "stake" in the data, which I don't believe to be the case here. 

This is the opposite of what I was suggesting a few months ago - I learned some new things and changed my mind.

Practically speaking, it turns out pybind11 cannot handle situations where a C++ class (or any member in a class hierarchy) may be returned in more than one type of pointer holder, e.g. shared_ptr<C> and unique_ptr<C>. Currently QPDF is managed with unique_ptr.

include/qpdf/QPDFJob.hh
```diff
@@ -397,6 +397,21 @@ class QPDFJob
...
+    // object to be modified prior to writing it out. See
+    // examples/qpdfjob-remove-annotations for an illustration of its use.
+
+    // Run the first stage of the job. Return a nullptr if the configuration is
+    // not valid.
+    QPDF_DLL
+    std::unique_ptr<QPDF> createQPDF();
+
+    // Run the second stage of the job. Do nothing if a nullptr is passed as
+    // parameter.
```
m-holger, line 411: The "Do nothing if a nullptr is passed ..." needs removing

m-holger:
@jbarlow83 , I agree with your points.

@jberkenbilt , I have pushed up the suggested changes. I had to change QPDFJob to use unique pointers rather than shared pointers but I don't think there is a downside to the change.

jberkenbilt:
> Do you think we need a separate C example?

I'd have to wait until I saw the C API, but I doubt it.

jberkenbilt (approved):

PR #920, opened 2023-03-08
Refactor QdfFixer::processLines

m-holger:
I refactored QdfFixer as an experiment to see how much impact string optimizations can have on sections of code that are heavy on string processing. My key headlines are:

- copying strings rather than using string_views more than doubles run times.
- using regex_search rather than string_view methods increases run times more than 5 fold.


jberkenbilt (approved):
fix-qdf used to be a perl script. When I made the switch to C++-11 and had std::regex, I did a pretty much straight-through port without attempting to improve it. I agree that this is not high priority, but it's a great example, and it's a useful improvement in any case. Thanks.

PR #922, opened 2023-03-11
Refactor iteration over object containers

m-holger:
Just flying a kite here. The following are still missing:

- doc comments
- libtests
- methods allOf, anyOf, noneOf
- dictionary methods
- ? C-API ?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@ cmake_minimum_required(VERSION 3.16)
 # the project line. When updating the version, check make_dist for all
 # the places it has to be updated.
 project(qpdf
-  VERSION 11.3.1
+  VERSION 11.4.0
   LANGUAGES C CXX)
 
 # Enable correct rpath handling for MacOSX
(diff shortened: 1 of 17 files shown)


jberkenbilt:
meh, this is okay, but I generally prefer for loops to higher order functions. It's a debate we have at work...the engineers who are more front-end oriented and use frameworks like Angular really like functional programming, and those of us who are more low-level, prefer explicit loops and conditionals. I'm definitely not against functional programming, especially when there can be heavy optimization under the covers, and things like map functions are pretty handy, but I don't know that forEach offers any real value. It makes a more functional style approach available to people who prefer that style, but it also adds even more stuff to qpdf's already very crowded API.

If you really felt like this would help, you could probably convince me to take it, but it's not something I feel like qpdf really needs. Just my $0.02.

jberkenbilt:
That said, if (based on the comment you made to me by email), an approach like this would make possible optimizations without having to expose more internals, I would be in favor.

In other words, if this is a vehicle for clean optimization, I have no objection to this. If it's just about making a more functional style available, then I'm luke warm but would be willing to accept if you think it would be an improvement.

m-holger:
I am looking at this purely from a point of how we can provide the fastest way to loop over qpdf containers.

I keep looking at aitems and forEach to see which one is going to be cleaner and faster.

m-holger:
I have now found a version of `aitems` that outperforms `getArrayAsVector` without exposing `QPDF_Array`. This removes the reason for investigating the `forEach` approach.

Issue #923, opened 2023-03-12
Add support for --optimize-images to optimize images nested in groups


cdosborn:
I'd like to fix this issue, but could benefit from a "Use this approach.." as I'm new to c++ and the code base.

1. optimize is not optimizing all images, it doesn't recurse into the document https://github.com/qpdf/qpdf/issues/909#issuecomment-1461198532
2. It's easy to make it recurse, but that breaks the replacement logic.

If you just make it recurse, it finds the images, but then always inserts them at the top level of the page, rather than in the resources dict where they were found. 

Here is a [test pdf](https://github.com/qpdf/qpdf/files/10952121/test.pdf)
 that when passed through `--optimize-images` does not compress the sole image.

I could use some help with a suggested approach to change this. I'm also okay if someone else wants to fix this.

jberkenbilt:
Looking more closely, I don't think there's anything to fix. The comments in `QPDFPageObjectHelper` for `getImages` explicitly states that it doesn't recurse into form XObjects and to use `forEachImage` instead. I'm going to close this for now and make comments in #909

jberkenbilt:
I read the original message too fast. I see you were using pikepdf for illustrative purposes.

jberkenbilt:
The key will be to use `forEachImage` instead of `getImages` in `QPDFJob::handleTransformations`. I think it will not be a complicated fix, but the code is pretty involved and easy to break, and we have to be careful about compatibility. If you'd like to take a crack at it, I'm happy to review. More likely than not, it would take me about the same amount of time to implement a fix as it would to figure out the fix enough to guide someone else through it. If you're really interested in contributing to qpdf and have the time to play around with it, please feel free. Otherwise, I can commit to fixing this for whatever the next release is as I believe it will not take me very long to fix it.

jberkenbilt:
Looking at change history, it seems that, when I added `forEachImages`, I just forgot to update `--optimize-images`. What I will need to do is fix that and also check to see whether there are other `getImages` calls within the library.

jberkenbilt:
Another possible fix would be to modify getImages to take an optional argument that tells it whether to recurse into form xobjects. I'd have to check more closely to see what the implications would be. Note that adding a literal optional argument would be ABI breaking so, for the time being, it would have to be an overloaded function with an ABI comment marker which I would resolve for qpdf 12.

PR #924, opened 2023-03-13
Improve --optimize-images to find images nested within XObjects

cdosborn:
See this issue: https://github.com/qpdf/qpdf/issues/923

@jberkenbilt Appreciate any feedback you have! How would you suggest me testing this?


jberkenbilt:
We'll have to update the test suite with a file that has this. You can search in qpdf/qtest/*.test for where `--optimize-images` is tested and see about adding a new test file. The test file has to be freely redistributed, unlicensed, etc.

cdosborn:
Getting the following diff when I run my test locally:
```
-qpdf: image /X1 on page 1: optimizing image reduces size from 2628 to 1592
+qpdf: image /X1 on page 1: optimizing image reduces size from 2628 to ...
 qpdf: wrote file a.pdf
```
Trying to figure out why this number is being truncated via ellipsis. Seems to be an issue with logging?

cdosborn:
> Getting the following diff when I run my test locally:
> 
> ```
> -qpdf: image /X1 on page 1: optimizing image reduces size from 2628 to 1592
> +qpdf: image /X1 on page 1: optimizing image reduces size from 2628 to ...
>  qpdf: wrote file a.pdf
> ```
> 
> Trying to figure out why this number is being truncated via ellipsis. Seems to be an issue with logging?

Figured it out.

libqpdf/QPDFJob.cc
```diff
@@ -2350,30 +2350,34 @@ QPDFJob::handleTransformations(QPDF& pdf)
...
+                        image);
+                    std::shared_ptr<QPDFObjectHandle::StreamDataProvider> sdp(
+                        io);
+                    if (io->evaluate(
+                            "image " + name + " on page " +
+                            std::to_string(pageno))) {
+                        QPDFObjectHandle new_image = pdf.newStream();
+                        new_image.replaceDict(image.getDict().shallowCopy());
+                        new_image.replaceStreamData(
+                            sdp,
```
m-holger, line 2375: ... and
```cpp
                                std::move(io),
```
it here.

libqpdf/QPDFJob.cc
```diff
@@ -2350,30 +2350,34 @@ QPDFJob::handleTransformations(QPDF& pdf)
...
+                    std::string name = key;
+                    QPDFObjectHandle& image = obj;
+                    ImageOptimizer* io = new ImageOptimizer(
+                        *this,
+                        m->oi_min_width,
+                        m->oi_min_height,
+                        m->oi_min_area,
+                        image);
+                    std::shared_ptr<QPDFObjectHandle::StreamDataProvider> sdp(
+                        io);
```
m-holger, line 2368: @jberkenbilt  It probably would be cleaner to create `io` as a unique_ptr 
```cpp
                    auto io = std::make_unique<ImageOptimizer>(
                        *this,
                        m->oi_min_width,
                        m->oi_min_height,
                        m->oi_min_area,
                        image);
```
and then ...

libqpdf/QPDFJob.cc
cdosborn: Feel free to push to my branch. I'll need to spend some time to understand this.

libqpdf/QPDFJob.cc
```diff
@@ -2350,30 +2350,33 @@ QPDFJob::handleTransformations(QPDF& pdf)
...
-                        .replaceKey(name, new_image);
-                }
-            }
+            ph.forEachImage(
+                true,
+                [this, pageno, &pdf](
+                    QPDFObjectHandle& obj,
+                    QPDFObjectHandle& xobj_dict,
+                    std::string const& key) {
+                    QPDFObjectHandle& image = obj;
```
m-holger, line 2359: Sorry, also no need to create a new to create a new ref to obj.

(As an aside, `(auto& image, auto& xobj_dict, auto const& key)` also works.)

libqpdf/QPDFJob.cc
```diff
@@ -2350,30 +2350,33 @@ QPDFJob::handleTransformations(QPDF& pdf)
         for (auto& ph: dh.getAllPages()) {
             ++pageno;
             QPDFObjectHandle page = ph.getObjectHandle();
```
m-holger, line 2352: I think `page` is no longer used

libqpdf/QPDFJob.cc
```diff
@@ -2350,30 +2350,34 @@ QPDFJob::handleTransformations(QPDF& pdf)
...
-                        .replaceKey(name, new_image);
-                }
-            }
+            QPDFPageObjectHelper(page).forEachImage(
+                true,
+                [&, pageno, &pdf = pdf](
+                    QPDFObjectHandle& obj,
+                    QPDFObjectHandle& xobj_dict,
+                    std::string const& key) mutable {
+                    std::string name = key;
```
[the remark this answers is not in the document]
cdosborn: :sweat_smile: Didn't realize that this constructed a new object.

cdosborn:
@m-holger Thanks for feedback. Let me know if there is anything else I can improve.

jberkenbilt (approved):
Thanks @cdosborn for the contribution and @m-holger for the review. All I have to do is click the button! (But I also reviewed -- looks great.)

PR #929, opened 2023-03-19
Add new convenience class QPDFObjGen::Guard

m-holger:
Support the common pattern of using sets of QPDFObjGen s to detect loops.  

To reduce boiler plate and for a slight performance improvement.

### DIFF
--- a/include/qpdf/QPDFOutlineDocumentHelper.hh
+++ b/include/qpdf/QPDFOutlineDocumentHelper.hh
@@ -22,13 +22,13 @@
 #ifndef QPDFOUTLINEDOCUMENTHELPER_HH
 #define QPDFOUTLINEDOCUMENTHELPER_HH
 
+#include <qpdf/QPDF.hh>
 #include <qpdf/QPDFDocumentHelper.hh>
 #include <qpdf/QPDFNameTreeObjectHelper.hh>
+#include <qpdf/QPDFObjGen.hh>
 #include <qpdf/QPDFOutlineObjectHelper.hh>
 
-#include <qpdf/QPDF.hh>
 #include <map>
-#include <set>
 #include <vector>
 
 #include <qpdf/DLL.h>
@@ -69,16 +69,16 @@ class QPDFOutlineDocumentHelper: public QPDFDocumentHelper
     {
         friend class QPDFOutlineObjectHelper;
 
+        // ABI: remove  QPDF_DLL and pass og by value.
         QPDF_DLL
         static bool
         checkSeen(QPDFOutlineDocumentHelper& dh, QPDFObjGen const& og)
         {
-            return dh.checkSeen(og);
+            return !dh.m->seen.add(og);
         }
     };
 
   private:
-    bool checkSeen(QPDFObjGen const& og);
     void initializeByPage();
 
     class Members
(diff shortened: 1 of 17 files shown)


include/qpdf/QPDFOutlineDocumentHelper.hh
```diff
@@ -69,16 +69,16 @@ class QPDFOutlineDocumentHelper: public QPDFDocumentHelper
     {
         friend class QPDFOutlineObjectHelper;
 
+        // ABI: remove  QPDF_DLL and pass og by value.
```
jberkenbilt, line 72: yep, thanks. This should never have been exported.

jberkenbilt (changes requested):
This is fine but I don't understand your choice of names of the local variables. If the set was called `seen`, why is the guard called `unseen`? When I see `unseen.enter(...)`, it makes me feel like it's saying that something is now unseen. I would prefer not to see inversion of the names of the guard variables. Whatever the old set was called, the guard should be called the same thing. In some cases, a name change may be in order. For example, if you only enter and never exit, `visited` is probably fine, but if you enter and exit, `visiting` would probably be better. Maybe you can explain why you chose the names you did, inverting the grammatical sense of my original choices in some but not all cases.

It would also be useful to describe in a comment how guard is intended to be used since it is in the public API.

I'm not sure I like the name `Guard`. To me, a guard is typically something whose existence prevents you from doing something and whose destruction allows it again, kind of like the lock of a mutex. Also, I would prefer `leave` to `exit`. I'd be concerned about `exit` being a macro on some platform. As a rule, I try to avoid using names that are very common and old C library functions. This has gotten me in trouble before. I have had to break API for this before as with strcasecmp.

Would you be comfortable with calling this `Tracker` instead of `Guard`? Or can you suggest a different name?

These comments are all about names. The functionality is fine.

m-holger:
I had difficulties thinking of a good name. I think I like `Tracker`.

>  If the set was called seen, why is the guard called unseen? 

I was thinking of a guard as a gatekeeper, in this case a gatekeeper that only allow unseen objects to enter. If we think of it as a tracker, `Tracker seen` makes sense to me.

With tracker, it would feel more natural to me to have `seen.remove(og)` rather than `seen.exit(og)`. 

I am also not sure whether `if (seen.enter(og)) ...` clearly communicates that this is true the first time we encounter `og`. Would `seen.addNew(og)` or `seen.firstTime(og)` be clearer?

jberkenbilt:
How about just `add` and `remove`?

m-holger:
I have reworked QPDFObjGen::Guard / Tracker to inherit from std::set<QPDFObjGen> because I found that occasionally I needed additional set methods.

In view that the class now supports all std::set methods I thought that QPDFObjGen::set would be a better name. I kept the method name 'add' since the method is significantly different from the standard insert, but replaced 'remove' with the standard 'erase'.

libqpdf/QPDF.cc
```diff
@@ -2193,70 +2194,62 @@ QPDF::reserveObjects(QPDFObjectHandle foreign, ObjCopier& obj_copier, bool top)
 
     if (foreign.isIndirect()) {
         QPDFObjGen foreign_og(foreign.getObjGen());
-        if (obj_copier.visiting.find(foreign_og) != obj_copier.visiting.end()) {
-            QTC::TC("qpdf", "QPDF loop reserving objects");
+        if (obj_copier.object_map.count(foreign_og) > 0) {
+            QTC::TC("qpdf", "QPDF already reserved object");
+            if (obj_copier.visiting.count(foreign_og)) {
+                QTC::TC("qpdf", "QPDF loop reserving objects");
+            }
```
m-holger, line 2201: It appears to me that obj_copier.visiting is unnecessary, since the actual loop avoidance can be achieved by checking the object_map.

jberkenbilt (approved):
I like this solution better.

PR #931, opened 2023-03-20
In QPDF refactor trailer access

m-holger:
For performance.

### DIFF
--- a/include/qpdf/QPDF.hh
+++ b/include/qpdf/QPDF.hh
@@ -1126,7 +1126,8 @@ class QPDF
 
     void parse(char const* password);
     void inParse(bool);
-    void setTrailer(QPDFObjectHandle obj);
+    void setTrailer(QPDFObjectHandle oh);
+    void updateTrailer(QPDFObjectHandle oh);
     void read_xref(qpdf_offset_t offset);
     bool resolveXRefTable();
     void reconstruct_xref(QPDFExc& e);
@@ -1718,7 +1719,8 @@ class QPDF
         std::set<int> deleted_objects;
         std::map<QPDFObjGen, ObjCache> obj_cache;
         std::set<QPDFObjGen> resolving;
-        QPDFObjectHandle trailer;
+        QPDFObjectHandle trailer_oh{};
+        QPDF_Dictionary* trailer{nullptr};
         std::vector<QPDFObjectHandle> all_pages;
         std::map<QPDFObjGen, int> pageobj_to_pages_pos;
         bool pushed_inherited_attributes_to_pages{false};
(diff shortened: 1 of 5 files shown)


jberkenbilt:
How much difference does this make? I'd prefer not to start leaking direct access to the underlying QPDF_* objects too much into the internals. It might be worth it for the trailer since it's the starting point for so many operations, but it makes me slightly uncomfortable since it feels like it bypasses certain functionality. What happens if someone replaces the trailer? (Do I even allow that?) Anyway if it's significant, I guess we can do it, but I don't always choose performance over safety/clarity, and I think the underlying objects should hide behind QPDFObjectHandle in most cases. Feel free to try to convince me otherwise.

m-holger:
>  What happens if someone replaces the trailer? (Do I even allow that?) 

Setting / replacing the trailer is private to QPDF.

> How much difference does this make?

At the time of writing I saw a 1% improvement. Against the current main this has gone pretty much away and there is no clear-cut advantage, which suggests shelving this at least for now.

> I'd prefer not to start leaking direct access to the underlying QPDF_* objects too much into the internals. ... I think the underlying objects should hide behind QPDFObjectHandle in most cases. 

My thinking is moving in the opposite direction, for two opposing reasons:

- QPDFObjectHandle is too permissive

  Object handles permit any modifications to any objects, without enforcing the pdf standard. Allowing core pdf objects like the trailer or the the root objects to be modified in illegal ways does not seem desirable. I feel there is a strong case for creating dedicated objects that check and repair any data read in once, and then maintain them in a legal state. This allows the user to make strong assumptions about the values accessed through those objects, and avoids the need for qpdf to re-validate the same data repeatedly.

- QPDFObjectHandle is too restrictive

   QPDFObjectHandle make strong assumptions about the handles calling certain methods and validate them on every call. When making multiple method calls for the same object  this can have significant overhead, and for core modules like QPDF, QPDFWriter and QPDFParser there is a strong case for by-passing QPDFObjectHandle in places. That said, there is benefit in thinking about a clean internal API to allow those cases to be handled safely and cleanly.

jberkenbilt:
Hmm. I guess I sort of thought we'd wait until your QPDFObjectHandle replacement to address the issues you mentioned. Let's hold off on this one for now. I think when you get your QPDFObjectHandle replacement ready, if that's still in your plans, we can do a more comprehensive sweep through the internals. Unless all the intervening work has shifted your views about the best way to get a way of accessing objects that has a better balance between protective and performant. I guess I see exposing trailer as a QPDF_Dictionary as an abstraction leak made as a trade for the fact that we don't have quite the right QPDFObjectHandle abstraction yet.

m-holger:
Closed as this is out of date

PR #933, opened 2023-03-21
C-API : expose QPDFJob::createQPDF and writeQPDF

m-holger:


qpdf/qpdfjob-ctest.c
```diff
@@ -97,6 +97,30 @@ run_tests()
     assert(qpdfjob_run(j) == 2);
     qpdfjob_cleanup(&j);
     printf("json error test passed\n");
+
+    /* qpdfjob_create_qpdf and qpdfjob_write_qpdf test */
+    argv[0] = "qpdfjob";
+    argv[1] = "minimal.pdf";
+    argv[2] = "d.pdf";
```
jberkenbilt, line 104: Should this be m.pdf

qpdf/qtest/qpdf/qpdfjob-ctest.out
```diff
@@ -9,3 +9,8 @@ json warn test passed
 |custom|qpdfjob json|custom|: |custom|an output file name is required; use - for standard output|custom|
 |custom|qpdfjob json|custom|: |custom|an output file name is required; use - for standard output|custom|
 json error test passed
+qpdfjob: d.pdf: write progress: 0%
+....other write progress....
+qpdfjob: d.pdf: write progress: 100%
+qpdfjob: open m.pdf: No such file or directory
```
jberkenbilt, line 15: Is this your intention?

jberkenbilt (changes requested):
The test outputs don't look hand-generated. Was it your actual intention to write to one file and then test that a different file didn't exist?

qpdf/qtest/qpdf/qpdfjob-ctest.out
m-holger: This is intentional.

This was entirely clear to me when I wrote it, but with the wisdom that comes from living another 2 weeks I can see that a comment would have been a good idea. Now added.

qpdf/qpdfjob-ctest.c
m-holger: d.pdf is correct.

qpdf/qtest/qpdf/qpdfjob-ctest.out
jberkenbilt: From the wisdom of one more day, I see that what concerned me was the lack of anything that actually looks at d.pdf to make sure it's correct. But, in fact, I actually see that check. So never mind.

jberkenbilt (approved):

qpdf/qtest/qpdf/qpdfjob-ctest.out
jberkenbilt: > This was entirely clear to me when I wrote it, but with the wisdom that comes from living another 2 weeks I can see that a comment would have been a good idea. Now added.

I'm going to merge as is with the comment, which I think adequately resolves the issue. As a habit, I typically call files like this "does-not-exist.pdf" or something similar.

m-holger: > As a habit, I typically call files like this "does-not-exist.pdf" 

That sounds like an obvious solution that should have occurred to me without prompting.

PR #943, opened 2023-04-04
Forward declare JSON_array for C++20 compatibility.

zclifford:
In C++20 the destructor for vector<T> is constexpr. This requires that T not be an incomplete type where the destructor is called.

JSON_array's `vector<JSON> elements` field violates this because JSON is still an incomplete type when JSON_array's destructor is defined.

To fix this forward declare JSON_array and only define it afer JSON's definition ends (once it is a complete type).

No new tests since I'm not really sure how to test it and qpdf doesn't officially support C++20 yet. If there's a regression I'll notice it on my end, but I'll only check once per release until my company is a bit farther along in our C++20 migration.

### DIFF
--- a/include/qpdf/JSON.hh
+++ b/include/qpdf/JSON.hh
@@ -374,16 +374,7 @@ class JSON
         std::map<std::string, JSON> members;
         std::set<std::string> parsed_keys;
     };
-    struct JSON_array: public JSON_value
-    {
-        JSON_array() :
-            JSON_value(vt_array)
-        {
-        }
-        virtual ~JSON_array() = default;
-        virtual void write(Pipeline*, size_t depth) const;
-        std::vector<JSON> elements;
-    };
+    struct JSON_array;
     struct JSON_string: public JSON_value
     {
         JSON_string(std::string const& utf8);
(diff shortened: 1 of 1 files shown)


jberkenbilt:
I doubt we'll probably make too many changes that break C++20. I could potentially add another CI job that runs with  C++20 to make sure we stay compatible with it. That's probably worth doing and would be an easy change for me to make. I already have a bunch of "run this in this special just to be sure" things in CI, like running with unsigned char.

zclifford:
That all sounds good to me. Per my bug I'm not convinced compilers shipping today with systems will hit all C++20 issues yet or not, but if not that'll probably work itself out with time. 

Also just checking / no hurry, but is there anything I should do at the moment? (I'm new to the github PR workflow). I'm on vacation tomorrow so I won't be able to respond to any action items then.

jberkenbilt:
@zclifford I usually catch up on my qpdf work on most weekends so hopefully I can take a deeper look then, but at a glance, what you've done looks great and I will probably be able to accept it as is. Thanks for taking the extra effort to provide a pull request.

jberkenbilt:
I'm accepting this change but it has merge conflicts, so I have reproduced it in a separate branch and verified that it builds with C++-20. I apologize for delay. If I wired it up right, this PR will close when I merge that change.

jberkenbilt:
This is closed in favor of #943 which is the same change. Thanks for the contribution.

PR #947, opened 2023-04-11
Prevent destruction of shared null objects

m-holger:
#863 uses a single null object for nulls that were previously implicit. In certain circumstances this shared null object gets destroyed (i.e changed to a QPDF_Destroyed object) when a QPDF object is destroyed.

Modify the QPDF destructor so that null objects get disconnected from the dying QPDF object but not destroyed to prevent this from happening.

### DIFF
--- a/libqpdf/QPDF.cc
+++ b/libqpdf/QPDF.cc
@@ -247,10 +247,11 @@ QPDF::~QPDF()
     // but we'll explicitly clear the xref table anyway just to
     // prevent any possibility of resolve() succeeding.
     this->m->xref_table.clear();
-    auto null_obj = QPDF_Null::create();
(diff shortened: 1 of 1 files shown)


jberkenbilt:
I'm going to merge this, but it makes me wonder about something. We've talked in the past about eventually not allowing an object to have two parents. I don't think this matters as much anymore, and maybe it doesn't make sense at all with direct scalars, which are immutable. So.....perhaps the full fix should be to not destroy immutable scalars? It's unlikely but not impossible that someone could try to be clever and reuse the same direct object for some constant. Like someone could create a `/Page` name and add that to a bunch of different QPDF objects. So I think this is good, but I wonder whether it goes far enough.

PR #958, opened 2023-05-18
Added method qpdfjob_run_from_json_with_result for C# wrapper

Sicos1977:
is it possible for you to add this method to your repository so that I don't have to add it when I make a new C# wrapper for qpdf? I use this method to capture the output of cout and cerr so that I can return it to my C# wrapper.

### DIFF
--- a/include/qpdf/qpdfjob-c.h
+++ b/include/qpdf/qpdfjob-c.h
@@ -81,6 +81,13 @@ extern "C" {
     QPDF_DLL
     int qpdfjob_run_from_json(char const* json);
 
+    /* A method that outputs out and err as a string instead of writing it
+    * to the console so that it can be captured and used from another
+    * language like C#
+    */
+    QPDF_DLL
+    int qpdfjob_run_from_json_with_result(char const* json, char const** cout_result, char const** cerr_result);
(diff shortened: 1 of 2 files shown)


Sicos1977:
I managed to get everything to work with the new logger class so this pull reqeust became obsolete

jberkenbilt:
I'm really pleased to see that recent enhancements to the library provided solutions to the issues you were having.

Sicos1977:
It also makes this much easier this way. C# interop is always like some black magic to me how everything works. That is why I like the json jobs because it makes everything so much easier that all the interop handling you otherwise have to do when working with C++ or C object from C# .

I still have to iron somethings out but with some luck I have a new wrapper ready tomorrow that uses the default qpdf29.dll without any addons like the method I had in this pull request.

jberkenbilt:
I plan on releasing 11.4 soon -- maybe this weekend -- so keep me posted. I also got a report that the mingw DLLs are much bigger than they used to be, and I plan on investigating that. Which Windows build do you use? I made msvc the default a few releases ago.

Sicos1977:
I use the msvc build ...

PR #959, opened 2023-05-19
Add new method QPDF::newReserved()

m-holger:
Logically, creation of a reserved object is intimately linked to a QPDF object and therefore should be QPDF method.

### DIFF
--- a/include/qpdf/QPDF.hh
+++ b/include/qpdf/QPDF.hh
@@ -413,6 +413,24 @@ class QPDF
     QPDF_DLL
     QPDFObjectHandle newStream(std::string const& data);
 
+    // A reserved object is a special sentinel used for qpdf to
+    // reserve a spot for an object that is going to be added to the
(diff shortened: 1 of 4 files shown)


jberkenbilt (approved):
The next habit for you is to update ChangeLog and release notes. This will force me to codify when I update them. Basically:
* Release notes are things that users should care about. All API changes go in the release notes. The idea is that you should know all about what has changed so you can update your knowledge from whatever the last release you studied was until now.
* ChangeLog is slightly more detailed and may include information about things that users don't necessarily care about. I would include some internal API changes here as well as significant bug fixes, but it is not exhaustive.
* Version control history is the ultimate, exhaustive source of truth.

I use ChangeLog mainly to remind my self roughly when a change was made and what the first version was that had some change/fix/improvement. I try to keep both up to date as I go but typically review a diff of the public include directory as well as the git history before a release to make sure everything important is there.

PR #963, opened 2023-05-20
Code tidy

m-holger:
Courtesy of Clang-Tidy


jberkenbilt (approved):
This is great. Thanks. I was thinking of doing this after seeing all these in CLion.

jberkenbilt:
I will merge once this is through CI (minus 32-bit failures which I will address separately).

PR #967, opened 2023-05-21
Fix doc comment in QPDFObjGen::set

m-holger:


jberkenbilt (approved):
No need to wait for CI on this one.

PR #969, opened 2023-05-21
Code formating : drop 'this->' in 'this->m'

m-holger:
Split into two commits for results of 'search & replace' and manual tweaks.

There are a couple of places where I would tweak the formatting. but I think this will be taken care off by increasing the max code width.


jberkenbilt (approved):
Thanks -- will merge after CI and after errands

PR #970, opened 2023-05-21
Add qpdf 12 information to release notes

jberkenbilt:
@m-holger Can you take a look at this before I push 11.4 out?

### DIFF
--- a/manual/release-notes.rst
+++ b/manual/release-notes.rst
@@ -6,6 +6,23 @@ Release Notes
 For a detailed list of changes, please see the file
 :file:`ChangeLog` in the source distribution.
 
+Planned changes for future 12.x (subject to change):
+  - ``QPDFObjectHandle`` will support move construction/assignment.
+    This change will be invisible to most developers but may break
+    your code if you rely on specific behavior around how many
+    references to a QPDFObjectHandle's underlying object exist. You
+    would have to write code specifically to do that, so if you're not
+    sure, then you shouldn't have to worry.
+
+  - ``QPDFObjectHandle`` will be implicitly convertible to ``bool``
+    with undefined objects evaluating to ``false``. This can simplify
+    error handling and will facilitate use of ``QPDFObjectHandle``
+    with some newer standard library constructs. This change won't
+    affect any existing code unless you have written your own
+    conversion methods to/from ``QPDFObjectHandle``. In that case,
+    it's possible that the new qpdf-provided conversion may override
+    your conversion.
+
 .. x.y.z: not yet released
 
 11.4.0: not yet released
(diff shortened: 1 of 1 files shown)


manual/release-notes.rst
```diff
@@ -6,6 +6,30 @@ Release Notes
...
+    affect any existing code unless you have written your own
+    conversion methods to/from ``QPDFObjectHandle``. In that case,
+    it's possible that the new qpdf-provided conversion may override
+    your conversion.
+
+  - It has always been invalid to attempt to dereference uninitialized
+    ``QPDFObjectHandle`` objects, but if you stored one somewhere and
+    never accessed it, you might have gotten away with it. The next
+    major version of qpdf will detect if you try to store an
+    uninitialized ``QPDFObjectHandle`` in a dictionary or an array and
```
m-holger, line 30: In 11.4 insertion into a QPDF_Array throws a logic error as part of the foreign ownership check

jberkenbilt: Meaning that it's not safe now either? What should this say instead?

jberkenbilt: Or do we need to say this for the 11.4 release notes?

jberkenbilt: I see that was added between 11.3 and now.

jberkenbilt: blargh, I guess it just moved.

m-holger: Probably should be part of the release notes

jberkenbilt: I reread your message. I will push up a change.

m-holger: for 10.4. This was part of the rework of QPDF_Array. In most cases it would have thrown a logic error anyway, but there ways to sneak them in, eg by creating a newArray from a std::vector<QPDFObjectHandle>

jberkenbilt:
@m-holger Is this more accurate?

manual/release-notes.rst
```diff
@@ -26,6 +43,10 @@ For a detailed list of changes, please see the file
     - Add ``QPDF::newReserved`` as a better alternative to
       ``QPDFObjectHandle::newReserved``.
 
+    - If you add an uninitialized ``QPDFObjectHandle`` to an array,
+      qpdf will through a ``logic_error``. It has always been invalid
+      to do this, but before, it wouldn't have been caught until later.
```
m-holger, line 48: through -> throw

m-holger: otherwise fine

jberkenbilt: oops, thnaks.

PR #975, opened 2023-05-27
Code tidy - reflow comments and strings

m-holger:


jberkenbilt (approved):
Thanks. In the future, if you do something in an automated way, can you include what you did in the commit comment? This can either be a command or a description of something you did in the IDE or whatever. I'm particularly curious about reflowing comments and strings. When I run clang-format, it doesn't do that. What do you do to reflow comments? I didn't look at all of them closely, but it looks like it preserved my bulleted lists.

Something that might be helpful to have would be a list of commits that are mainly formatting so we can populate that git file that ignores for blame/annotate. We discussed it ages ago and I didn't think it was worth it, but I think it probably is worth it now. I can't remember what project it was, but there was some repo I was looking at recently whose CONTRIBUTING (or whatever it was) file had a list of commits that you should write to whatever that file is. Sorry I'm being so vague -- I'd have to go back to the docs to remember the exactly details, but hopefully I'm clear.

m-holger:
>  if you do something in an automated way, can you include what you did in the commit comment? 

I do. I am afraid  this is tedious manual reformatting. The only way I can get it to work is joining two or more comment lines until they are > 100 cols and running clang-format.

> Something that might be helpful to have would be a list of commits that are mainly formatting so we can populate that git file that ignores for blame/annotate.

I remember. I'll set up a file.

PR #978, opened 2023-06-02
Code tidy - reflow comments and strings

m-holger:
The first commit is mechanical but manual, i.e combine two comment lines / 2 strings and then let clang-tidy re-flow the code.

Should commits correcting doc typos be added to .git-blame-ignore-revs? My view would be yes.


jberkenbilt (approved):
I'm accepting this without careful review because
* I trust your care in handling comments and strings and in the test suite to catch any string mistakes
* This was a lot of work and, as you pointed out, there's a high risk of merge conflicts
Thanks. There's no way I would ever taken the time to do this.

m-holger:
The cost of a slight tendency towards OCD

PR #979, opened 2023-06-02
In FUTURE make various QPDFObjectHandle methods const

m-holger:
Making methods const where possible 

- makes code clearer,
- avoids unnecessary copying, and
- gives compilers greater scope for optimization.

In order to allow object handle methods to be made const it is necessary to refactor resolving  of objects. In particular, QPDF_Unresolved objects are made responsible for copying, unparsing, etc themselves. Where QPDFObject methods are accessing QPDFValue data members directly rather than through QPDFValue methods, the QPDFObject method is made responsible for handling unresolved objects correctly. At a basic level, this is achieved by modifying the QPDF::resolve method to return a pointer to the resolved object.

In addition to permitting QPDFObjectHandle methods to be made const, this approach to resolving object handles is also more efficient. Accessing unresolved object handles is (relatively) rare: it only applies to indirect objects and only to the first significant access to an object (or even, any object of an object stream). Confining the necessary checks to unresolved object
access only rather than to any object access reduces overhead.

### DIFF
--- a/include/qpdf/QPDF.hh
+++ b/include/qpdf/QPDF.hh
@@ -792,12 +792,13 @@ class QPDF
     class Resolver
     {
         friend class QPDFObject;
+        friend class QPDF_Unresolved;
 
       private:
-        static void
-        resolve(QPDF* qpdf, QPDFObjGen const& og)
+        static QPDFObject*
+        resolved(QPDF* qpdf, QPDFObjGen og)
         {
-            qpdf->resolve(og);
+            return qpdf->resolve(og);
         }
     };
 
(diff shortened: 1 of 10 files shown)


jberkenbilt:
I like the idea and the approach. Would you be willing to rework this without FUTURE now that we are considering the ABI window to be open? It would make the changes easier to review.

m-holger:
I have been thinking about qpdf 12 and would prefer not to break ABI at the moment. In fact, I only reworked the PR yesterday to change it back to FUTURE.

jberkenbilt:
Okay, how would you like me to review this? Are you just looking for my reaction, or do you want a review of specific changes? I had tried a few times in the past to make these const, but I didn't have a solution to the lazy resolve. After the object/value rework, the solution you have put forth seems to be a good one for that problem, and being able to use more const or const& QPDFObjectHandles would be good.

m-holger:
I just sent you a private email to give context to where my head is at the moment.

I see this PR as fundamental for further changes and like to get it it to a mergeable state.

jberkenbilt (approved):
I like this. I think this shows how the object/value split and other changes you've made are moving things in the right direction. This is actually fairly simple, whereas when I attempted a similar change several years ago, it was too hard to be worth doing.

PR #980, opened 2023-06-05
Avoid unnecessary copying of stream dictionary in QPDF::readObject

m-holger:
Unless I am missing something, there is no reason to differentiate between a missing /Length key and one with a null value.


jberkenbilt (approved):
I don't think you're missing anything. I'm not sure why I coded it that way.

PR #982, opened 2023-06-09
Apply various Clang-Tidy rules

m-holger:
The issues were identified Clang-Tidy, but where addressed individually.

The fixes are mainly manual, partly because Clang-Tidy cannot  fix the issues automatically but also because some fixes do not apply to qpdf and therefore require suppression of warnings rather than fixes.


jberkenbilt:
I'm looking forward to studying this. I am going to learn things. I am not aware of emplace, for example.

m-holger:
I find the modernize-use-emplace rule quite interesting. Theoretically using emplace should improve performance as it avoids unnecessary copying/moving. The reality is that for both clang-15  and gcc-12 it doesn't make a difference performance wise. At some level this makes sense: if clang-tidy can detect that emplace should be used, why would clang not optimize the unnecessary copy away? I guess this explains why the rule is called "modernize-use-emplace" rather than "performance-use-emplace".

I still feel that it is useful to apply the modernize-use-emplace rule for three reason:
- it expresses in code more clearly the cost of the operation
- it is likely to improve code quality metrics for qpdf
- it may well have a significant impact for users of other/older compilers

examples/pdf-count-strings.cc
```diff
@@ -36,7 +35,7 @@ class StringCounter: public QPDFObjectHandle::TokenFilter
     int getCount() const;
 
   private:
-    int count;
+    int count{0};
```
jberkenbilt, line 36: This is my favorite "recent" (in my C++ coding life) improvement to C++.

jberkenbilt (approved):
This is great. Before I merge it, did you happen to run the ABI test on this? It's not part of CI, but I'd like extra assurance that this doesn't break ABI. I'd run the whole ./abi-perf-test on it.

m-holger:
just I ran ./abi-perf-test at various stages, and reran it now to make sure.

PR #983, opened 2023-06-10
 Add new Buffer method copy and deprecate copy constructor / assignment operator

m-holger:
Add to change log:

Add new method Buffer::copy and deprecate Buffer copy constructor and assignment operator. Buffer copy operations are expensive as they always involve copying the buffer content. Use "buffer2 = buffer1.copy();" or "Buffer buffer2{buffer1.copy()};" to make it explicit that copying is intended.

Add to release notes:

Add new method Buffer::copy and deprecate Buffer copy constructor and assignment operator. 

Add to release notes - planned changes for 12.x:

Buffer copy constructor and assignment operator will be removed. Buffer copy operations are expensive as they always involve copying the buffer content. Use "buffer2 = buffer1.copy();" or "Buffer buffer2{buffer1.copy()};" to make it explicit that copying is intended.

### DIFF
--- a/include/qpdf/Buffer.hh
+++ b/include/qpdf/Buffer.hh
@@ -41,10 +41,10 @@ class Buffer
     QPDF_DLL
     Buffer(unsigned char* buf, size_t size);
 
-    QPDF_DLL
-    Buffer(Buffer const&);
-    QPDF_DLL
-    Buffer& operator=(Buffer const&);
+    [[deprecated("Move Buffer or use Buffer::copy instead")]] QPDF_DLL Buffer(Buffer const&);
+    [[deprecated("Move Buffer or use Buffer::copy instead")]] QPDF_DLL Buffer&
+    operator=(Buffer const&);
+
     QPDF_DLL
     Buffer(Buffer&&) noexcept;
     QPDF_DLL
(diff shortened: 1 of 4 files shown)


m-holger:
Is there a way of testing deprecated methods in CI?

jberkenbilt:
> Is there a way of testing deprecated methods in CI?

Yes, there are pragmas you can put in the code to disable warnings around tests. I've definitely done it before. I'd say check git logs from around the time I removed deprecated functions when doing the ABI breaks for qpdf 11. I was probably clear in the history...might coincide with my having bumped the major version. If you can find it, let me know, and I'll dig. I'll be away from the computer for a while.

m-holger:
Looks like I found it. Thanks

m-holger:
I guess it would have been a good idea to read the README-maintainer 641e92c6a7662a01f488947c3791f3b77e85517f

m-holger:
Documentation is a wonderful thing if people could be bothered to read it.

include/qpdf/Buffer.hh
```diff
@@ -56,6 +56,13 @@ class Buffer
     QPDF_DLL
     unsigned char* getBuffer();
 
+    // Create a new copy of the Buffer. The new Buffer owns an independent copy of the data.
+    QPDF_DLL
+    Buffer copy() const;
+
+    // Only used during CI testing.
```
jberkenbilt, line 63: Please add an ABI comment with a reminder to remove this when we remove the deprecated operator=.

jberkenbilt (changes requested):
Great -- just need a comment so we don't forget to remove setTestMode.

include/qpdf/Buffer.hh
m-holger: Done

jberkenbilt (approved):

PR #985, opened 2023-06-13
 Change JSONHandler::m to std::unique_ptr and declare Members in implementation file

m-holger:


jberkenbilt:
I thought it was only safe to use std::unique_ptr for things that can't be copied. If so, is the change to README-maintainer overreaching? Or am I misunderstanding something?

m-holger:
Objects that use this pattern cannot be default copied. If it is intended for them to be copied, the copy constructor / assignment operator will need to create a copy of the Member object which means that we have a true copy - future changes to one will no affect the other. This is inline with would have happened if no member pattern is used.

jberkenbilt:
Changing from `shared_ptr` to `unique_ptr` is API breaking except when we've deleted the copy constructor. There are certain cases where `shared_ptr` is in use intentionally, though in retrospect, this is bad design. For example, JSON uses shared_ptr which makes copying JSON by object really like copying it by reference (like QPDFObjectHandle). This is clearly documented but, in retrospect, an unfortunate design choice. If I were doing it over, I would have made it not copyable and forced people to put it in a shared pointer if they wanted to copy it.

So, I agree with the code changes here but think the documentation change needs a little tweaking. But I'm going to accept this and then tweak the docs.

jberkenbilt (approved):

m-holger:
> Changing from shared_ptr to unique_ptr is API breaking except when we've deleted the copy constructor. 

It is definitely (I think always)  ABI breaking. I guess that ought to have been made very explicit in the doc.

> For example, JSON uses shared_ptr
 
I agree that JSON is very much like QPDFObjectHandle. All the state is now held by the underlying JSON_value struct and for qpdf 12 this can easily changed so that JSON is simply a shared pointer to JSON_value.

In cases like this I think the pattern of them being a shared pointer to an underlying object does make sense.

jberkenbilt:
I pushed a change directly to main (sorry -- I will try to break that habit) with updates to README-maintainer. If you don't like what I wrote, feel free to suggest further modifications.

jberkenbilt:
We already have something in #785 about revisiting the `shared_ptr` vs `unique_ptr` choice for existing things. When possible, we can switch.

PR #986, opened 2023-06-13
Change code formatting from clang-format >= 15 to clang-format-16

m-holger:
There are minor differences in the output produced by different versions of clang-format. The formatting used at the moment is version 16.

The script is what I am using because the standard script produces formatting that is inconsistent with the code on github.

### DIFF
--- a/format-code
+++ b/format-code
@@ -23,7 +23,7 @@ cd $(dirname $0)
 for i in $(find . -name 'build*' -prune -o '(' \
 		-name '*.hh' -o -name '*.h' -o -name '*.cc' -o -name '*.c' \
 		')' -print); do
-    if clang-format < $i >| $i.new; then
+    if clang-format-16 < $i >| $i.new; then
 	if diff -q $i $i.new >/dev/null 2>/dev/null; then
 	    echo "okay:    $i"
 	    rm $i.new
--- a/manual/contributing.rst
+++ b/manual/contributing.rst
@@ -20,11 +20,10 @@ https://github.com/qpdf/qpdf/discussions.
 Code Formatting
 ---------------
 
-The qpdf source code is formatting using clang-format ≥ version 15
+The qpdf source code is formatting using clang-format version 16
 with a :file:`.clang-format` file at the top of the source tree. The
 :file:`format-code` script reformats all the source code in the
-repository. You must have ``clang-format`` in your path, and it must be
-at least version 15.
+repository. You must have ``clang-format-16`` in your path.
 
 For emacs users, the :file:`.dir-locals.el` file configures emacs
 ``cc-mode`` for an indentation style that is similar to but not


jberkenbilt:
Are you around to resolve merge conflicts? If not, I can resolve this and #985.

m-holger:
yes, I'll push updates in a few minutes

jberkenbilt:
> yes, I'll push updates in a few minutes

Great. I'm going through your notes.md and hoping to get stuff merged. I have a block of time this morning before we get going with weekend plans. I'm itching to get some qpdf time. My migration project at work keeps taking longer than expected. I know, it's a first in engineering history, and I am privileged to be a part of it.

jberkenbilt:
I already have clang-format-17 on my system. I'd rather specify a minimum version and bump it if there are incompatible changes. I'll have to look this over. I could force my system to have clang-format-16.

format-code
```diff
@@ -23,7 +23,7 @@ cd $(dirname $0)
 for i in $(find . -name 'build*' -prune -o '(' \
 		-name '*.hh' -o -name '*.h' -o -name '*.cc' -o -name '*.c' \
 		')' -print); do
-    if clang-format < $i >| $i.new; then
+    if clang-format-16 < $i >| $i.new; then
```
jberkenbilt, line 26: What I'd like to see here is a check that the version is >= 16 since this works with 16 and 17. I think we should just bump the minimum version required whenever a new version changes the output. If I recall, the changes from 15 to 16 were bug fixes. I'd suggest just putting a version check here.

manual/contributing.rst
```diff
@@ -20,11 +20,10 @@ https://github.com/qpdf/qpdf/discussions.
 Code Formatting
 ---------------
 
-The qpdf source code is formatting using clang-format ≥ version 15
+The qpdf source code is formatting using clang-format version 16
```
jberkenbilt, line 23: Let's leave this as ≥ 16 here and below.

jberkenbilt (changes requested):
I'd prefer to just bump the minimum to 16 rather than locking to 16. I'm okay with updating the docs and leaving format-code as it is, and I'm also okay with adding code in format-code that enforces a minimum version.

format-code
m-holger: The reason why I suggested having a fixed version was that the formatting script never worked for me. I always had to use a modified script with the clang version hard coded.

I have no problem using 17 instead of 16.

manual/contributing.rst
m-holger: If we stick with >= 16, should we add a note saying how to fix format-code if it picks up the wrong version of clang-format?

jberkenbilt: Yeah, I think so. I think we can just say that if a newer version of clang-format produces different results, then fix on a more recent version of clang-format by bumping the minimum required version in the docs and the format-code script. Did you have something different in mind? Since both contributing.rst and format-code have this, I'd probably put a comment in each place pointing to the other place.

jberkenbilt: Let me know if you want me to take this. It's probably just a couple of lines of bash. I can probably knock it out in a couple of minutes if you're less comfortable with bash. (I don't know your comfort level with it.)

jberkenbilt: I'm going to push something up.

m-holger: > I don't know your comfort level with it

zero comfort level.

jberkenbilt: See #991

jberkenbilt:
Closing in favor of #991.

manual/contributing.rst
m-holger: > Did you have something different in mind?

What I had in mind was a comment along the lines of "If 'format-code' picks the wrong version of clang-tidy or cannot find clang-tidy, change 'clang-tidy' to 'clang-tidy-xx' where 'xx' is the version installed on your system".

jberkenbilt: Okay, feel free to suggest any tweaks to the documentation to reflect that. Are we going to rely completely on the IDE for clang-tidy or have a .clang-tidy file (assuming there is such a thing)? I have yet to read up on clang-tidy.

m-holger: CLion uses the .clang-tidy file (if so configured).

I tend to use CLion to reformat code not checked into git and the script as a final safety check. I find this avoids the problem of accidental reflowing of strings and comments when I don't terminate blocks correctly.

PR #989, opened 2023-06-15
Add FUTURE build option

m-holger:


jberkenbilt (approved):
I have no objection to this. I might want to write a little more documentation around it. I'll approve now and merge when I have time to tweak after merge. Per your question about cpp20, I don't think we need a flag for it. I think just adding a CI test that explicitly configures the build to use the --std=c++20 (or whatever it is), much like I do for unsigned char or sanitizers, should be adequate. I'd probably have a CI test that just tests with the next standards version or perhaps the latest one. I'm looking forward to C++20 because I want std::filesystem among other things. Maybe in 2024.

jberkenbilt:
See comments in #992

PR #990, opened 2023-06-15
Ensure copied annotations are fixed (fixes #600)

### LINKED ISSUE #600: Annotations on duplicated pages
When a page is added by qpdf to an output file multiple times (eg using  ` --pages . 1,1 --`), all copies of the page share the same `Annots` array (if present) and `Annots` dictionaries.

This is non-compliant with ISO 32000-1 which specifies in "12.5.2 Annotation Dictionaries"

 > A given annotation dictionary shall be referenced from the Annots array of only one page

Sharing of the `Annots` array/dictionaries causes unexpected behaviour that for the user of the PDF file may be difficult to understand, such as:
 
- editing an annotation on one page changes the the annotations on all copies of the page
- a newly added annotation may or may not appear on all copies of the page (depending on whether the copied page had a (possibly empty) `Annots` array

m-holger:
Case where pages from a main input without /AcroForm key are duplicated was omitted

### DIFF
--- a/libqpdf/QPDFJob.cc
+++ b/libqpdf/QPDFJob.cc
@@ -2496,7 +2496,7 @@ QPDFJob::handlePageSpecs(QPDF& pdf, std::vector<std::unique_ptr<QPDF>>& page_hea
             // the original file until all copy operations are completed, any foreign pages that
             // conflict with original pages will be adjusted. If we copy any page from the original
             // file more than once, that page would be in conflict with the previous copy of itself.
-            if (other_afdh->hasAcroForm() && ((!this_file) || (!first_copy_from_orig))) {
+            if ((!this_file && other_afdh->hasAcroForm()) || !first_copy_from_orig) {
                 if (!this_file) {
                     QTC::TC("qpdf", "QPDFJob copy fields not this file");
                 } else if (!first_copy_from_orig) {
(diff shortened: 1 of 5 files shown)


jberkenbilt:
Can you provide a verbal description of how you generated the new test files? The fix looks right, but I want to make sure the tests are right, and a description could save me some detective work.

jberkenbilt:
Once you provide your description, I'll download and check the test files to make sure they look right, or you can just include in your description, "I modify file X by ...., then ran this command to get the output file, then checked the output file to make sure it ...". To answer your other question, I expect that this will be more fully addressed with the pages work, but this is based on comments on #600 and the pages tag since this issue was paged out of my brain. In any case, I intend to accept this PR with the requested information.

m-holger:
The input 'annotations-no-acroform.pdf' is 'form-fields-and-annotations.pdf' with the /AcroForm entry removed from the root dictionary and run through 'fix-qdf'. The output files where generated using
```
qpdf --qdf --static-id --no-original-object-ids  annotations-no-acroform.pdf --pages . 1,1 -- annotations-same-file.out.pdf
qpdf --qdf --static-id --no-original-object-ids minimal.pdf--pages annotations-no-acroform.pdf 1,1 -- annotations-foreign-file.out.pdf
```
In the output files the /Annots arrays for the the duplicated pages ((5 0) and (6 0) / (4 0) and (5 0) respectively)  do not share any common objects.

jberkenbilt:
Thanks. I'll review by pulling these down and looking at them, and if all is good, I'll go ahead and merge.

jberkenbilt (approved):
Looks good, and the tests look right as well. Thanks!

PR #992, opened 2023-06-17
Future

jberkenbilt:
@m-holger This  incorporates your `future` PR #989 (and includes its commits) and adds a little more documentation. I suggest you grab the commit from here and append it to #989. Once you have done this, I will close this PR. Then you can add your operator bool either to #989 or a separate PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -104,6 +104,8 @@ option(INSTALL_PKGCONFIG "Install pkgconfig file" ON)
 option(INSTALL_CMAKE_PACKAGE "Install cmake package files" ON)
 option(INSTALL_EXAMPLES "Install example files" ON)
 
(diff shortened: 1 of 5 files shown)


m-holger:
Why not just close 989 in favor of this one and merge this. I am not possessive about my PRs.

I am happy with your addition, although personally I would be slightly more liberal. To my mind anybody who is happy to run their programs against a local build of the current unreleased head of qpdf is welcome to use future as long as they accept the caveats.

m-holger:
> Then you can add your operator bool either to https://github.com/qpdf/qpdf/pull/989 or a separate PR.

I prefer to keep that out until this can pass the pikepdf tests.

jberkenbilt:
> To my mind anybody who is happy to run their programs against a local build of the current unreleased head of qpdf is welcome to use future as long as they accept the caveats.

True. I'll soften a little and then just merge. Since it includes your original commits, yours will also be marked as merged.

PR #995, opened 2023-06-23
Refactor linearization

m-holger:


libqpdf/QPDF_linearization.cc
```diff
@@ -1058,83 +1054,53 @@ QPDF::calculateLinearizationData(std::map<int, int> const& object_stream_data)
...
-            lc_outlines.insert(og);
-        } else if (in_open_document) {
-            lc_open_document.insert(og);
-        } else if ((in_first_page) && (others == 0) && (other_pages == 0) && (thumbs == 0)) {
-            lc_first_page_private.insert(og);
-        } else if (in_first_page) {
-            lc_first_page_shared.insert(og);
-        } else if ((other_pages == 1) && (others == 0) && (thumbs == 0)) {
-            lc_other_page_private.insert(og);
-        } else if (other_pages > 1) {
```
m-holger, line 1130: @jberkenbilt 

I think the test above should be ```other_pages > 0```. As it stands, if 
```other_pages == 1 && (others > 0 || thumbs > 0)```
the object ends up in part9 rather than part6.   This doesn't make sense to me and is also inconsistent with how shared objects refered to by the first page are handled.

File: libqpdf/QPDF_linearization.cc [hunk 2]
```diff
@@ -1058,83 +1054,53 @@ QPDF::calculateLinearizationData(std::map<int, int> const& object_stream_data)
...
-            lc_other_page_shared.insert(og);
-        } else if ((thumbs == 1) && (others == 0)) {
-            lc_thumbnail_private.insert(og);
-        } else if (thumbs > 1) {
```
m-holger, line 1134: ditto

libqpdf/QPDF_linearization.cc
m-holger: On reflection, I think the problem is as described in #997

jberkenbilt:
I haven't had a chance to look at this yet, but two things I should mention (you probably already know):
* Linearization is the ugliest part of the qpdf codebase. I wrote the whole thing in three days very close to the beginning of qpdf. I have always wanted to find time to gut it.
* There are a handful of open issues regarding linearization. Some of them are related to missing features, and some are related to better inspection. There are also notes in the TODO file.

As you may have already noticed or been aware, one significant aspect of linearization is knowing what objects use what other objects. This is mostly handled in QPDF_optimization.cc. That code is almost certainly inefficient in terms of copying, extra allocations, etc. Then there's the hint stream generation. I remember having worked a little bit to optimize the bitstream code back in the early days, maybe even pre-qpdf when I was using some of that code for image processing, but it is surely been 20 years since I gave it any real attention.

qpdf's linearization is not a 100% complete implementation of the spec, but in the early days of my studying linearized files and implementing the spec, I didn't ever find even a single implementation of linearization that was complete, including anything from Adobe. I used to test linearization by creating a large linearized file and downloading it over an artificially slowed down internet connection using the adobe PDF plugin. This method hasn't worked for ages. It has been a very long time since I have tested to see whether qpdf's linearized files actually load faster in the browser or whether any PDF readers in use these days still optimize loading for linearized files. I'm guessing they do since people are still using qpdf for linearization, but it's certainly not as common as it used to be. For years, Google used qpdf to linearize files when exporting Google docs to PDF, but maybe two years or so ago, they stopped doing this as PDFium has gotten more advanced.

There's probably more, but that's a good brain dump of my thoughts on linearization. All that said, I don't think it would take a huge amount of effort to significantly improve the code. There is probably quite a bit of "low-hanging fruit" there.

jberkenbilt:
I am not likely to consider this in the foreseeable future.

PR #997, opened 2023-06-24
Linearization parts

m-holger:
Spec 1.7 F.3.1 contains

> ... all shared objects (objects referenced from more than one page, not counting objects referenced from the first page) ...

This suggests to me that shared objects are objects shared between pages rather than objects shared between a page and another non-page object. This interpretation results in objects that are reachable from a low priority object and a higher priority object are promoted to the higher priority part rather than demoted to the lower priority part, which intuitively seems to make sense.

This suggests to me that the tests to determine allocation to parts should be as given here.

(Some tests suppressed.)


jberkenbilt:
I don't want any changes that cause objects to land in different places or different orders. There are very few programs that create linearized files, and they all do it a little differently. When I originally implemented linearization, the spec was a guide, but a bigger guide was how Adobe Acrobat actually generated linearlized files and, simulating a very slow connection, what the Acrobat browser plugin of the day accepted that enabled it to display a partially downloaded file. A lot of this was based on observation and heuristic. Linearization is what put qpdf on the map. For years, Google used it to linearize PDF files when exported from Google Drive, and I know of a handful of commercial software operations that use qpdf for linearization. When I was at Apex, we created millions of pages of linearized PDF using the current algorithm. Linearization is less and less relevant as network bandwidth increases and other ways are more commonly used for distributing content over slow connections. I think that changing qpdf's linearization is much more likely to cause problems for people than leaving it alone even if its current behavior may not exactly match the spec.

PR #998, opened 2023-06-25
Add content table to TODO file

m-holger:


m-holger:
Re-flowing was done by CLion with Editor/Code Style/Markdown options set to Hard wrap at = 100, When reformatting keep line breaks inside text blocks = un-set, keep indents on empty lines = un-set.

The final commit has the required hand adjustments.

Overall I find that CLion does a reasonable job formatting markdown, including removing extraneous white-space. For working with markdown I find it useful to have  Editor/Code Style/Markdown option wrap on typing set.

jberkenbilt (approved):
Thanks. For the last 500 years or so, I've been using long lines for markdown. Somehow I was under the mistaken impression that line breaks were significant. Perhaps it's in an older dialect of markdown, the GFM (GitHub Flavored Markdown) seems to be prevalent. I like flowed text like this better. I will have to tweak my emacs settings. Alas, now all my old markdown is going to annoy me. Oh well. Thanks for all this.

PR #999, opened 2023-06-25
Add doxygen files to .gitignore

m-holger:
For convenience.

Also fix formatting error.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ appimage/build
 /distribution/
 .cache
 /.idea
+/html
+Doxyfile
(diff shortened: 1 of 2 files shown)


jberkenbilt:
Curious...why do we need this? Is it something CLion is doing? I'd considered creating doxygen comments years ago but never did. If the IDE consumes them, it could be useful. Using GoLand, I have become much more diligent about doc comments, and I now fully understand some of your earlier changes of moving comments outside of functions. It's weird....things I've been doing for decades now feel wrong to me, having been replaced by things I've only been doing for a few months. Even my muscle memory is shifting a little. I always code with GoLand and emacs side by side, and I doubt I will be comfortable working on qpdf without CLion pretty soon.

m-holger:
We don't need it - (at the moment) it is purely for my convenience. I use doxygen at times to navigate through qpdf.

That said CLion supports doxygen and I think it is definitely worth considering switching to doxygen style commands.

jberkenbilt (approved):

PR #1004, opened 2023-07-08
Split QPDF::readObject into readTrailer, readObject and readObjectInStream

m-holger:
Mainly to improve code readability, but also for a small performance improvement and to suppress misleading warning messages for corrupt trailers.


jberkenbilt:
This looks good with a cursory review. Do you feel like you need me to do a thorough review? At this point, I'm willing to trust your change based on a cursory review when it is this kind of refactoring and tweaking private API. I read through all the commits but didn't go through the mental exercise of absolutely convincing myself that the logic is equivalent. I think the removal of tons of spurious warnings in these broken cases is a good benefit. Clearly those warnings are not of any real use.

I'll merge, but if you want me to scrutinize, I can do so after the fact, and you can do another PR with any suggested changes.

I'll try to find time tomorrow to release 11.5.0.

jberkenbilt (approved):
Approved based on a quick review, agreement that the private API changes are improvements, and recognition that the removal of useless warnings is of benefit.

PR #1012, opened 2023-07-24
Fix QPDFObjectHandle::disconnect (partially fixes #1011)

### LINKED ISSUE #1011: qpdf/libqpdf/QPDFObjectHandle.cc:235:null pointer
When using the qpdf --empty --pages command to merge 1.pdf with other PDF files, a segmentation fault occurred due to a null pointer.

version:11.5.0
My system OS：Ubuntu 22.04
My Compilation Process：
git clone https://github.com/qpdf/qpdf
cd qpdf
cmake -S . -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo ..
cd build
cmake --build .
sudo cmake --install .

reproduce: qpdf --empty --pages 1.pdf 2.pdf -- output.pdf

The stack information is as follows：

Program received signal SIGSEGV, Segmentation fault.
QPDFObjectHandle::disconnect (this=this@entry=0x7fffffffdd20) at /home/ljh/qpdf/libqpdf/QPDFObjectHandle.cc:235
235	        this->obj->disconnect();
(gdb) bt
#0  QPDFObjectHandle::disconnect (this=this@entry=0x7fffffffdd20)
    at /home/ljh/qpdf/libqpdf/QPDFObjectHandle.cc:235
#1  0x00007ffff7dc58f6 in QPDFObjectHandle::DisconnectAccess::disconnect (
    o=...) at /home/ljh/qpdf/include/qpdf/QPDFObjectHandle.hh:1268
#2  QPDF_Dictionary::disconnect (this=0x555555619ba0)
    at /home/ljh/qpdf/libqpdf/QPDF_Dictionary.cc:52
#3  0x00007ffff7cb9056 in QPDFObject::disconnect (this=0x555555617790)
    at /home/ljh/qpdf/libqpdf/qpdf/QPDFObject_private.hh:148
#4  QPDF::~QPDF (this=this@entry=0x5555555a4700, __in_chrg=<optimized out>)
    at /home/ljh/qpdf/libqpdf/QPDF.cc:236
#5  0x00007ffff7c6b4cb in std::default_delete<QPDF>::operator() (
    __ptr=0x5555555a4700, this=<optimized out>)
    at /usr/include/c++/11/bits/unique_ptr.h:85
#6  0x00007ffff7c6f03e in std::default_delete<QPDF>::operator() (
    __ptr=<optimized out>, this=<optimized out>)
    at /usr/include/c++/11/bits/unique_ptr.h:79
#7  std::unique_ptr<QPDF, std::default_delete<QPDF> >::~unique_ptr (
    this=0x7fffffffddc8, __in_chrg=<optimized out>)
    at /usr/include/c++/11/bits/unique_ptr.h:361
#8  QPDFJob::createQPDF (this=<optimized out>)
    at /home/ljh/qpdf/libqpdf/QPDFJob.cc:481
#9  0x00007ffff7d375f7 in QPDFJob::run (this=this@entry=0x7fffffffde70)
    at /home/ljh/qpdf/libqpdf/QPDFJob.cc:516

[1.pdf](https://github.com/qpdf/qpdf/files/12141225/1.pdf)
[2.pdf](https://github.com/qpdf/qpdf/files/12141226/2.pdf)

m-holger:


jberkenbilt:
Should this be merged when through CI, or should I wait for additional fixes?

jberkenbilt:
I'd probably do a 10.5.1 to get this out but not immediately...once we're sure we have a good fix. But as soon as I could make the time to churn through the mechanics.

m-holger:
I think this should be merged as it fixes the nullptr access in disconnect, which is an issue in its own right.

There is a second issue with copyForeignObject, which may return an uninitialized object handle. In this case it is due to an attempt to copy an "annotation", which happens to be the root of the pages tree. It probably should return a null object instead. Do you agree?

jberkenbilt (approved):

jberkenbilt:
> There is a second issue with copyForeignObject, which may return an uninitialized object handle. In this case it is due to an attempt to copy an "annotation", which happens to be the root of the pages tree. It probably should return a null object instead. Do you agree?

Sounds reasonable, but I think I need to understand the case a little more clearly.

PR #1013, opened 2023-07-25
Simplify QPDFObjectHandle::isPageObject

m-holger:
The removed code is redundant for pages that are part of the page tree due to the repairs carried out by getAllPages.

If the intention is to identify page objects that are not part of the pages tree a comment explaining that use would be helpful, including the reasoning for accepting a damaged page object without warning/repair.


jberkenbilt (approved):
Right, thanks. I guess when I added the code in one place, I forgot it was already in the other place.

PR #1014, opened 2023-07-25
Change QPDF::copyForeignObject to return a null object when called wi…

### LINKED ISSUE #1011: qpdf/libqpdf/QPDFObjectHandle.cc:235:null pointer

m-holger:
…th a /Pages object (fixes #1011)

If copyForeignObject is called with a pages object, the object will not be inserted into the obj_copier.obj_map during normal processing. In the return statement the missing entry is default constructed, with the consequence that an uninitialized object handle is returned. All future references to the object are also replaced with an uninitialized object rather than a null object.

### DIFF
--- a/libqpdf/QPDF.cc
+++ b/libqpdf/QPDF.cc
@@ -2034,13 +2034,13 @@ QPDF::copyForeignObject(QPDFObjectHandle foreign)
     }
     obj_copier.to_copy.clear();
 
-    auto& result = obj_copier.object_map[foreign.getObjGen()];
-    if (!result.isInitialized()) {
-        result = QPDFObjectHandle::newNull();
-        warn(damagedPDF("Unexpected reference to /Pages object while copying foreign object. "
-                        "Replacing with Null object."));
+    auto og = foreign.getObjGen();
+    if (!obj_copier.object_map.count(og)) {
+        warn(damagedPDF("unexpected reference to /Pages object while copying foreign object; "
+                        "replacing with null"));
+        return QPDFObjectHandle::newNull();
     }
-    return result;
+    return obj_copier.object_map[foreign.getObjGen()];
 }
 
 void
--- a/qpdf/qtest/qpdf/copy-foreign-objects-25.out
+++ b/qpdf/qtest/qpdf/copy-foreign-objects-25.out
@@ -1,1 +1,2 @@
+WARNING: minimal.pdf (object 6 0, offset 556): Unexpected reference to /Pages object while copying foreign object. Replacing with Null object.
 test 25 done
--- a/qpdf/qtest/qpdf/copy-foreign-objects-25.out
+++ b/qpdf/qtest/qpdf/copy-foreign-objects-25.out
@@ -1,2 +1,2 @@
-WARNING: minimal.pdf (object 6 0, offset 556): Unexpected reference to /Pages object while copying foreign object. Replacing with Null object.
+WARNING: minimal.pdf (object 6 0, offset 556): unexpected reference to /Pages object while copying foreign object; replacing with null
 test 25 done
--- a/qpdf/test_driver.cc
+++ b/qpdf/test_driver.cc
@@ -954,6 +954,8 @@ test_25(QPDF& pdf, char const* arg2)
     // Copy qtest without crossing page boundaries.  Should get O1
     // and O2 and their streams but not O3 or any other pages.
 
+    // Also verify that attempts to copy /Pages objects return null.
+
     assert(arg2 != nullptr);
     {
         // Make sure original PDF is out of scope when we write.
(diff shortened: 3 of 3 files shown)


m-holger:
There probably should be a test case for this.

jberkenbilt:
Is this made obsolete by #1028?

m-holger:
No. This one deals with severely damaged files with garbage objects pretending to be page tree nodes

jberkenbilt:
I agree that there should be a test for this. Do we know how to reproduce it? Anyway, the new code is clearly better than the old code, so I'm inclined to merge it, though I'm worried that taking the change will cover up some other bug. I think I need to understand better when this happens. I think it would be better if we never inserted uninitialized objects rather than detecting them and returning null instead.

m-holger:
> Do we know how to reproduce it? 

All that is required is to change the /Type entry of a top level foreign object, e.g. the /AcroForm dictionary, to /Pages. 
That will result in the ObjGen not being inserted into obj_copier.object_map. On trying to retrieve the missing object from the object_map it gets default constructed as an uninitialized object handle. Similarly, deliberately trying to foreign copy a /Pages  object will have the same result.

m-holger:
> Is this made obsolete by https://github.com/qpdf/qpdf/pull/1028?

> No. This one deals with severely damaged files with garbage objects pretending to be page tree nodes

Sorry, it's been a while since I did this. This PR is actually the first commit in #1028

m-holger:
I've added a test.

>  I think it would be better if we never inserted uninitialized objects rather than detecting them and returning null instead.

I thought it was alright given that the check/fix was at the point of default creation but I've updated to move the check to before the map access.

libqpdf/QPDF.cc
```diff
@@ -2034,13 +2034,13 @@ QPDF::copyForeignObject(QPDFObjectHandle foreign)
     }
     obj_copier.to_copy.clear();
 
-    auto& result = obj_copier.object_map[foreign.getObjGen()];
-    if (!result.isInitialized()) {
-        result = QPDFObjectHandle::newNull();
+    auto og = foreign.getObjGen();
+    if (!obj_copier.object_map.count(og)) {
         warn(damagedPDF("Unexpected reference to /Pages object while copying foreign object. "
```
jberkenbilt, line 2039: I missed this before, but to be consistent with the style of most warnings, you could you change this to lower-case and semicolon-separated, as in `unexpected reference to /Pages object while copying foreign object;  replacing with null`

jberkenbilt (changes requested):
Great, thanks. Requesting a small change to the warning that I overlooked in my previous review.

jberkenbilt:
I'm stepping away for a few hours, but if this is not updated when I return, I will merge as is and tweak the warning message in a separate commit.

jberkenbilt (approved):

PR #1016, opened 2023-07-27
Change QUtil::read_file_into_string to work with pipes (fixes #1010)

### LINKED ISSUE #1010: fix-qdf cannot read from a pipe
This used to work, but it looks like it was broken by [011b1d7](https://github.com/qpdf/qpdf/commit/011b1d7e3af803991ee7e4143b01965296af3cce). The `QUtil::read_lines_from_file()` function would read the input in chunks (albeit extremely inefficiently), reallocating as needed; the `QUtil::read_file_into_string()` function that replaces it attempts to preallocate a buffer large enough to hold the entire input and then read the whole thing in one go. This fails when the input stream is unseekable.

A simple solution would be to check the return code from `fseek()` and, if it fails, fall back to an alternative method like `QUtil::read_lines_from_file()` (slow and functional is better than non-functional, though it wouldn't be a bad idea to update that function to read more than a single byte at a time...).

The ideal solution would rewrite the input processing to handle the file line-by-line without needing to read the whole thing in beforehand. In addition to resolving this issue, that would also solve the outstanding problem of high memory usage for large PDFs (which will ultimately result in the allocation failing if the input is too large to hold in memory all at once). This would be more complex than simply patching `QUtil::read_file_into_string()`, but not inordinately so; there is no special reason why `QdfFixer::processLines()` should iterate over a string or list of strings as opposed iterating over the input stream directly with e.g. `fscanf()` or `std::getline()` (at a glance, it looks like this would actually be a substantial simplification compared to the existing code).

m-holger:
The changed function passes locally fix-qdf.test extended with
```perl
for (my $n = 1; $n <= 2; ++$n)
{
    $td->runtest("fix-qdf $n",
                 {$td->COMMAND => "cat fix$n.qdf | fix-qdf "},
                 {$td->FILE => "fix$n.qdf.out",
                  $td->EXIT_STATUS => 0});

    $td->runtest("identity fix-qdf $n",
                 {$td->COMMAND => "fix-qdf fix$n.qdf.out"},
                 {$td->FILE => "fix$n.qdf.out",
                  $td->EXIT_STATUS => 0});
}
```

I am not sure how to test reading from a pipe in a portable way, or how to limit tests to Linux.

I did not follow the suggestions to use fscanf (too complex) or std::getline (all the existing code dealing with platform difference is based on C rather than C++ streams; this option would be attractive if/when this changes).

### DIFF
--- a/libqpdf/QUtil.cc
+++ b/libqpdf/QUtil.cc
@@ -1178,21 +1178,43 @@ std::string
 QUtil::read_file_into_string(FILE* f, std::string_view filename)
 {
     fseek(f, 0, SEEK_END);
-    auto size = QIntC::to_size(QUtil::tell(f));
-    fseek(f, 0, SEEK_SET);
-    std::string result(size, '\0');
-    if (auto read = fread(result.data(), 1, size, f); read != size) {
+    auto o_size = QUtil::tell(f);
+    if (o_size >= 0) {
+        // Seekable file
+        auto size = QIntC::to_size(o_size);
+        fseek(f, 0, SEEK_SET);
+        std::string result(size, '\0');
+        if (auto n_read = fread(result.data(), 1, size, f); n_read != size) {
+            if (ferror(f)) {
+                throw std::runtime_error(
+                    std::string("failure reading file ") + std::string(filename) +
+                    " into memory: read " + uint_to_string(n_read) + "; wanted " +
+                    uint_to_string(size));
+            } else {
+                throw std::runtime_error(
+                    std::string("premature eof reading file ") + std::string(filename) +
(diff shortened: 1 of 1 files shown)


jberkenbilt:
Sorry for the long delay. Thanks for doing this. It looks good except for a nit: I avoid using common system call or library names like `read` as local variable names as these sometimes tend to be macros on certain platforms. In any case, I can take this and write good, portable tests for it. I'll try to squeeze it in so I can get 11.5.1 released.

m-holger:
read -> n_read

jberkenbilt (approved):
Merging as is; will then attempt to incorporate your test in a portable way. I am a little surprised what you have isn't already good enough for Windows.

PR #1017, opened 2023-07-29
Check for missing mediaboxes

m-holger:
Where necessary, a letter size mediabox is added. Both document viewer and Okular use letter size for missing media boxes (even when the content is larger than letter).


jberkenbilt (approved):
I am fine with this change. Most of the time, qpdf doesn't modify the data, but I think it's okay with a warning, and there are other places where qpdf modifies the pages tree.

Issue #1024, opened 2023-08-22
QPDF incompatible with LLVM 18's LibC++ due to removal of non-standard char_traits<unsigned char>


zclifford:
Hello,

Once again my company is trying to compile with _bleeding edge C++_. And tasked me with resolving a QPDF compilation error that occurs with LLVM 18's LibC++ -- I don't think this has been released yet, much less to any distros, so standard tooling won't yet reproduce the issue.

This time it's char_traits. Previously LibC++ came with a non-standard implementation of char_traits which would happily work with types not listed in the standard.  (Further links: [Announcement](https://discourse.llvm.org/t/deprecating-std-string-t-for-non-character-t/66779), [Bug](https://reviews.llvm.org/D157058), [Commit](https://github.com/llvm/llvm-project/commit/e30a148b098d462d0267c479cd9e4783363c2761))

> We are deprecating std::char_traits with types other than char, wchar_t, char16_t, char32_t and char8_t, and that specialization will go away in LLVM 18. 

(n.b. char8_t was added to the spec in C++20, and isn't part of C++17)

It seems like QPDF relies on this [here](https://github.com/qpdf/qpdf/blob/5d6ee83e3f39b5ee8768e93599ec09fa66d25908/include/qpdf/Pl_Buffer.hh#L80):

```
...
In file included from third_party/qpdf/src/libqpdf/Pl_Buffer.cc:1:
In file included from third_party/qpdf/src/include/qpdf/Pl_Buffer.hh:32:
In file included from third_party/qpdf/src/include/qpdf/Pipeline.hh:42:
In file included from third_party/stl/cxx17/string:26:
third_party/crosstool/v18/stable/toolchain/bin/../include/c++/v1/string:723:46: error: implicit instantiation of undefined template 'std::char_traits<unsigned char>'
  723 |     static_assert(( is_same<_CharT, typename traits_type::char_type>::value),
      |                                              ^
third_party/qpdf/src/include/qpdf/Pl_Buffer.hh:80:42: note: in instantiation of template class 'std::basic_string<unsigned char>' requested here
   80 |         std::basic_string<unsigned char> data;
      |                                          ^
third_party/crosstool/v18/stable/toolchain/bin/../include/c++/v1/__fwd/string.h:23:29: note: template is declared here
   23 | struct _LIBCPP_TEMPLATE_VIS char_traits;
      |                             ^
...
```

Here `unsigned char` is none of the above types, and so fails to compile.

[...]

PR #1025, opened 2023-08-22
Remove use of non-standard `char_traits<unsigned char>` from Pl_Buffer

zclifford:
`basic_string<unsigned char>` implies use of
`char_traits<unsigned char>`.

This char_traits specialization is not standard C++, and will be removed from LibC++ as of LLVM 18. To ensure continued LibC++ compatibility it needs to be removed.

There are two possible replacements here: `std::string` (e.g. `std::basic_string<char>`), or `std::vector<unsigned char>`.

I have opted for vector since this code is dealing with a binary buffer; though probably either way is fine (why does C++ even have strings anyway??).

https://github.com/qpdf/qpdf/issues/1024

### DIFF
--- a/include/qpdf/Pl_Buffer.hh
+++ b/include/qpdf/Pl_Buffer.hh
@@ -33,6 +33,7 @@
 #include <qpdf/PointerHolder.hh> // unused -- remove in qpdf 12 (see #785)
 
 #include <memory>
+#include <vector>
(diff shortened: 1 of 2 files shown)


jberkenbilt (approved):
Looks like a good change. I will merge after CI passes. Thanks.

PR #1030, opened 2023-09-03
Fix exit code when an input other than the primary input generates wa…

m-holger:
…rnings


jberkenbilt:
Thanks -- I fixed it before I saw this, so closing this in favor of my already committed fix.

jberkenbilt:
though your fix is better, so I'm going to use it.

jberkenbilt:
Your fix waits slightly longer before doing the checks, giving more opportunities for warnings. I wish this were less fragile, but I think this is right now.

jberkenbilt (approved):

PR #1028, opened 2023-08-25
Maintain links to foreign pages when copying foreign objects (fixes #1003)

m-holger:
Use indirect nulls (dangling references) to reserve foreign objects.

Reserve foreign page objects when references to them are encountered so that links are not broken if the page is subsequently copied.

Comments and tests will need to be updated if this approach is accepted.

Issue #1003, opened 2023-07-07
split with hyperlinks


chrisrex:
Hi,
I want to split a multi-page PDF into chapters. There are hyperlinks within the chapters.
This line of code preserves the hyperlinks. Everything is working well:
`"qpdf.exe"  --linearize  "qpdf-test-hyperlinks.pdf" --pages  "qpdf-test-hyperlinks.pdf" 1-4   -- "chapter_01.pdf"`

Now I want to add a cover before each chapter:
`"qpdf.exe"  "cover.pdf " --linearize --pages . "qpdf-test-hyperlinks.pdf" 1-4   -- "chapter_01.pdf"`
The hyperlinks within the chapter no longer work.
The problem: The cover still contains metadata such as title and author. I would like to get that too
Attached are sample files
[cover.pdf](https://github.com/qpdf/qpdf/files/11983619/cover.pdf)
[qpdf-test-hyperlinks.pdf](https://github.com/qpdf/qpdf/files/11983620/qpdf-test-hyperlinks.pdf)

Is there a way to keep these hyperlinks alive?

Best regards.
Christoph

jberkenbilt:
This is a known and long-standing issue and is related to a broader issue. There are many open issues related to this, all tagged with the `pages` label. Fixing this is next up on my qpdf to-do list. It will be a significant effort, but I'm hoping to get a large chunk of time to work on it sometime in the next few months.

chrisrex:
Thank you for your report. I will wait with patience and look forward to the result.
Best regards
Christoph

m-holger:
Hyperlinks are preserved for the main input file. If you make `qpdf-test-hyperlinks.pdf' the main file:
```
qpdf qpdf-test-hyperlinks.pdf --linearize --pages cover.pdf . 1-4 -- chapter1.pdf
```
the hyperlinks to the figures will be preserved.

chrisrex:
Thank you m-holger,

it almost works. When opening chapter_01, page 1 (cover) is not opened, only the second page.

m-holger:
Try opening the file, navigating back one page to the cover page and than exiting the viewer. When you reopen the file, hopefully you will land on the cover page.

chrisrex:
Thanks. Acrobat does not remember the setting when exiting.
In the document settings, "Open on page" says 2 of 5.

m-holger:
Try

```
[...]

PR #1035, opened 2023-09-05
Fix #1028

m-holger:
#1028 has a serious bug, namely it failes to copy a foreign page if previously a foreign object was copied that has an indirect reference to the page.

### DIFF
--- a/libqpdf/QPDF.cc
+++ b/libqpdf/QPDF.cc
@@ -2058,27 +2058,27 @@ QPDF::reserveObjects(QPDFObjectHandle foreign, ObjCopier& obj_copier, bool top)
 
     if (foreign.isIndirect()) {
         QPDFObjGen foreign_og(foreign.getObjGen());
+        if (!obj_copier.visiting.add(foreign_og)) {
+            QTC::TC("qpdf", "QPDF loop reserving objects");
+            return;
+        }
         if (obj_copier.object_map.count(foreign_og) > 0) {
             QTC::TC("qpdf", "QPDF already reserved object");
-            if (obj_copier.visiting.count(foreign_og)) {
-                QTC::TC("qpdf", "QPDF loop reserving objects");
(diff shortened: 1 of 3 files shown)


jberkenbilt:
I have to step away for a bit. Can you explain what exactly the bug was and how this fixes it? It will help me think through the fix and the original problem.

m-holger:
The bug arose from the final commit of the original PR, namely the change from not copying references to pages unless the referenced pages had already been copied, to creating a dangling reference to pages not yet copied which was meant to be fixed up if and when the page was explicitly copied.

The mechanics of this is to insert an indirect null into obj_copier.object_map  when a new page is encountered but not into obj_copier.to_copy unless the page is encountered  as the 'top' object.

The bug was that reserveObjects returned as soon as it found the the object was in obj_copier.object_map, when it was meant to return unless the object is a 'top' page object and the entry in the object map is a null.

jberkenbilt:
I looked at the changes between this commit and 11.5.0, in other words, combining the original fix with this one. I guess it's still not completely clear to me why it was wrong for this to be a reserved object rather than an indirect null. In both cases, the reference should have sprung to life if the object got copied and disappeared if it didn't.

m-holger:
I think so. The commit fixes the serious issue,

To fully fix #1003  some code will be needed that the deals with named destinations, but this is not time critical (and may be part of pages anyway)

jberkenbilt:
Yes, that's why I reopened the issue.

m-holger:
> ... and disappeared if it didn't.

There was no obvious mechanism for that.

For the the test case, prior to 1028 all the links on page one would have been broken, on page 2 the links to page 3 and 4 would be broken and so on.

With 1028, for the test case it makes no difference whether we use nulls or reserved objects. However, if instead of importing the whole file we only imported some pages the test case would fail with a logic error when the links to the missing pages are unparsed in by QPDFWriter.

I did consider whether to suggest that reserved should be treated as null for unparsing and writing as JSON, but that would have made it more difficult for QPDFWriter.

PR #1059, opened 2023-10-28
Refactor QPDFParser::parse

m-holger:
The main changes are:

- combine the state and object stacks. This required the refactoring of some parser states to prevent the two stacks from going out of sync.

- create separate methods for handling the first token and subsequent tokens.

- avoid the creation of temporary QPDF_Integer and QPDF_Name objects by refactoring the parsing of indirect references and dictionary keys.

- create dictionaries on the fly rather than accumulating key and value objects and processing them once the close token has been found. This requires splitting st_dictionary  in st_dictionary_key and st_dictionary_value.

### DIFF
--- a/libqpdf/QPDFParser.cc
+++ b/libqpdf/QPDFParser.cc
@@ -38,11 +38,343 @@ QPDFParser::parse(bool& empty, bool content_stream)
     std::shared_ptr<QPDFObject> object;
     bool set_offset = false;
 
-    std::vector<StackFrame> stack{{input, st_top}};
+//    std::vector<StackFrame> stack{{input, st_top}};
+    stack.clear(); // NEW
+    stack.emplace_back(input, st_top); // NEW
     bool done = false;
     bool b_contents = false;
     bool is_null = false;
-    auto* frame = &stack.back();
+    frame = &stack.back(); // CHANGED
+
+    while (!done) {
+        bool indirect_ref = false;
+        is_null = false;
+        object = nullptr;
+        set_offset = false;
+
+        if (!tokenizer.nextToken(*input, object_description)) {
+            warn(tokenizer.getErrorMessage());
+        }
+        ++good_count; // optimistically
+
+        switch (tokenizer.getType()) {
+        case QPDFTokenizer::tt_eof:
+            if (stack.size() > 1) {
+                warn("parse error while reading object");
+            }
+            if (content_stream) {
+                // In content stream mode, leave object uninitialized to indicate EOF
+                return {};
+            }
+//            QTC::TC("qpdf", "QPDFParser eof in parse");
+            warn("unexpected EOF");
+            return {QPDF_Null::create()};
+
+        case QPDFTokenizer::tt_bad:
+//            QTC::TC("qpdf", "QPDFParser bad token in parse");
+            if (tooManyBadTokens()) {
+                return {QPDF_Null::create()};
+            }
+            is_null = true;
+            break;
+
+        case QPDFTokenizer::tt_brace_open:
+        case QPDFTokenizer::tt_brace_close:
+//            QTC::TC("qpdf", "QPDFParser bad brace");
+            warn("treating unexpected brace token as null");
+            if (tooManyBadTokens()) {
+                return {QPDF_Null::create()};
+            }
+            is_null = true;
+            break;
+
+        case QPDFTokenizer::tt_array_close:
+            if (frame->state == st_array) {
+                if (stack.size() < 2) {
+                    throw std::logic_error("QPDFParser::parseInternal: st_stop encountered with "
+                                           "insufficient elements in stack");
+                }
+                object = QPDF_Array::create(std::move(frame->olist), frame->null_count > 100);
+                setDescription(object, frame->offset - 1);
+                // The `offset` points to the next of "[".  Set the rewind offset to point to the
+                // beginning of "[". This has been explicitly tested with whitespace surrounding the
+                // array start delimiter. getLastOffset points to the array end token and therefore
+                // can't be used here.
+                set_offset = true;
+                stack.pop_back();
+                frame = &stack.back();
+            } else {
+//                QTC::TC("qpdf", "QPDFParser bad array close");
+                warn("treating unexpected array close token as null");
+                if (tooManyBadTokens()) {
+                    return {QPDF_Null::create()};
+                }
+                is_null = true;
+            }
+            break;
+
+        case QPDFTokenizer::tt_dict_close:
+            if (frame->state == st_dictionary) {
+                if (stack.size() < 2) {
+                    throw std::logic_error("QPDFParser::parseInternal: st_stop encountered with "
+                                           "insufficient elements in stack");
+                }
+
+                // Convert list to map. Alternating elements are keys.  Attempt to recover more or
+                // less gracefully from invalid dictionaries.
+                std::set<std::string> names;
+                for (auto& obj: frame->olist) {
+                    if (obj) {
+                        if (obj->getTypeCode() == ::ot_name) {
+                            names.insert(obj->getStringValue());
+                        }
+                    }
+                }
+
+                std::map<std::string, QPDFObjectHandle> dict;
+                int next_fake_key = 1;
+                for (auto iter = frame->olist.begin(); iter != frame->olist.end();) {
+                    // Calculate key.
+                    std::string key;
+                    if (*iter && (*iter)->getTypeCode() == ::ot_name) {
+                        key = (*iter)->getStringValue();
+                        ++iter;
+                    } else {
+                        for (bool found_fake = false; !found_fake;) {
+                            key = "/QPDFFake" + std::to_string(next_fake_key++);
+                            found_fake = (names.count(key) == 0);
+//                            QTC::TC("qpdf", "QPDFParser found fake", (found_fake ? 0 : 1));
+                        }
+                        warn(
+                            frame->offset,
+                            "expected dictionary key but found non-name object; inserting key " +
+                                key);
+                    }
+                    if (dict.count(key) > 0) {
+//                        QTC::TC("qpdf", "QPDFParser duplicate dict key");
+                        warn(
+                            frame->offset,
+                            "dictionary has duplicated key " + key +
+                                "; last occurrence overrides earlier ones");
+                    }
+
+                    // Calculate value.
+                    std::shared_ptr<QPDFObject> val;
+                    if (iter != frame->olist.end()) {
+                        val = *iter;
+                        ++iter;
+                    } else {
+//                        QTC::TC("qpdf", "QPDFParser no val for last key");
+                        warn(
+                            frame->offset,
+                            "dictionary ended prematurely; using null as value for last key");
+                        val = QPDF_Null::create();
+                    }
+
+                    dict[std::move(key)] = std::move(val);
+                }
+                if (!frame->contents_string.empty() && dict.count("/Type") &&
+                    dict["/Type"].isNameAndEquals("/Sig") && dict.count("/ByteRange") &&
+                    dict.count("/Contents") && dict["/Contents"].isString()) {
+                    dict["/Contents"] = QPDFObjectHandle::newString(frame->contents_string);
+                    dict["/Contents"].setParsedOffset(frame->contents_offset);
+                }
+                object = QPDF_Dictionary::create(std::move(dict));
+                setDescription(object, frame->offset - 2);
+                // The `offset` points to the next of "<<". Set the rewind offset to point to the
+                // beginning of "<<". This has been explicitly tested with whitespace surrounding
+                // the dictionary start delimiter. getLastOffset points to the dictionary end token
+                // and therefore can't be used here.
+                set_offset = true;
+                stack.pop_back();
+                frame = &stack.back();
+            } else {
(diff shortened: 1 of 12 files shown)


libqpdf/QPDFParser.cc
```diff
@@ -38,11 +38,343 @@ QPDFParser::parse(bool& empty, bool content_stream)
     std::shared_ptr<QPDFObject> object;
     bool set_offset = false;
 
-    std::vector<StackFrame> stack{{input, st_top}};
+//    std::vector<StackFrame> stack{{input, st_top}};
```
m-holger: Add to commit message:

The new method is temporarily an (almost) complete copy of parse, which is
temporarily (almost) unchanged.

libqpdf/QPDFParser.cc
```diff
@@ -33,10 +33,7 @@ QPDFParser::parse(bool& empty, bool content_stream)
     empty = false;
 
     std::shared_ptr<QPDFObject> object;
-    stack.clear();
-    stack.emplace_back(input, st_top);
-    frame = &stack.back();
-    object = nullptr;
+    auto offset = input->tell();
```
m-holger: Remove offset

libqpdf/QPDFParser.cc
```diff
@@ -132,7 +131,7 @@ QPDFParser::parse(bool& empty, bool content_stream)
         return {QPDF_Null::create()};
     }
 
-    setDescription(object, frame->offset);
+    setDescription(object, offset);
```
m-holger: Replace offset with input->tell()

libqpdf/QPDFParser.cc
```diff
@@ -197,17 +193,17 @@ QPDFParser::parseRemainder(bool content_stream)
...
                 setDescription(object, frame->offset - 1);
                 // The `offset` points to the next of "[".  Set the rewind offset to point to the
                 // beginning of "[". This has been explicitly tested with whitespace surrounding the
                 // array start delimiter. getLastOffset points to the array end token and therefore
                 // can't be used here.
                 set_offset = true;
+                if (stack.size() <= 1) {
+                    return object;
+                }
+                set_offset = true;
```
m-holger: remove duplicate

libqpdf/QPDFParser.cc
```diff
@@ -364,15 +348,13 @@ QPDFParser::parseRemainder(bool content_stream)
...
-
-                if (name == "/Contents") {
-                    b_contents = true;
-                } else {
-                    b_contents = false;
-                }
+            if (frame->state == st_dictionary_key) {
+                frame->key = tokenizer.getValue();
+                frame->state = st_dictionary_value;
+                b_contents = decrypter && frame->key == "/Contents";
```
m-holger, line 331: Can you check b_contents carefully. I tried to read the spec but I do not feel confident in my understanding.

jberkenbilt:
Is this serving a purpose other than performance? Is it working toward std::variant or new object handles or one of the other goals?

m-holger:
Two of the commits (elimination of temporary QPDF_Integer and QPDF_Name objects) are purely for performance reasons. They have a slightly greater impact on runtime than the variant work with a lot less effort. (At the moment I see a overall runtime reduction of over 16.5 - 17% with 1059 and variant combined on gcc -12 and clang-15, with just over 10% due to to 1059.)

The rest started of from temporary changes I made while investigating options for the std::variant (and other) refactoring of QPDFObject/QPDFVariant. I feel 1059 makes it easier to investigate options and implement required changes.

jberkenbilt:
Please let me know when this is ready for final review. When I'm ready to dive deeply into the pages work (which I hope is extremely soon now -- one side project slid in front of it, but I might have that wrapped up this weekend), I will probably want to pause on looking at pull requests that are not bug fixes or relevant to the pages work. Getting the pages work done is higher priority for me than string_view or variant or some of the other things we're discussing. Maybe after the pages work is stable, we can open an API window.

I realize that this is a small, tactical change, and I'm still willing to work this one in.

m-holger:
This is ready for review.

jberkenbilt:
It may be a long time before I have time to digest this. To my mind, this makes the parser much harder to reason about. There are now lots of early returns in functions that are doing a lot of things after the ends of switch statements. It's much harder for me to read this code and be confident that the parser is doing what it's supposed to be. I think there are a few good things here in terms of eliminating extra copies, and I know the optimizations have been greatly appreciated by users. What I will have to do to review this is to first reread the parser as it is currently implemented and make sure I remember all the different states it gets into, then try to convince myself that this is doing the same thing, and if I spend time doing that right now, it will be at the expense of other work.

m-holger:
I was rather hoping that it makes the parser easier to reason about, and that the PR would be easy to review. Perhaps it would have been helpful if I had been a bit more verbose.

The general idea is that every token is handled by a single case clause in a single switch statement. (There is an exception for  integer tokens, which I ignore for now.) This compares to the current situation where the processing loop contains:

- one switch statement to do some processing based (mainly) on the token type
- some code to do some sanity checks on the output of the first case statement as well as accounting for bad tokens
- a second switch statement to do further processing based (mainly) on the parser state.  

To get to the desired point, I took the following steps (each in a separate commit. With a few exception, the PR can be reviewed to any commit and can be accepted for the first n commits only): 

- some straight forward fixes to PR 864. See commit message.

- move bad token house keeping to a separate method, ```tooManyBadTokens```, which is called at each point a bad token is encountered. The result of too many bad tokens is always that the parser returns a null object, so no further processing is required in this case.  Key facts for this commit are:
  -  bad tokens are rare
  - counting good tokens is cheap
  - decision making is only required when  (and can be deferred until) a bad token is encountered. 

- refactor eof handling. This is currently split into handling eof token , which transitions into the the st_eof state. Since there is a 1-to-1 correspondence between token and state, the processing can be merged. The eof state always results in the termination of the processing loop and the return of an uninitialized or null object.

- refactor handling of the array_close token. This consists currently  of checking that the current state is st_array and than transitioning to the st_stop state. The st_top state consists of a small amount of code that is common to the processing of both array and dictionary close tokens at the beginning and end if the state code, and sandwiched in between a large amount of code that is specific to one or the other. The common code deals with with popping the current state and object off the stacks. I copied the common state code to the token processing case code as well as moving the array specific state code to the token processing case. A couple of points to take note of:

  - In the existing code, only one state case per token is executed. Logically, once the st_array state is popped off the state stack, the state case for whatever state is on top of  the stack should be executed. Because of the one state case per token rule, this does not happen in the existing code and therefore the relevant parts of code for those states is duplicated at the end of the st_stop code. With state stack popping happening as part of token processing, we can now let the state switch statement take care of it, keeping in mind that the offset for containers (set in setDescription) is different to the offset for scalars and therefore we need to take care to ensure that the offset set in the code for handling the array_close token does not get overwritten later by guarding calls to ```setDescription``` in the state handling code with the ```set_offset``` flag. 

  - Because of the above point, popping of the state and object stacks in the st_stop case code has to happen at different times. This is not required in the new code and to allow future merging of the two stacks, popping of both stacks is synced in the new code.

-  refactor handling of the dict_close token. Same as previous commit, mutatis mutandis. The ```st_stop``` is now redundant and is removed.

(to be continued)

jberkenbilt:
Okay, I'll read over your comments and take a fresh look. It would be good to get this merged given the size of the improvement you are reporting.

m-holger:
(continued)

- refactor handling of  array_open and dict_open tokens. At the moment, the token handling code pushes st_array or st_dictionary onto the state stack for handling subsequent tokens and sets the state temporarily to st_start, which is a no op state. We can simply end the processing of the token here with a  continue statement and remove the st_start state. Note that for excessively deeply nested objects the parser always issues a warning and returns a null object, The existing code also treats the open token as a bad token and possibly issues a further warning about too many errors being encountered, which I do not think adds anything and increases the warning clutter for  badly damaged files. I have removed this extra warning message.

- merge the state and objects stacks. They are now always in sync and therefore there is no longer a reason for maintaining separate stacks. This change simply requires adding a state member to the StackFrame struct and removing all references to the state stack or replacing them with the combined stack. The commit introduces another couple of changes:

  - the definition of the StackFrame struct is moved to the header file to make further changes easier.

  - the definition of frame, a reference to the top of the stack, is changed to a pointer and is updated whenever the stack changes rather than at thye beginning of each loop iteration. This avoids unnecessary access to the stack and also ensures that the frame variable always points to the current stack frame.

  - some references to fields of the current stack frame which are at the moment set at the beginning of each loop iteration have been removed and replaced with access through the frame variable to avoid confusion when the stack changes during a loop iteration.

The next three commits deal with separating the code handling the first token from the code handling subsequent tokens, the reason being that a lot of the existing code only applies to one of these two cases. Separating the cases simplifies the code and allows us to eliminate temporary variables. The separation is achieved by creating a second copy of the parse method, called parseRemainder, which will deal with all but the first token. The details of the commits are:

- create new method parseRemainder. The new method starts of as a straight copy of the parse method. Both parse and parseRemainder are then modified to the minimum extent necessary for the parser to work correctly. Where necessary, code is deleted by commenting it out and additions and modifications are marked "// NEW" or "// MODIFIED".  The necessary changes are:

  - The stack and frame variables are changed into class variables as they are required by both (and later other) methods.

  - All QTC test cases are commented out in one of the two methods.

  - in parse, in the case handling array_open / dict_open tokens, add code to call parseRemainder and return the result. These are the only first tokens for which the parser ever processes more than one token.

  - in paseRemainder, don't create a  ParseGuard object (which has already be created in parse. Only one object can exist at a time) and don't set the empty parameter (which indicates that no token has been found, which is obviously false by the time we get here).  

- tidy up the previous commit. Remove temporary comments, remove redundant code and fix the QTC test cases:

  - In parse:

    - Remove the main processing loop and done temporary variable. parse only processes one token.

    - Remove all calls to tooManyBad tokens. If the first token is bad, the existing code allways return a null object. There is also no point incrementing good_count as no previous bad token can have been encountered.

    - Remove the is_null and null_oh variables. Where necessary, null objects are immediately returned.

    - Remove the b_contents and indirect_ref variables. Both require that the are previous tokens.

    - Remove the set_offset variable. It is only used when creating or inserting into container objects.

    = Remove the code for array_close and dict_close  tokens other than the code that treats it as a bad token.

    = Remove the code for array_open and dict_open  tokens remove the code that checks for nesting levels. This is the outermost container. 
    
    - For name and string tokens remove the code that handles /Contents keys. This is a scalar name object, not a dictionary key.

    - For word tokens, remove the code handling indirect reference. Impossible for first token.

    - Remove the state handling switch statement. The only possible state here is st_top, and with the removal of the processing loop this is a no op state.

   - Remove the object sanity checks. With the simplified switch statement it is clear that in all cases either parse returns from the switch statement or sets the object variable. In any case, the code will be modified later to always return from within the switch statement.

  - In parseRemainder, remove checks for state st_top or equivalently, stack.size() == 1,  and code that applies to state st_top only. 

  - Add new QTC entries and add additional test for testing bad first tokens.
   
- Remove state st_top, As st_top now corresponds to the parse method, there is no need for a separate state.  Removing the state has the obvious implication that as there is no state, we cannot  push it onto the stack. The first state on the stack is going to be the state for the outer container and therefore
  
    - in parse , we need a new variable offset to provide the offset that previously would have been picked up from the stack frame.

    - Any tests for stack size must be reduced by 1.
    
    - We cannot use a test for st_top as a condition for detecting the closing of  the outermost container. The condition becomes stack size == 1. With this becoming the condition for returning from parseRemainder, we are automatically protected from popping the last entry off the stack.


(to be continued)

jberkenbilt:
I may run out of time this weekend. Next weekend is also looking busy. :-( But I still intend to do a deep review on this before I move onto my next project. A couple of things have come up that have sucked most of my free time away, but I will get back to it as soon as I can. I haven't had a chance to read over your comments yet, but I appreciate that you have added them.

libqpdf/qpdf/QPDFParser.hh
```diff
@@ -31,18 +31,63 @@ class QPDFParser
     QPDFObjectHandle parse(bool& empty, bool content_stream);
 
   private:
-    enum parser_state_e { st_top, st_start, st_stop, st_eof, st_dictionary, st_array };
+    // Parser state.  Note:
+    // state < st_dictionary_value == (state = st_dictionary_key || state = st_dictionary_value)
```
jberkenbilt, line 35: Should be `state <= st_dictionary_value`

jberkenbilt (approved):
I put some more significant time into reading through this, and I think it looks okay. I find the old code easier to understand, but maybe that's because it's one step closer to what I originally wrote before your last refactor. That said, I can see that this is more efficient -- less copying, etc., so I think it's worth taking, and in some ways, the code is cleaner. I think the differences are our usual stylistic differences. I found one typo in a comment, which I will fix in a separate commit. As a rule, I prefer the `if` half of a conditional to be positive and probably would have reversed the sense of the conditional in the add* methods, but I think there is room for individual style, so I am not going to change it. Thanks for your work on this.

jberkenbilt:
I created a new PR that literally includes these commits via merge to run it through CI again since there have been several CI changes. Once that is through CI, I will merge it, which will bring this along for the ride.

libqpdf/QPDFParser.cc
jberkenbilt: I went through it carefully. It looks correct.

PR #1061, opened 2023-11-03
Extend c api

jberkenbilt:
@m-holger I'm interested in your reaction to the idea implemented in this PR. I haven't written any tests, and I'm not sure I'm going to do it this way.

My use case is that I am writing some go code that needs to interact with qpdf to perform operations that far exceed what the C API can do. The advice I've given people in these cases is to write your own `extern "C"` functions that cal qpdf using the C++ API. Trying to follow this advice myself made me realize that the C API does a lot of the dirty work of handling memory allocation and error handling, and it would be nice to not repeat that. The strategy I implemented here is to allow developers to attach their own stuff to a `qpdf_data` object and get a handle to it that they can retrieve and to provide methods to get the underlying QPDF object and the data back from the `qpdf_data` object. This makes it possible to write a fairly thin (boiler-platey) bit of glue code to expose a C++ function through C. It's not trying to be general like the mechanisms used in Python or SWIG or anything like that -- just simple and special-purpose for QPDF calls.

For my particular use case, I am also considering just integrating through the shell. My use case uses qpdf to do a bunch of analysis on the PDF, then the go code does some processing based on that analysis, and then qpdf does additional work based on the analysis to write a new PDF file. The implementation of my project will be much simpler (and not require cgo) if I just have a pair of shell invocations. In this case, the go is actually replacing image data and could create a qpdf JSON file that the second invocation could use. I'm leaning toward that approach, but since I did this work of allowing user-defined extension to the C API, I might want to keep it, and I wanted your thoughts since you did a lot of work on improving the usability of the C API.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@ cmake_minimum_required(VERSION 3.16)
 # the project line. When updating the version, check make_dist for all
 # the places it has to be updated.
 project(qpdf
-  VERSION 11.6.4
+  VERSION 11.7.0
   LANGUAGES C CXX)
 
 # Enable correct rpath handling for MacOSX
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2023-12-16  Jay Berkenbilt  <ejb@ql.org>
+
+	* Add new C++ functions "qpdf_c_get_qpdf" and "qpdf_c_wrap" to
+	qpdf-c.h that make it possible to write your own extern "C"
+	functions in C++ that interoperate with the C API. See
+	examples/extend-c-api for more information.
+
 2023-12-10  Jay Berkenbilt  <ejb@ql.org>
 
 	* 11.6.4: release
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -34,6 +34,11 @@ foreach(PROG ${EXAMPLE_C_PROGRAMS})
 endforeach()
 target_include_directories(pdf-create PRIVATE ${JPEG_INCLUDE})
 
+# extend-c-api contains a mixture of C and C++ files.
+add_executable(extend-c-api extend-c-api-impl.cc extend-c-api.c)
+set_property(TARGET extend-c-api PROPERTY LINKER_LANGUAGE CXX)
+target_link_libraries(extend-c-api libqpdf)
+
 add_test(
   NAME examples
   COMMAND ${RUN_QTEST}
@@ -47,7 +52,7 @@ add_test(
   --tc "${qpdf_SOURCE_DIR}/examples/*.cc"
   --tc "${qpdf_SOURCE_DIR}/examples/*.c")
 
-file(GLOB EXAMPLES_SRC "*.c" "*.cc")
+file(GLOB EXAMPLES_SRC "*.c" "*.cc" "*.h")
 if(INSTALL_EXAMPLES)
   install(FILES ${EXAMPLES_SRC}
     DESTINATION ${CMAKE_INSTALL_DOCDIR}/examples
--- a/examples/extend-c-api-impl.cc
+++ b/examples/extend-c-api-impl.cc
@@ -1,0 +1,29 @@
+// This is an example of how to write C++ functions and make them usable with the qpdf C API. It
+// consists of three files:
+// - extend-c-api.h -- a plain C header file
+// - extend-c-api.c -- a C program that calls the function
+// - extend-c-api.cc -- a C++ file that implements the function
+
+#include "extend-c-api.h"
+
+// Here, we add a function to get the number of pages in a PDF file and make it callable through the
+// C API.
+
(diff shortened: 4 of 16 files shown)


jberkenbilt:
I have decided that I am not going to use the approach of calling the C API from go code, though I have a very small proof of concept that it works. The go code is very cumbersome, and it adds too much complexity to the build and debugging. I can actually do what I want quite elegantly using qpdf JSON with a pure C++ executable and a pure go executable. So I have to decide whether this idea is worth keeping or whether I should push it to my attic area and document it in TODO in case I ever decide to do it. There are some apparently decent PDF libraries in go (though I haven't personally evaluated them), so it seems not that likely that someone would try to use qpdf from go. There have been attempts at calling to qpdf from rust, but I wonder whether it's really worth it. Mapping between an exception-based, object-oriented language and a language like go or rust is pretty hard, and you give up lots of things in both cases.

m-holger:
My gut feeling is that JSON is probably the more flexible approach for crossing the language barrier, and is probably worth investigating as an alternative.

The main benefit I see is that it allows much more complex data structures to be passed across the C API  which can be processed by the client without the need for further calls to the C API.

One problem I see at the moment is the the current version of is optimized for human consumers/authors. It would be worthwhile to have an alternative aimed at programs. For example, while "1 0 R" or "/Type" are much easier for humans to deal with than ["R", 1, 0] or ["N", "/Type"], the opposite is true for humans.  

I'll think about this further.

jberkenbilt:
I personally think the qpdf JSON format works well for both humans and computers, and I use it programmatically on a routine basis, though so far only for ad-hoc purposes. I find JSON arrays to be problematic in any strongly typed environment. That said, it would be easy enough for someone to create a transformer from one JSON format to another, though I doubt I would accept that in the library -- I don't want to have to maintain backward compatibility and support for more than one format.

I appreciate your input and insight into this. My gut is that, while what I have created here would work, it is not very likely to be used. Decent PDF libraries are showing up in other languages, and I suspect that many will eventually surpass qpdf in capability (or already have). There are still a few things qpdf does that other libraries don't do, and I still think qpdf has one of the best interfaces for working directly with low-level PDF structures, but over time, the need to integrate with qpdf via foreign function calls is probably decreasing, not increasing. Also, people who would be sophisticated enough to write custom libqpdf C++ code that they want to call from other languages are probably able to come up with less cumbersome solutions.

So whether I decide to add this to the library or not boils down to whether I feel like supporting it in the public API. I see it as a fairly low-value addition, but at the same time, it does offer one piece of functionality that is not otherwise available (getting at the underlying QPDF object from the libqpdf side of the C API) and as such makes certain things possible that you can't otherwise do. So I'll probably keep it. It's not that much code, and it doesn't change any existing code except for adding a very small amount of added overhead to the qpdf_cleanup call.

I've decided that the architecture for my project will be to have two programs: one will create a QPDF JSON file, and the second will consume the file after intervening code manipulates it. Using the change in this PR, I could "invoke" the two programs via foreign function interface instead of via the shell, but I don't think there's much of an advantage. Shell invocation is very fast compared to all the work the programs will do, and it avoids the risks of having the C++ part of the program crash and bring down the caller (a problem I have seen all too often with people who try to do this with other libraries from Python code).

In a nutshell, my specific application does this:
* Get a PDF file through some incoming transfer mechanism as it is ingested into our system
* Look at all the images in the file and potentially change their scaling or compression to optimize for size while maintaining a minimum quality threshold
* Replace some of the images based on the result of the analysis
* Split the PDF into pages

My current program is pure C++ with qpdf and GraphicsMagick++ with the occasional call out to ghostscript. The overall system does lots of things that are hard to do in C++ (interfacing with AWS, connecting to Postgres -- stuff the overall system has existing code for). What I'm planning on doing is

* Go code handles everything up to the beginning of the PDF processing
* A C++ program, invoked from the shell, will do minimal analysis to identify by looking at image dictionaries which streams are candidates for image processing and write to a temporary file on disk a qpdf JSON v2 file with json-stream-data=file after applying all filters except DCT and CCITT filters.
* A Go program will ingest the JSON and do all the image manipulation work using superior go libraries that are more performant and have fewer bugs that GraphicsMagick, plus it can do it concurrently with much greater ease. When it's done, it will replace the stream data files and mutate the JSON as needed, potentially dropping some streams that it decides to leave alone, and write out a new qpdf JSON file
* A C++ program will "update from json" using the JSON generated by the go program and continue with the page splitting operations

The overhead of invoking these from the shell will be extremely low in comparison to all the other work, and although there are temporary files involved, qpdf json is efficient about streams (maintaining qpdf's process that no more than one non-object stream is ever loaded at once during normal processing), and in my measurements, the time doing disk I/O is trivial in comparison to the time actually doing image analysis.

I could "invoke" the programs using the C API instead of the shell, but I think the disadvantages of doing that outweigh the advantages.

jberkenbilt:
I'm dropping the extension idea but keeping the two interoperability functions.

PR #1062, opened 2023-11-04
Add new method QPDFObjectHandle::force

m-holger:
Proposed change for qpdf 12.

Requiring that non-compliant operations are explicitly flagged has two benefits:

- it helps with the transparent transition to a more strongly type object handle.

- it helps with avoiding the need to repeatedly check the same structures.

### DIFF
--- a/include/qpdf/QPDFObjectHandle.hh
+++ b/include/qpdf/QPDFObjectHandle.hh
@@ -308,6 +308,17 @@ class QPDFObjectHandle
     QPDF_DLL
     bool isSameObjectAs(QPDFObjectHandle const&) const;
 
+    // Indicate that the object handle is about to be used in a way that violates the PDF
+    // specification. For example, use:
+    //   qpdf.getRoot().replaceKey("/Pages", "<< /Type /Pages /Kids [] /Count 0>>\n"_qpdf.force());
+    // to indicate that you are deliberately using a direct object where the specification requires
+    // an indirect object. For future compatibility with qpdf 12.x.
+    QPDFObjectHandle
+    force() const noexcept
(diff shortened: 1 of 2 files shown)


jberkenbilt:
This will require some discussion. A vast number of PDF files in the wild don't conform to the PDF specification. qpdf doesn't care/know about most aspects of the PDF specification.

m-holger:
The key here is that the behavior for invalid files/operations is undefined. Therefore qpdf can continue to ignore violations that it does not care about. 

However, for features that qpdf does care about it allows qpdf to enforce the condition. At the moment it is perfectly legal to write
```c++
page.replaceKey("/MediaBox", "/Green"_qpdf);
page.replaceKey("/Type", "42"_qpdf);
```
The proposed change would allow qpdf to decide to disallow the above code without it being an API change and therefore requiring a major version change

m-holger:
Of course, the user would be free to write
```c++
page.replaceKey("/MediaBox", "/Green"_qpdf.force());
page.replaceKey("/Type", "42"_qpdf.force());
```
and live with the consequences.

jberkenbilt:
I'm not really ready to consider this yet. I'd have to see some good use cases. We still have the possibility of objects being in multiple places, being moved around, being created out of context and attached to objects, etc. Also, I consider making things that were previously valid to be API breaking even if it is not literally so in the shared library sense.

jberkenbilt:
I don't want to take this change. It moves qpdf in a direction I don't want to go. I also think the abstraction is enforced in the wrong place. I'm happy to discuss, but I see adding this kind of checking to qpdf as low priority.

PR #1067, opened 2023-11-18
Reduce copying of Buffer contents

m-holger:
Pl_Buffer objects are frequently used to build strings. To allow this without unnecessary copying, use std::string for internal storage to allow strings to be created by moving the Pl_Buffer content. 

Add the option to create a Buffer from a std::string&&.

Read stream content in a single read to avoid repeated copying of the Pl_Buffer content as the capacity of the internal string/vector grows.

### DIFF
--- a/libqpdf/QPDF_Stream.cc
+++ b/libqpdf/QPDF_Stream.cc
@@ -216,29 +216,28 @@ QPDF_Stream::getStreamJSON(
     auto dict = this->stream_dict;
     JSON result = JSON::makeDictionary();
     if (json_data != qpdf_sj_none) {
-        std::shared_ptr<Buffer> buf;
+        Pl_Discard discard;
+        Pl_Buffer buf_pl{"stream data"};
+        // buf_pl contains valid data and is ready for retrieval of the data.
+        bool buf_pl_ready = false;
         bool filtered = false;
         bool filter = (decode_level != qpdf_dl_none);
         for (int attempt = 1; attempt <= 2; ++attempt) {
-            Pl_Discard discard;
-            std::shared_ptr<Pl_Buffer> buf_pl;
-            Pipeline* data_pipeline = nullptr;
+            Pipeline* data_pipeline = &discard;
             if (json_data == qpdf_sj_file) {
                 // We need to capture the data to write
-                buf_pl = std::make_shared<Pl_Buffer>("stream data");
-                data_pipeline = buf_pl.get();
-            } else {
-                data_pipeline = &discard;
+                data_pipeline = &buf_pl;
             }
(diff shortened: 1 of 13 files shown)


libqpdf/QPDF_Stream.cc
```diff
@@ -216,29 +216,27 @@ QPDF_Stream::getStreamJSON(
...
+            if (!succeeded || (filter && !filtered)) {
                 // Try again
                 filter = false;
                 decode_level = qpdf_dl_none;
+                buf_pl.getString(); // reset buf_pl
             } else {
-                if (buf_pl.get()) {
-                    buf = buf_pl->getBufferSharedPointer();
+                if (json_data == qpdf_sj_file) {
+                    buf_pl_valid = true;
```
jberkenbilt, line 239: This is less safe than the original code. I think it's correct though. Now we're setting buf_pl_valid based on knowing here that we initialize it based on the same condition earlier. Before, it was setting it based on actually checking the validity of the pointer. The new code is less defensive and would be more likely to result in a crash if the code were changed. I think this is a difference in our coding styles.

libqpdf/QPDF_encryption.cc
```diff
@@ -1046,7 +1046,7 @@ QPDF::decryptStream(
     Pipeline*& pipeline,
     QPDFObjGen const& og,
     QPDFObjectHandle& stream_dict,
-    std::vector<std::shared_ptr<Pipeline>>& heap)
+    std::unique_ptr<Pipeline>& heap)
```
jberkenbilt, line 1049: Now it's not a heap anymore. It can only hold one item. Maybe change the name of the variable. I'd call it `decrypt_pipeline` and put a comment in the implementation stating that the caller needs to own the pointer to ensure that the pipeline lives until it's finished being used. (Where are rust lifetimes when you need them?)

jberkenbilt (changes requested):
One minor suggestion about the name change.

libqpdf/QPDF_Stream.cc
m-holger: I changed buff_pl to a Pl_Buff rather than a shared_ptr<Pl_Buff>, so the object itself will always be valid. Perhaps rename to buf_pl_ready to express that buf_pl.finish() has been called and valid data can be retrieved?

libqpdf/QPDF_encryption.cc
m-holger: Will do.

m-holger: Updated.

jberkenbilt (approved):

PR #1070, opened 2023-11-22
Change default for SF_FlateLzwDecode::columns to 1

m-holger:
Current value is different from the default specified in table 8 of the PDF spec.

Test file is png-filters.pdf with the /Columns entries removed. The file uses /Columns 32 rather that /Columns 1, but I don't know how to generate a test file that genuinely uses /Columns 1.


jberkenbilt:
Hopefully I can catch up on PRs over the break (Thanksgiving here in the US). I'd be grateful for another note indicating which PRs you consider to be ready to go. I really want to start the pages work. Something has come up that has pulled time away I expected to have for this, but I hope to have it resolved in a few weeks. It is very unlikely that I will have anything in time for the Ubuntu 24.04 feature freeze. I would like to focus on stability and make sure that version will be good for long-term support.

I may be able to hand-create a file with /Columns 1 and test it out, but I haven't looked at this part of the spec for ages and don't have it in my head.

m-holger:
I have added the same file with /Columns explicitly set to 1 for comparison.

jberkenbilt (approved):
I'm okay with this and the testing strategy.

PR #1071, opened 2023-11-23
For 'FUTURE' builds add "(Future)" to version string

m-holger:
Also, change so version to 0.


jberkenbilt:
I will do this but in a different way.

jberkenbilt:
Closing in favor of #1081

PR #1080, opened 2023-12-15
Fix option reference in cli.rst

joram-berger:
To reference another command line option description, `:qpdf:ref:` is needed instead of `:qpdf:option:` which is not resolved properly in the HTML documentation.


jberkenbilt:
Thanks. I'll take a look.

jberkenbilt:
Thanks for the fix.

jberkenbilt:
I merged this manually. Stable is a doc branch.

PR #1084, opened 2023-12-19
Zlib ng

jberkenbilt:
Make sure qpdf works with an alternative zlib.


jberkenbilt:
First push: work is incomplete. I want to make sure the zlib-ng tests fail in CI.

PR #1096, opened 2023-12-25
Add additional tests for #1088

m-holger:
Add additional examples of names that are not valid utf8.

Tests marked as expected failures as they currently are not correctly handled.

### DIFF
--- a/qpdf/qtest/qpdf-json.test
+++ b/qpdf/qtest/qpdf-json.test
@@ -356,18 +356,18 @@ $td->runtest("handle binary names",
 $td->runtest("check json",
              {$td->FILE => "a.json"},
              {$td->FILE => "weird-tokens.json"},
-             $td->NORMALIZE_NEWLINES);
+             $td->EXPECT_FAILURE);
 # Make sure we can properly handle JSON with scientific notation.
 $td->runtest("weird tokens round trip json",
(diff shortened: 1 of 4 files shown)


qpdf/qtest/qpdf-json.test
```diff
@@ -356,18 +356,18 @@ $td->runtest("handle binary names",
 $td->runtest("check json",
              {$td->FILE => "a.json"},
              {$td->FILE => "weird-tokens.json"},
-             $td->NORMALIZE_NEWLINES);
+             $td->EXPECT_FAILURE);
```
jberkenbilt, line 359: For future reference, if you want to illustrate a bug, this shuld be `$td->NORMALIZE_NEWLINES | $td->EXPECT_FAILURE` to reduce the likelihood of forgetting to add `$td->NORMALIZE_NEWLINES` when fixing the bug.

jberkenbilt:
This ultimately traces to a problem with my UTF-8 encoder/decoder. How did you think of this?

jberkenbilt:
Thanks for this -- I'm fixing the underlying problem and keeping your test cases, though I manually edited the PDF without using fix-qdf or qpdf to avoid breaking other aspects of the test.

jberkenbilt:
Closed in favor or #1097

Issue #1087, opened 2023-12-21
getDictAsMap/ditems behavior difference for null values


jbarlow83:
getDictAsMap returns keys that are set to null, but ditems omits them.

I'm replacing some usage of getDictAsMap to avoid the extra copying in pikepdf and noticed this. I know nulls aren't that important generally but wanted to bring it to your attention in case this difference wasn't intended. 

It's necessary to use parse to construct a dictionary with a null value since replaceKey deletes it. Or perhaps the issue is that .parse() permits construction of a dictionary with a null value in the first place?

getArrayAsVector and aitems will both output null values.

```c++

#include <qpdf/QPDF.hh>
#include <qpdf/QPDFObjectHandle.hh>

int
main(int argc, char* argv[])
{
    auto d = QPDFObjectHandle::parse("<< /Foo 42 /Bar null >>");

    std::cout << "getDictAsMap iteration:\n";
    for (auto& [k, v]: d.getDictAsMap()) {
        std::cout << k << ": " << v.unparse() << "\n";
    }
    std::cout << "ditems iteration:\n";
    for (auto& [k, v]: d.ditems()) {
        std::cout << k << ": " << v.unparse() << "\n";
    }

    auto a = QPDFObjectHandle::parse("[1 2 null 3]");

    std::cout << "getArrayAsVector iteration:\n";
    for (auto& v: a.getArrayAsVector()) {
        std::cout << v.unparse() << "\n";
    }
    std::cout << "aitems iteration:\n";
    for (auto& v: a.aitems()) {
        std::cout << v.unparse() << "\n";
    }

    return 0;
}

// c++ -I ./include ditems.cpp -lqpdf -Lbuild/libqpdf -o ditems -Wl,-rpath,build/libqpdf
```

Output
```
getDictAsMap iteration:
/Bar: null
/Foo: 42
ditems iteration:
/Foo: 42
getArrayAsVector iteration:
1
2
null
3
aitems iteration:
1
2
null
3
```

jberkenbilt:
I feel like I've noticed this before and think it was just a performance optimization since creating a dictionary using a map can include nulls, and I didn't want to have to prematurely iterate through a dictionary to remove all the nulls. How big of a problem is this? Fixing it would add an extra linear scan of the map in some cases.

jbarlow83:
It's not a problem for me. It's actually a bit more consistent to remove keys for null values. It was just unexpected, and I wanted to check with you if it was an issue.

[...]

PR #1098, opened 2023-12-27
Add new method QPDFObjectHandle::dItems

m-holger:
Inspired by #1087, provide a more efficient alternative to getDictAsMap.

Using dItems in QPDFWriter saves a good 2% on both clang-15 and gcc-12.

I haven't added any documentation or tests yet.

### DIFF
--- a/include/qpdf/QPDFObjectHandle.hh
+++ b/include/qpdf/QPDFObjectHandle.hh
@@ -759,6 +759,10 @@ class QPDFObjectHandle
     QPDF_DLL
     QPDFDictItems ditems();
 
+    class DictItems;
+    QPDF_DLL
+    DictItems dItems();
+
     // Return true if key is present.  Keys with null values are treated as if they are not present.
     // This is as per the PDF spec.
     QPDF_DLL
@@ -1408,6 +1412,41 @@ QPDFObjectHandle operator ""_qpdf(char const* v, size_t len);
 /* clang-format on */
 
 #endif // QPDF_NO_QPDF_STRING
+
+class QPDFObjectHandle::DictItems
+{
+    friend class QPDFObjectHandle;
+
+  public:
+    class iterator: public std::map<std::string, QPDFObjectHandle>::iterator
+    {
+      public:
+        iterator(std::map<std::string, QPDFObjectHandle>::iterator it) :
+            std::map<std::string, QPDFObjectHandle>::iterator(it)
+        {
+        }
+    };
+
+    iterator begin()
+    {
+        return dict.begin();
+    }
+
+    iterator end()
+    {
+        return dict.end();
+    }
+
+  private:
+    DictItems(QPDFObjectHandle& oh, std::map<std::string, QPDFObjectHandle>& dict) :
+        oh(oh),
+        dict(dict)
+    {
+    }
+
+    QPDFObjectHandle oh;
+    std::map<std::string, QPDFObjectHandle>& dict;
+};
 
 class QPDFObjectHandle::QPDFDictItems
 {
--- a/libqpdf/QPDFObjectHandle.cc
+++ b/libqpdf/QPDFObjectHandle.cc
@@ -961,6 +961,19 @@ QPDFObjectHandle::ditems()
     return {*this};
 }
 
+QPDFObjectHandle::DictItems
+QPDFObjectHandle::dItems()
+{
+    static std::map<std::string, QPDFObjectHandle> empty;
+    if (auto dict = asDictionary()) {
+        return {*this, dict->getAsMap()};
+    } else {
+        typeWarning("dictionary", "treating as empty");
+        //        QTC::TC("qpdf", "QPDFObjectHandle dictionary empty map for dItems");
+        return {*this, empty};
+    }
+}
+
 bool
 QPDFObjectHandle::hasKey(std::string const& key)
 {
--- a/libqpdf/QPDFWriter.cc
+++ b/libqpdf/QPDFWriter.cc
@@ -1129,7 +1129,7 @@ QPDFWriter::enqueueObject(QPDFObjectHandle object)
                 enqueueObject(item);
             }
         } else if (object.isDictionary()) {
-            for (auto& item: object.getDictAsMap()) {
+            for (auto& item: object.dItems()) {
                 if (!item.second.isNull()) {
                     enqueueObject(item.second);
                 }
@@ -1487,7 +1487,7 @@ QPDFWriter::unparseObject(
 
         writeString("<<");
 
-        for (auto& item: object.getDictAsMap()) {
+        for (auto& item: object.dItems()) {
             if (!item.second.isNull()) {
                 auto const& key = item.first;
                 writeString(indent);
--- a/libqpdf/QPDF_Dictionary.cc
+++ b/libqpdf/QPDF_Dictionary.cc
@@ -112,8 +112,8 @@ QPDF_Dictionary::getKeys()
     return result;
 }
 
-std::map<std::string, QPDFObjectHandle> const&
-QPDF_Dictionary::getAsMap() const
+std::map<std::string, QPDFObjectHandle>&
+QPDF_Dictionary::getAsMap()
 {
     return this->items;
 }
@@ -138,3 +138,5 @@ QPDF_Dictionary::removeKey(std::string const& key)
     // no-op if key does not exist
     this->items.erase(key);
 }
+
+
--- a/libqpdf/qpdf/QPDF_Dictionary.hh
+++ b/libqpdf/qpdf/QPDF_Dictionary.hh
@@ -24,8 +24,7 @@ class QPDF_Dictionary: public QPDFValue
     bool hasKey(std::string const&);
     QPDFObjectHandle getKey(std::string const&);
     std::set<std::string> getKeys();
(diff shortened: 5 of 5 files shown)


jberkenbilt:
I'm not ultimately going to accept having both dItems and ditems in the API. If we want to swap out the ditems implementation for something more efficient in qpdf 12 that is source-compatible (or very nearly so), I'm okay with that. I know you're still experimenting. I feel that the QPDF API is already too bloated.

m-holger:
Please refer back to #876 for background.

I do not think that ditems can be fixed safely in a way that it could provide a full alternative for getDictAsMap, both in terms of performance and functionality (e.g. getDictAsMap is currently the only way to find dangling references in dictionaries). To my mind that leaves the alternatives of deprecating ditems or marking it as for backward compatibility only.

Personally I don't see a good reason for deprecating ditems. My choice would be to mark both ditems and getDictAsMap as for backward compatibility only, with a recommendation to replace existing uses of ditems and a recommendation to consider using dItems instead of getKeys as in most cases it is a better alternative.

jberkenbilt:
I'm not going to add `dItems`. Our choices are to fix `ditems` or live with it the way it is. I'm still missing why we can't just fix `ditems` instead of adding yet another way to iterate through dictionary items.

m-holger:
For the record, doing naive replacements of a few key calls to ditems, getKeys and getDictAsMap I see performance improvements of 4.75% - 5% for the overall test suite, with over 7% for the linearization tests.

jberkenbilt:
@m-holger I don't understand your last comment. What kinds of naive replacements? Will there be another PR with those that includes these naive replacements? Or are you planning a PR that fixes ditems, getKeys, and getDictAsMap? I'm not saying we shouldn't improve it. Just that I'm not going to add dItems (or aItems) to the API.

jberkenbilt:
@m-holger If I'm missing some reason why you think we can't fix these and have to add a fourth way to iterate through keys, please explain.

jberkenbilt:
Reopening pending answers to questions. Sorry for the noise.

jberkenbilt:
@m-holger I guess I don't understand why we can't just swap out my ditems for your dItems. If it causes minor source breakage that's obvious how to fix, then I'm good with it for qpdf 12. The only thing that would be a problem would be if identical source that meant something completely different such that working code became broken code that compiled without errors.

m-holger:
> I guess I don't understand why we can't just swap out my ditems for your dItems. If it causes minor source breakage that's obvious how to fix ...

Swapping dItems for ditems would compile fine and should work most of the time and for some code all of the time. My concern is that in some circumstances it might work too well for users to realize that with dItems their code only works most of the time. While it would be obvious how to fix the code, it may not be obvious to users that there is a need to fix the code. 

> What kinds of naive replacements? 

I am talking about replacing for loops

```cpp
for (auto& iter: oh.ditems() ) {...}
for (auto& iter: oh.getDictAsMap() ) {...}
for (auto const& key: oh.getKeys() ) {...}
```
with 
```cpp
for (auto& [key, value]: oh.dItems() ) {...}
```
, adding guards against null values where necessary and replacing unnecessary 
```cpp
auto value = oh.getKey(key);
```
calls. I haven't tried to update values directly (i.e. not using replaceKey) or to see whether I can make use of the fact that the dItems iterators are standard map iterators.

> I'm missing some reason why you think we can't fix these

Basically I cannot think of a non-breaking change to ditems or getKeys that would be better than using getDictAsMap instead.

jberkenbilt:
> Basically I cannot think of a non-breaking change to ditems or getKeys that would be better than using getDictAsMap instead.

What would a breaking change look like? That's what I'm still missing. Unless I'm forgetting my own code (which is possible), the iterator returned by `ditems` is a std::pair and should behave pretty much the same as a map iterator. What would break if we just replaced by existing ditems with your ditems? Would some code stop working? Would some code work but behave differently? If the latter, would it behave differently for cases other than keys with direct null values? If your dItems is that much better than ditems, then what I'd like to do is _replace_ ditems with your new implementation for qpdf 12 and document what the risks are, if any. But since I still don't understand any specific example of how your dItems behaves differently from my ditems, I'm not getting what the problem is with this. What would happen (ignoring ABI changes) if we literally just dropped my ditems and used your implementation instead? And is there similar optimization to be gained with aitems?

jberkenbilt:
> y concern is that in some circumstances it might work too well for users to realize that with dItems their code only works most of the time. While it would be obvious how to fix the code, it may not be obvious to users that there is a need to fix the code.

I need a specific example.

m-holger:
>  is there similar optimization to be gained with aitems?

Compared to aitems - yes. Compared to getArrayAsVector - to a much smaller extent. One major overhead for ditems, getDictAsMap and getKeys is that it involves copying a map (or its keys), which requires at least one memory allocation per item (two for longer keys). A second smaller overhead is the reference counting required when copying object handles. For getArrayAsVector only one memory allocation plus reference counting is required in the normal case. (For sparse arrays it is most efficient to copy the elements into a vector and therefore there is no real scope for improvement). 

> What would a breaking change look like? 

Taking getDictAsMap as an example, the obvious change would be to return a const& rather than a value. In many cases this would work without changes. In most others it would require minor changes to variable or loop variable definitions, from an auto& or auto const& to an auto [key, value].  Where the behaviour would potentially change is in cases where the dictionary gets modified in the loop body. In cases where the (original of) the current element gets deleted, the iterator would now be invalidated and the program would now crash. Where elements are changed the behaviour of the loop would be altered, potentially in subtle ways which may not necessarily be detected immediately. While this is easily fixed, it could also be easily overlooked in some programs.

The same problem would occur in any scenario where we tried to eliminated the overhead of copying the dictionary structure. This is also one of the reason why I would be concerned about using dItems as a direct replacement for ditems.

The other reason is the difference in how null values are handled. In many cases this will make no difference either because nulls and missing keys are equivalent or because the type of items of interest is checked in the existing code anyway, but in others it may, eg in code that checks whether a xyzBox can be inherited. I suspect there are not that many PDF files with an /MediaBox explicit null entry and therefore such code may work fine for a long while. 

>  I'd like to do is replace ditems with your new implementation for qpdf 12 and document what the risks are

My view on human nature is that > 50% of users don't read the release notes until after their program doesn't compile, fails its test suite or obviously doesn't work anymore.

m-holger:
> the iterator returned by ditems is a std::pair and should behave pretty much the same as a map iterator

There are a few significant differences. 

- The value pointed to by the iterator can be updated, and the new value will used until the iterator is moved. Once the iterator is moved, the update will be lost. Other iterators pointing at the same element will not see the new value.

- The set of elements that the iterator iterates over is fixed at the time begin() is called. Any items subsequently inserted will be ignored, and any items subsequently deleted will have a null value.

- The values are determined at the time the iterator is moved. Any prior updates made directly to the dictionary are reflected in value accessed through the iterator, but any subsequent updates are ignored until the iterator is moved away and back to the element. 

It is conceivable that iter1 == iter2, but ++iter1 != ++iter2.

jberkenbilt:
I will study it. I feel like these are corner cases, and it might be worth documenting the behavior changes for qpdf 12. It's hard to imagine that anyone would be relying on the behavior of modifying an iterator and expecting the changes to be lost, and I never specified the behavior for what would happen if you modify the dictionary while iterating through it. If there is a significant performance or usability improvement, it might be okay to cause behavior changes in a few corner cases that would not likely occur in typical usage. If there's something dangerous, we can do a DITEMS_TRANSITION or something. I just can't tolerate the idea of the API being "use dItems if you want, modern, efficient behavior, but we'll leave `ditems` there in case you're too lazy to read the release notes.

m-holger:
> It's hard to imagine that anyone would be relying on the behavior of modifying an iterator and expecting the changes to be lost

I agree. However, I think there is a greater chance that some code accidentally works because ditems provides a reference to a copy rather than a reference to the original, and changes made through the iterator are not persistent. That said, I am not sure how widely ditems is used. I cann't think of that many uses for ditems outside of qpdf itself. 

>  I just can't tolerate the idea of the API being "use dItems if you want, modern, efficient behavior, but we'll leave ditems there in case you're too lazy to read the release notes.

In that case my preference would be to deprecate ditems and perhaps getDictAsMap and remove it in qpdf 13.

Incidentally, I think a better name for DictItems  and dItems would be Dict and asDict. The more I think about it, the more I see potential in Dict beyond iterating over its elements. I'll expand on this, but I'm still snowed under with other things so it may take a while.

Relating to this, I think it would be highly beneficial if something like  #979 makes it into qpdf 12. A light review and agreement to the principle would be useful.

jberkenbilt:
Okay, let's hold on dItems for now and come up with a more comprehensive plan on how we want to do this. I'm okay with deprecating ditems and getDictAsMap in favor of something completely different like Dict and asDict for qpdf 12 or 13 depending on the timing. In the interim, let's not introduce churn. #979 is not currently in a reviewable state. I'm not sure of const overloads is worth it. You know I always struggle with the tradeoff between code complexity and performance. I've always liked the fact that QPDFObjectHandle resolves lazily, and it's easier to reason about the code if you don't have to think about whether or not resolving will be required. I basically don't think there is much use for const QPDFObjectHandles. Feel free to make an argument to the contrary on #979 when you get a chance. As you know, I'm not interested in any big changes while I'm working on pages.

m-holger:
superseded by plans to provide an iterable dictionary object

PR #1109, opened 2024-01-07
Tune QPDF::getCompressibleObjGens

m-holger:
A couple of minor changes to container types.

The only controversial bit is a change to loop detection, which only uses the object id (not the generation number). As a result, in files with accessible indirect references  `x y R` and `x z R`, `y != z`, getCompressibleObjGens may exclude one of `x y R` and `x z R`, and possibly other objects only accessible through them.

I don't think its worth the cost of checking for this possibility, otherwise there should also be a test case for this.

### DIFF
--- a/libqpdf/QPDF.cc
+++ b/libqpdf/QPDF.cc
@@ -2377,12 +2377,13 @@ QPDF::getCompressibleObjGens()
     QPDFObjGen encryption_dict_og = encryption_dict.getObjGen();
 
     QPDFObjGen::set visited;
-    std::list<QPDFObjectHandle> queue;
-    queue.push_front(m->trailer);
+    std::vector<QPDFObjectHandle> queue;
+    queue.reserve(512);
+    queue.push_back(m->trailer);
     std::vector<QPDFObjGen> result;
     while (!queue.empty()) {
-        QPDFObjectHandle obj = queue.front();
-        queue.pop_front();
+        QPDFObjectHandle obj = queue.back();
+        queue.pop_back();
         if (obj.isIndirect()) {
             QPDFObjGen og = obj.getObjGen();
             if (!visited.add(og)) {
@@ -2409,18 +2410,18 @@ QPDF::getCompressibleObjGens()
                         QTC::TC("qpdf", "QPDF exclude indirect length");
                     }
                 } else {
-                    queue.push_front(value);
+                    queue.push_back(value);
                 }
             }
         } else if (obj.isDictionary()) {
             std::set<std::string> keys = obj.getKeys();
             for (auto iter = keys.rbegin(); iter != keys.rend(); ++iter) {
-                queue.push_front(obj.getKey(*iter));
+                queue.push_back(obj.getKey(*iter));
             }
         } else if (obj.isArray()) {
             int n = obj.getArrayNItems();
             for (int i = 1; i <= n; ++i) {
-                queue.push_front(obj.getArrayItem(n - i));
+                queue.push_back(obj.getArrayItem(n - i));
             }
         }
     }
--- a/libqpdf/QPDF.cc
+++ b/libqpdf/QPDF.cc
@@ -2376,20 +2376,26 @@ QPDF::getCompressibleObjGens()
     QPDFObjectHandle encryption_dict = m->trailer.getKey("/Encrypt");
     QPDFObjGen encryption_dict_og = encryption_dict.getObjGen();
 
-    QPDFObjGen::set visited;
+    const size_t max_obj = getObjectCount();
+    std::vector<bool> visited(max_obj, false);
     std::vector<QPDFObjectHandle> queue;
     queue.reserve(512);
     queue.push_back(m->trailer);
     std::vector<QPDFObjGen> result;
     while (!queue.empty()) {
-        QPDFObjectHandle obj = queue.back();
+        auto obj = queue.back();
         queue.pop_back();
         if (obj.isIndirect()) {
             QPDFObjGen og = obj.getObjGen();
-            if (!visited.add(og)) {
+            const size_t id = toS(og.getObj() - 1);
+            if (id >= max_obj)
+                throw std::runtime_error(
+                    "unexpected object id encountered in getCompressibleObjGens");
+            if (visited[id]) {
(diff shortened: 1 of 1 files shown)


jberkenbilt:
X Y n and X Y n+k are not allowed to both be defined. If X Y n+k exists, then X Y n is automatically null. Coincidentally, I was just rereading this in the spec yesterday since I'm thinking about adding better support for incremental updates. As long as this change is not going to mess that up, it should be okay based on your description. I haven't looked at the change yet but will scrutinize it. I may add some tests for generation anyway -- I haven't thought about it since the very beginning, and I'm not 100% sure it's right (though I have no reason to believe it isn't). I know QPDFWriter assumes generation = 0, and I will have to address that. Or at least it did. Maybe I fixed it already. I can't remember. Weird to think I've had qpdf for almost half my life.

m-holger:
> X Y n and X Y n+k are not allowed to both be defined. If X Y n+k exists, then X Y n is automatically null.

I realized that. I guess there are three questions

- is it legal for an (effectively dangling) reference to X n R to exist (in a reachable object) , if the current generation is n + k ?

- how likely is it to happen in practice, and if so, how likely is it that X n R is scanned before X n+k R?

- if it happens, is it acceptable that qpdf treats the X n+k R as non-compressible?

If we need to deal with this case, we would need to either ensure that dangling references are not included in the visiting set, or are only included if the next entry in the object table has a different object id.

jberkenbilt:
Ugh, qpdf is broken with handling of generations because of a bug that has been there since the very beginning.

jberkenbilt:
qpdf allows two different generations to exist together. I can't believe I made that mistake 20 years ago and it's been sitting there undetected all this time.

jberkenbilt:
Probably this bug has been hiding since qpdf correctly handles PDFs that delete and reuse objects per the spec. It's just that it would not handle certain types of errors that viewers handle properly; as such, people are not likely to create files like this, and there's probably no reason for the problem to have ever occurred. To be specific, in qpdf, if you create, say 4 0 obj, then delete it, then create 4 1 obj, qpdf will still resolve 4 0 R, which should now be considered null. I will have to see what the implications of this are and how hard it is to fix. Anyway, it does seem that this bug was actually in qpdf 2.0, the very first open source release of qpdf.

jberkenbilt:
I made my Ugh comment before I read your comment. Objects with generation > 0 are never compressible.

jberkenbilt:
> I realized that. I guess there are three questions
> 
>     * is it legal for an (effectively dangling) reference to X n R to exist (in a reachable object) , if the current generation is n + k ?

Yes. This is normal.

>     * how likely is it to happen in practice, and if so, how likely is it that X n R is scanned before X n+k R?

This is the normal case. What order it is scanned in will depend on how the file is referenced. When an object is deleted by incremental update, all the places where it was previously referenced will contain dangling references. Unless the incremental update replaces all the objects that referenced the deleted object, then the dangling references will stay around and be null.
 
>     * if it happens, is it acceptable that qpdf treats the X n+k R as non-compressible?

qpdf must treat X n+k R as non-compressible. There is no provision in the PDF spec for objects with generation != 0 to be compressed.

> If we need to deal with this case, we would need to either ensure that dangling references are not included in the visiting set, or are only included if the next entry in the object table has a different object id.

qpdf is broken in some cases here. I added some expected fail tests to illustrate this. I will scrutinize this change to make sure that it's okay with my refreshed understanding of generations including qpdf's current broken behavior regarding them.

libqpdf/QPDF.cc
```diff
@@ -2376,19 +2376,26 @@ QPDF::getCompressibleObjGens()
...
-        QPDFObjectHandle obj = queue.front();
-        queue.pop_front();
+        auto obj = queue.back();
+        queue.pop_back();
         if (obj.isIndirect()) {
             QPDFObjGen og = obj.getObjGen();
-            if (!visited.add(og)) {
+            const size_t id = toS(og.getObj() - 1);
+            if (id >= max_obj)
+                throw std::runtime_error(
```
jberkenbilt, line 2392: It is not correct for this to be a `std::runtime_error`. If this can only happen because of a programming error, it should be a `std::logic_error`. If it could happen because of something about the input file, which I doubt, unless `getObjectCount` is broken or there is concurrent access, which is not allowed, then it should be handled more gracefully. By the time this is called, we're already in QPDFWriter.

jberkenbilt (changes requested):

jberkenbilt:
Let me amend my previous comments. I'm thinking ahead to incremental update. Right now, `QPDFWriter` _never_ writes a generation > 1, so this is free to return ObjGen with generation > 1. They won't have generation > 1 once they are written because `QPDFWriter` will renumber them.

jberkenbilt:
I think we can't take this as it is. However, in the vast majority of cases, particularly where object streams are in use, there are no objects with generation > 0. I think a small tweak that mantains a set for things with generation > 0 would fix this. Do you want to try it? If not, I can.

jberkenbilt:
I know it's late for you already.

jberkenbilt:
I already made a fix. I will push up to a branch and add a test. If it works, I'll take it, and that will grab this change with it.

libqpdf/QPDF.cc
m-holger: Agree this should be a logic error.

m-holger:
>  Objects with generation > 0 are never compressible.

I knew that, (but failed to think of it). In any case, such an object mar refer to compressible objects that are not otherwise reachable.

jberkenbilt:
I am creating a meaningful test for this.

jberkenbilt:
I'll have something up in a few minutes.

jberkenbilt:
@m-holger Please see #1110. The new test files were completely hand-generated since I don't have qpdf support for updating xref tables with incremental updates at this time. (But I have emacs lisp functions to help me generate xref tables and migrate around PDF files.)

jberkenbilt:
> I knew that, (but failed to think of it). In any case, such an object mar refer to compressible objects that are not otherwise reachable.

Right -- I made sure to do that in my test file.

jberkenbilt:
> Right -- I made sure to do that in my test file.

well, almost...fixing my test to actually do that.

PR #1110, opened 2024-01-07
Tune compressible

jberkenbilt:
@m-holger Here's an enhancement to your PR #1109 that also improves upon the generation > 0 test for object streams and adds some expected fail tests for qpdf's broken generation handling.


jberkenbilt:
@m-holger would love your thoughts on the code changes; still tweaking the test files.

jberkenbilt:
I'm going to hold 11.8.0 until I have a chance to think about how to fix this generation bug.

jberkenbilt:
Even with my making this very slightly worse, it's still a big performance change. @m-holger I'm a little worried that you're going to make qpdf so fast that it is going to run in negative time. Then it will be faster to run qpdf than not to run it, and somce I spend most of my time not running qpdf, so I'm not sure how that's going to affect things. (But seriously, this is a great change.)

jberkenbilt:
> Even with my making this very slightly worse, it's still a big performance change.

Actually, it seems inconsistent. Most things got faster.  A few things got a little slower. I think object stream preservation and generation are faster and a few other things are within the margin of error. It appears to be a net positive change. I'll go ahead and merge this once through CI, but since I'm holding 11.8, there's time to tweak if you feel like you want to. I think the code as modified should better handle the dangling reference case. Hopefully when I fix the bug, a bunch of xfail cases will start passing.

jberkenbilt:
I think the fix is going to be something vaguely like removing any older generations from the xref table and cache when inserting an object with generation > 0. A better fix might be to make the xref table and cache indexed only by object ID but to handle generations in a sensible way so that accessing an object by the wrong generation fails. Anyway, I'll have to give it some thought. I bet changing the index of xref_table and obj_cache to int would improve performance too.

jberkenbilt:
@m-holger Even though this is merged (it's better than what was there before), feel free to suggest tweaks.

PR #1111, opened 2024-01-07
Bug fix: treat old generations of reused objects as null

jberkenbilt:
This is a partial fix to the problem of deleted/reused objects. It handles the case when they're reused, but not the case when they're deleted and not reused. I'm going to create a test case for that and update this.


jberkenbilt:
cc @m-holger

jberkenbilt:
I think it's right now.

jberkenbilt:
@m-holger I this fixes the generation bug. I will await your comments on this and #1110 before releasing. In any case, I won't release today -- I need to let my head people have at least one night to process this.

jberkenbilt:
(@m-holger last message from me for now -- just letting you know that you can look at this tomorrow if you want -- I know it's late for you. If you don't have time to look at it, I will proceed after hearing that.)

PR #1112, opened 2024-01-08
Tweaks to QPDF::read_xref, removeObject and getCompressibleObjGens

m-holger:


jberkenbilt:
I will look at this in a couple of hours. It occurred to me that, with the fix in #1111, while there may be dangling references to old generations, there will never be more than non-null value for any object id. As such, my second set is not actually needed and your original submission is probably correct (except runtime -> logic error).

libqpdf/QPDF.cc
```diff
@@ -1987,15 +1987,15 @@ QPDF::replaceObject(QPDFObjGen const& og, QPDFObjectHandle oh)
 }
 
 void
-QPDF::removeObject(QPDFObjGen const& og)
+QPDF::removeObject(QPDFObjGen og)
 {
-    auto null = QPDFObjectHandle::newNull();
     m->xref_table.erase(og);
-    if (isCached(og)) {
+    if (auto cached = m->obj_cache.find(og); cached != m->obj_cache.end()) {
```
jberkenbilt, line 1993: Interesting...I've never seen this syntax before.

jberkenbilt (approved):
Thanks for the fixes. I'm not as comfortable with the new order.

jberkenbilt:
Okay, I looked at it now. But I still think we can improve #1110 by removing the extra set. If you don't beat me to it, I'll try it later. It is still actually possible for someone to insert multiple objects with different generations, but we could tweak things to make that impossible. I started at it, but it was too big of a change to make without thinking it through.

m-holger:
> I still think we can improve https://github.com/qpdf/qpdf/pull/1110 by removing the extra set

I agree. I am trying it but having problems at the moment.

There was a CI failure building AppImage with a very unhelpfull error message. Any thoughts?

PR #1113, opened 2024-01-08
Remove unneeded workaround

jberkenbilt:
After the last fix, a previous workaround is no longer needed.
@m-holger thoughts? I haven't thought it through carefully enough. I may not be handling dangling references correctly, though it passes the test suite. e.g. if a null dangling comes first, will it block traversal through the live one? I have to run for an appointment. Maybe we need a null check before adding to visited -- out of time to test. I'll be back in about 2 hours.

### DIFF
--- a/libqpdf/QPDF.cc
+++ b/libqpdf/QPDF.cc
@@ -2398,7 +2398,6 @@ QPDF::getCompressibleObjGens()
 
     const size_t max_obj = getObjectCount();
     std::vector<bool> visited(max_obj, false);
-    QPDFObjGen::set visited_gen; // for objects with generation > 0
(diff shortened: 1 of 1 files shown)


m-holger:
I have done the same thing in #1114.

However, I've added a check to make sure we are handling the current object. I think there are still ways in which obj_table can have 2 non-null object and I think it is safer (at least for the moment) to have that extra check.

Still looking into the CI failure

m-holger:
I found a fix for the CI failure.

jberkenbilt:
closing in favor of #1114

jberkenbilt:
> I found a fix for the CI failure.

#1115 unless you have a better way

PR #1114, opened 2024-01-08
Tweak QPDF::getCompressibleObjGens

m-holger:


jberkenbilt:
Will wait for your fix to the CI failure. I was thinking it was probably a transient issue and was going to ignore it.

jberkenbilt:
I see what's causing the CI failure. I can push a fix.

m-holger:
How did you do the diagnosis of the CI failure?

Closed in favour of #1115

jberkenbilt:
I guess C++ isn't rust. Or at least not the gcc version in Ubuntu 18.04 which is used to build the AppImage for maximum compatibility.

m-holger:
Thanks.

As an aside, for gcc-12 /clang-15, unpacking of pairs in for loops does not have a detectable overhead (unlike the versions I was using 18 month ago) .

For my gcog tweak, std::deque has a slightly better performance than std::vector, but it looks like Microsoft doesn't implement that part of the standard library.

jberkenbilt:
> As an aside, for gcc-12 /clang-15, unpacking of pairs in for loops does not have a detectable overhead (unlike the versions I was using 18 month ago) .

I didn't even know that syntax until I saw it in your PR.

> For my gcog tweak, std::deque has a slightly better performance than std::vector, but it looks like Microsoft doesn't implement that part of the standard library.

The thought occurred to me to use a regular array, but really, I think a lot of these differences are going to shift around based on the compiler and even the version of the compiler, so I wouldn't want to go overboard trying to optimize this further by tweaking the data type. Using an indexed data structure by object ID rather than a binary tree is a good change for this, but I think overly tweaking the data type is not worth the effort.

Issue #1048, opened 2023-10-10
Add facility to copy/move annotations from one page to another page


m-holger:
Use cases:

Edit a pdf file in a qpdf based application while simultaneously annotating (a copy of) it using an external tool. Import the newly created annotations.

Copy or split a page. Selectively move some of the annotations to the new page, for example, based on the location of the annotation. 

See for example https://github.com/pdfarranger/pdfarranger/issues/940#issuecomment-1755119962

jberkenbilt:
I think qpdf can already do this. See `QPDFPageObjectHelper::copyAnnotations`. I'm closing for now, but feel free to reopen if you think something beyond this is needed. I don't see this as something that would make sense to add to the CLI. pdfarranger could make use of the annotation copying functionality from libqpdf, assuming it is exposed to pikepdf.

elmimmo:
I, for one, could make use of this feature in the CLI.

InDesign does not provide a way to export [editorial notes](https://helpx.adobe.com/indesign/using/adding-editorial-notes-indesign.html) to PDF as annotations, but InCopy does, albeit without the rich PDF settings that InDesign allows for, so I export with both and then import the annotations from the PDF of the latter into that of the former with Acrobat. That, each time I need them, which ideally is always, so gets tedious fast.

I can currently do this with qpdf’s CLI by overlaying one PDF onto the other, but I must have made sure that the PDF exported with InCopy had all layers hidden (in which case it will still export notes as annotations) or else, chaos, so it's prone to user error and therefore a fragile workflow.

PR #1116, opened 2024-01-09
Add copy annotation test

m-holger:
I am not sure #1048 is fully implemented, at three different levels:

- I am not sure the code to copy an individual annotation is complete, e.g. the /P key is not fixed on copying. I think it would be worthwhile to check the existing code against the PDF spec.

- There is no high-level code (that I am aware of) to copy all annotations (or all annotations from a region) of one page to another page.

- There is no way to specify in the qpdf CLI  when selecting pages whether annotations should be included or not.

### DIFF
--- a/qpdf/qtest/copy-annotations.test
+++ b/qpdf/qtest/copy-annotations.test
@@ -14,7 +14,7 @@ cleanup();
 
 my $td = new TestDriver('copy-annotations');
 
-my $n_tests = 46;
+my $n_tests = 48;
 
 $td->runtest("complex copy annotations",
              {$td->COMMAND =>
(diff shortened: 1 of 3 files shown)


jberkenbilt:
> * I am not sure the code to copy an individual annotation is complete, e.g. the /P key is not fixed on copying. I think it would be worthwhile to check the existing code against the PDF spec.

This would be in scope for the pages work.
 
>     * There is no high-level code (that I am aware of) to copy all annotations (or all annotations from a region) of one page to another page.

There is almost a single way to do that, but it requires more than one call. `QPDFAssembler` will help with this.

>     * There is no way to specify in the qpdf CLI  when selecting pages whether annotations should be included or not.

What is a use case for not including annotations? But anyway, if we wanted to exclude annotations, it would be possible to add a flag for it. I'm just not sure why I would add something like that. There is flatten annotations.

PR #1119, opened 2024-01-12
Fix null pointer issue on array copy

jberkenbilt:
@m-holger FYI


m-holger:
Sorry about this

jberkenbilt:
> Sorry about this

No worries. I could have caught it, and the test suite should have covered this case. Maybe I need to do something about coverage analysis beyond QTC, which is good, but perhaps insufficient. Then we could check easily to see whether some code we are changing is covered.

m-holger:
Perhaps it would also be worth considering freezing the code for a couple of days before releasing.

jberkenbilt:
Yeah. I usually give it overnight at least. With more complex changes, it's worth a little longer. Though I don't think very many people use unreleased versions. As you know, I watch the code freeze cycle of debian and Ubuntu and am extra careful with things that are going into LTS releases. I'm not sure what version will go into 24.04, but I doubt it will have the pages work in it. We'll see how much more time I can get for this. I think I have at least another week or two before I get pulled back into my normal work.

PR #1120, opened 2024-01-13
Refactor QPDF object cache and xref table

m-holger:
This PR contains some preparatory work for ensuring that the object cache only contains null entries for deleted objects.

As the approach taken may be controversial some early feedback would be useful.

(This is a very early draft with obvious omissions, eg deleting constructors, controlling access, etc)

### DIFF
--- a/include/qpdf/QPDF.hh
+++ b/include/qpdf/QPDF.hh
@@ -985,14 +985,8 @@ class QPDF
     QPDFObjectHandle reserveObjectIfNotExists(QPDFObjGen const& og);
     QPDFObjectHandle reserveStream(QPDFObjGen const& og);
     QPDFObjGen nextObjGen();
-    QPDFObjectHandle newIndirect(QPDFObjGen const&, std::shared_ptr<QPDFObject> const&);
-    QPDFObjectHandle makeIndirectFromQPDFObject(std::shared_ptr<QPDFObject> const& obj);
+    QPDFObjectHandle makeIndirectFromQPDFObject(std::shared_ptr<QPDFObject>&& obj);
     void removeObject(QPDFObjGen og);
-    void updateCache(
-        QPDFObjGen const& og,
-        std::shared_ptr<QPDFObject> const& object,
-        qpdf_offset_t end_before_space,
-        qpdf_offset_t end_after_space);
     static QPDFExc damagedPDF(
         std::shared_ptr<InputSource> const& input,
         std::string const& object,
--- a/include/qpdf/QPDF.hh
+++ b/include/qpdf/QPDF.hh
@@ -839,18 +839,7 @@ class QPDF
     };
 
     // JobSetter class is restricted to QPDFJob.
-    class JobSetter
-    {
-        friend class QPDFJob;
-
-      private:
-        // Enable enhanced warnings for pdf file checking.
-        static void
-        setCheckMode(QPDF& qpdf, bool val)
-        {
-            qpdf.m->check_mode = val;
-        }
-    };
+    class JobSetter;
 
     // For testing only -- do not add to DLL
     static bool test_json_validators();
@@ -865,28 +854,7 @@ class QPDF
 
     static std::string const qpdf_version;
 
-    class ObjCache
-    {
-      public:
-        ObjCache() :
-            end_before_space(0),
-            end_after_space(0)
-        {
-        }
-        ObjCache(
-            std::shared_ptr<QPDFObject> object,
-            qpdf_offset_t end_before_space,
-            qpdf_offset_t end_after_space) :
-            object(object),
-            end_before_space(end_before_space),
-            end_after_space(end_after_space)
-        {
-        }
-
-        std::shared_ptr<QPDFObject> object;
-        qpdf_offset_t end_before_space;
-        qpdf_offset_t end_after_space;
-    };
+    class ObjTable;
 
     class ObjCopier
     {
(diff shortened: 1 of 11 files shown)


jberkenbilt:
Just FYI, I am not going to distract myself from pages work until I get it into a sustaining place.

jberkenbilt:
You'll have to describe verbally what you're trying to achieve and why. I don't want to spend time trying to reverse engineer your thoughts from the code. As you know, I'm not inclined to get distracted by architectural changes while I'm doing pages work, but I want to start looking at better support for incremental update as a step on the path toward digital signature support of some type, and I feel like what you're doing here may be relevant to that.

m-holger:
I am only looking for feedback on two points at the moment:

- are you ok with moving QPDF::Members (and other inner classes that expose Members in inline methods) into a private header file (which I have named QPDF_private), and

- are you ok with encapsulating the object cache and xref table into their own classes and restricting access to their underlying containers to those classes?

The immediate driver for this work is the multiple generations for a single object id problem. The one way to ensure that there are never multiple generations for the same id in the object cache and xref table is to index them by id only rather than by objgen, and making the gen part of the value.

However, changing the xref table and object cache is made difficult by the fact that they are accessed directly in numerous places in QPDF.cc and QPDF_xxx.cc as well as (through copies) by QPDFWriter. The current work tries to address that difficulty by eliminating the direct access and replacing it by access through a limited number of specific methods. Those methods can then be modified to implement the desired changes, as well as any future changes that will be required.

There are two complications that affect this work:
- access to the object cache is performance critical, and
- there is a close connection between the object cache and the private QPDFObject  and QPDFValue classes.

To make the encapsulation viable a number of methods that include access to private classes need to be in-lined and therefore 
can not live in a public header file.

In addition to dealing with the multiple generations issue, encapsulating access and/or changing the key from objgen to id only should provide a number of benefits and future opportunities:, e.g.

- Implementation of a flag for unchanged objects should be easier. (This could be extended to other flags, eg validated. Clearly an object that has been validated does not need to be re-validated unless it has been changed).

- The underlying container type for both tables could be changed from map to vector. This should reduce runtime and, except for files with an extremely high proportion of deleted objects, footprint.

- More methods could be cleanly exposed to the rest of qpdf without making it part of the public API. A public inline method implemented in a private header file can't be called from outside qpdf.

- Direct access to QPDFObject and QPDFValue can be moved to those inner classes that really need it, reducing or eliminating such access by QPDF itself.

jberkenbilt:
What you're describing sounds fine in principal, and these seem like good, worthwhile changes. That said, I really want to focus on pages work while I still have time to devote to it, and I don't want to distract myself on this issue right now, so there is some risk that if you do it, it will sit for a long time and that you will deal with merge conflicts. This may be on the minor side because I don't think my pages work will touch the QPDF object itself that heavily.

Also, I am thinking about incremental updates. A good part of that work will be modifying QPDFWriter, but my current thinking is that QPDF will need to somehow represent incremental updates. What I'm thinking is having QPDF objects be layerable in very much the same way PDF files are. This would not always be on. Most of the time, QPDF would be have as it does today, but I would want the option of having an incremental update QPDF. My thought is that it would have a "previous" that it would resolve objects that were unknown with some semantics to allow an incremental update to create a new generation without deleting the old one (an error, but it happens). Having the object cache and xref table be based on object ID only would be an important aspect of this, but obviously there has to be a way to make sure you only get a hit when the generation is right. Consider:

* Original creates object 8 0
* First incremental update deletes 8 setting the next gen to 1
* Second incremental update defines 8 1

We'd have three QPDF objects that are stacked in some way with bidirectional knowledge. An attempt to resolve 8 0 R from even the original QPDF would have to result in `null`, but the `8 0` object definition would have to remain in that file so that, when it is written incrementally, it will still be there. This will require a `QPDFWriter` mode that preserves object IDs (a feature that has been requested anyway) or that renumbers consistently and has generation awareness. `QPDFWriter` currently never writes a generation > 0.

*PLEASE* don't go off and try to implement what I just described. I am not prepared to have this aspect of qpdf evolve through someone else's work. That will cause me to lose track even further of the internals, and I already feel like it's slipping through my fingers with all your work. I'm sharing this with you in hopes that, if you work on refactoring the xref table and object cache, you do so in a way that at least doesn't make what I've described harder than it already is. When I get to this, I will welcome your participation and support, but I don't want you to run off and build it on your own. Going through the process of designing and building a solution is an integral part of my maintaining an accurate mental model, and I need the insights I gain by doing it wrong and fixing it over multiple increments to optimize how I create future feature enhancements.

jberkenbilt:
> I did not spot a requirement to write an increment update when deleting an object before it can be reused. A logical conclusion of such a requirement would seem to be that an application that updates a pdf file (not incrementally) must always renumber objects.

I don't understand this comment.

> I am not sure I understand this. My expectation was that to support incremental updates, QPDFWriter would copy the original file and then write the updates. For this it needs to know what objects are used in the original file but would not need to know how to recreate the objects. What am I missing?

The ability to look at incrementally updated files in QDF mode or to perform other updates for inspection purposes. That would invalidate signatures, etc., but is still a useful case that people have asked for.

jberkenbilt:
Only one generation of an object is allowed to exist. In practice, if object `a b` exists, then any `a c` where `c < b` should be treated as null regardless of whether there are any explicit deletes. As far as I recall from previous experimentation, this is how all readers work. Adding a deleted object to the xref table allows you to delete the object without immediately reusing it and stores the next generation that should be used so that reusing the object doesn't accidentally resurrect any dangling references to the old generation. I probably misspoke and it's probably valid to reuse with deleting, but I haven't reread that part of the spec in a while. It doesn't matter though. In practice, the correct behavior is to treat the highest generation as valid and anything as null, and this definitely means obj_cache and xref_table should be indexed by object id only but have awareness of generation. I was on the verge of making that change myself and it probably would be the next thing I would do after pages unless you do it first. I would want to create some manual examples of things like adding a lower generation in an incremental update to see what various readers do, but I'm reasonably certain that is the highest generation that wins.

m-holger:
I probably will beat you to it.

m-holger:
Closed as this is very much out of date

PR #1121, opened 2024-01-14
Handle default-constructed JSON objects

m-holger:
Incomplete.

Presumably this is the expected behavior.

### DIFF
--- a/libqpdf/JSON.cc
+++ b/libqpdf/JSON.cc
@@ -191,7 +191,7 @@ JSON::JSON_blob::write(Pipeline* p, size_t) const
 void
 JSON::write(Pipeline* p, size_t depth) const
 {
-    if (nullptr == m->value) {
+    if (!m) {
         *p << "null";
     } else {
         m->value->write(p, depth);
@@ -201,6 +201,9 @@ JSON::write(Pipeline* p, size_t depth) const
 std::string
 JSON::unparse() const
 {
+    if (!m) {
+        return "null";
+    }
     std::string s;
     Pl_String p("unparse", nullptr, s);
     write(&p, 0);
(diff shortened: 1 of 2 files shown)


jberkenbilt:
I guess you're going to walk behind me with a mop and keep me humble while I code the pages epic. I used to think of myself as an excessively careful programmer, but I suppose I'm not as careful as I thought. Thanks for the fixes. Obviously I'll wait until this is ready and passing CI.

m-holger:
There is some inconsistency on how calling inappropriate methods is treated, e.g calling addDictionaryMember on a non-dictionary is a runtime error, while calling checkDictionaryKeySeen is a logic error. Given that both could occur as the result of incorrect input a runtime error seems more appropriate.

Also, given that JSON::write treats objects with uninitialized m->value as null objects it probably would make sense to treat default constructed objects the same.

m-holger:
On reflection, given the purpose of checkDictionaryKeySeen, it is appropriate that it throws a logic error.

Tests for !m->val are redundant since all non-default JSON objects are created by factories that ensure this is impossible.

jberkenbilt:
Is there an actual problem here? If not, I don't want to get distracted on the JSON implementation.

m-holger:
```
std::map<std::string, JSON> a_map;
a_map["no_key"].isNull();
```
de-references a nullptr, which would be bad. On the other hand, I don't think there is a problem for qpdf itself, and it is probably unlikely that users of the library would use JSON in a way that they would encounter the problem. So probably not urgent to fix this.

jberkenbilt:
Right, qpdf's JSON is not intended to be a general-purpose JSON API. There are better choices. The JSON.hh header states this explicitly. I have no interest in making improvements to the JSON classes that are not needed by qpdf.

m-holger:
I agree that there is no point supporting JSON as a general purpose JSON library. At the same time, JSON methods called with valid parameters should not cause segmentation faults. To my mind that calls for the JSON default constructor to be deprecated and removed for qpdf 12, or at least be made private or else made to return a JSON null object.

On a different JSON matter, the duplicate key check costs 12.5% of the runtime of the json-input test with the pdf-spec file, with a low but definite impact on other tests (eg  >1% on the no parameters test with the abuild-manual file). Given that the RFC permits duplicate keys, and the ECMA-262 specifies for conforming implementations of  JSON.parse:

> NOTE 2
In the case where there are duplicate name Strings within an object, lexically preceding
values for the same key shall be overwritten.

and that the qpdf policy is to be strict with outputs but lenient with inputs,
is it necessary to detect duplicate keys?

jberkenbilt (approved):
I'm okay with this change. I don't have a strong opinion about detecting duplicate keys. I'm okay with getting rid of it. I guess you can do that in a separate PR.

PR #1122, opened 2024-01-14
Tweak Handlers::beginUnderOverlay and beginPages

m-holger:
No tests for getDictItems pending feedback.

### DIFF
--- a/libqpdf/JSON.cc
+++ b/libqpdf/JSON.cc
@@ -410,6 +410,17 @@ JSON::isNull() const
     return m->value->type_code == vt_null;
 }
 
+JSON
+JSON::getDictItem(std::string const& key) const
+{
+    if (auto v = dynamic_cast<JSON_dictionary const*>(m->value.get())) {
+        if (auto it = v->members.find(key); it != v->members.end()) {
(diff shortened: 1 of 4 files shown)


libqpdf/JSON.cc
```diff
@@ -410,6 +410,16 @@ JSON::isNull() const
     return m->value->type_code == vt_null;
 }
 
+JSON
+JSON::getDictItem(std::string const& key) const
+{
+    if (auto v = dynamic_cast<JSON_dictionary const*>(m->value.get())) {
+        if (auto it = v->members.find(key); it != v->members.end())
```
jberkenbilt, line 417: Please put braces around the code in the condition.

jberkenbilt (changes requested):
Thanks. I should have done it this way. I added the new code and almost did, but for some reason, I didn't.

libqpdf/JSON.cc
m-holger: Done. Also a couple of tests added.

jberkenbilt (approved):
Thanks. Will merge in spite of the extraneous blank lines.

PR #1124, opened 2024-01-16
Tweak QPDFJob::handleUnderOverlay

m-holger:


m-holger:
A side question: under what circumstances are multiple QPDFAcroFormDocumentHelpers for a single QPDF object required?

m-holger:
Re programming style: given that `some_map.count(some_key) == 0` is a work-around for the missing `!some_map.contains(some_key)`, should the preferred style be `!some_map.count(some_key)` until qpdf requires C++20?

jberkenbilt:
> A side question: under what circumstances are multiple QPDFAcroFormDocumentHelpers for a single QPDF object required?

As far as I know, you would never want to do this, but it's a loaded question -- I need context to know why you're asking and what problem you're trying to solve.

jberkenbilt:
> Re programming style: given that `some_map.count(some_key) == 0` is a work-around for the missing `!some_map.contains(some_key)`, should the preferred style be `!some_map.count(some_key)` until qpdf requires C++20?

Personally, I think the idea of non-boolean values that are true or false is a language design bug, which is why I generally explicitly compare with 0.

jberkenbilt:
I'm not going to look at this PR for the foreseeable future. QPDFJob setup, parsing, etc., is not performance-critical. I feel like this is probably premature optimization. If there is something not stylistic or related to performance in code that is performance-critical, let me know. I want to get functionality working before messing around with tweaks. It will be very disruptive to me to have code I am actively working on being changed while I'm still working in it.

jberkenbilt:
Sorry for my last comment. I've got a lot going and may be a bit irritable. I do welcome the input. We should find a rhythm. Maybe I can leave PRs up for your commentary before merging, treating you more as an equal collaborator on API and efficiency concerns.

jberkenbilt:
I'll try to give this a look before I dive into the next chunk. Keep in mind that a lot of this code is being reworked. I'm going to end up moving a lot of QPDFJob into QPDFAssembler by current design. If you'd like to be a more equal contributor in terms of API, efficiency, and general stylistic concerns, I will await your review on PRs. I think that rhythm will feel less disruptive to me than these after-the-fact PRs. I should have waited for your commentary but had been under the impression that you were not going to be available.

jberkenbilt:
I'm not going to take this one. I'm going to significantly rework this logic in QPDFAssembler.

m-holger:
> > A side question: under what circumstances are multiple QPDFAcroFormDocumentHelpers for a single QPDF object required?

> As far as I know, you would never want to do this, but it's a loaded question -- I need context 

The immediate context is that I noticed that QPDFAcroFormDocumentHelpers are cached in various places in QPDFJob and I was wondering whether it would  make sense to have a single m->afdh_cache?

The wider context is that I seem to recall that the helper caches significant amounts of state and that my impression at the time was that if multiple helpers existed at the same time that there was a danger actions by one helper might cause the cached state of the other to become out of date. If my recollection/impression is correct, this would raise the question whether the state should be cached by the QPDF object and all helpers should share this single state.

If I recall correctly, I had the same question about the outline helpers.

m-holger:
I am not suggesting any involvement by QPDF other than to provide a member eg m->afdh_cache for the helper to store its (shared) state, with the helper responsible for creating and managing it. This would make the creation of a new helper instance cheap and eliminate the need for application code, including QPDFJob, to cache helpers to avoid the cost of creating them. This should reduce overall code complexity.

If I recall correctly, of the object helpers maintains a reference to its document helper with the obvious risks if it should outlive the document helper. The solution of storing the state with the QPDF object would remove this risk.

m-holger:
@jberkenbilt can you reopen for the discussion of document helpers.

I think this is an area that warrants further consideration for the following reasons:

1. It is generally good to avoid stale references to QPDF objects.

2. Some of the helpers are expensive to create and there is no simple way to avoid the unnecessary creation of them. While they can be cached, this is error prone. To take an example - QPDFJob creates a cache of QPDFAcroFormDocumentHelpers for handling Over/Underlays. It appears to me that separate caches are created for each call to doUnderOverlayForPage and therefore caching is a pure overhead rather than an optimization.

3. It is very difficult to feel certain that document helpers clean up their resources properly. I certainly had concerns about this when I looked at some of them in the past (without investigating further), and #1177 certainly does nothing to reassure me. Personally I would find it reassuring if QPDF objects tracked their document helpers and ensured that their resources get properly cleaned up on the death of the QPDF object.

jberkenbilt:
@m-holger I don't like how document helpers (or object helpers) work. Are you suggesting I should reopen this PR? You can always create a discussion issue. I'm not sure about having QPDF keep track of them. It feels like leaking in the wrong direction of the abstraction. That said, I recognize your concerns about efficiency, etc. My original conception of these is that they would be short-lived objects, usually stack-allocated, that would just help to reduce boilerplate and reduce clutter of QPDFObjectHandle's interfaces. I think revisiting them may be useful in a broader context.

PR #1126, opened 2024-01-17
Fix incorrect handling of invalid negative object ids

m-holger:
Fix two errors introduced in #1109 and #1112. Since #1109, encountering the invalid indirect reference  -2147483648 n R produces an integer underflow which, if undetected, immediately trigger a logic error. Since #1112, object -1 0 R may be incorrectly identified as an earlier generation of itself and deleted, invalidating a live iterator.


libqpdf/QPDF.cc
```diff
@@ -709,10 +709,11 @@ QPDF::read_xref(qpdf_offset_t xref_offset)
 
     // Make sure we keep only the highest generation for any object.
     QPDFObjGen last_og{-1, 0};
-    for (auto const& og: m->xref_table) {
-        if (og.first.getObj() == last_og.getObj())
+    for (auto const& item: m->xref_table) {
```
jberkenbilt, line 712: You may recall that my pre-tweak code gathered things up in to_delete and did a separate pass. I have developed the habit of coding that way to just avoid the pattern of modifying something I'm iterating through. I know that many of the standard library containers have well-defined semantics for modifying while iterating, but since I can't keep them in my head (except for std::list), I avoid using them as avoiding them reduces the cognitive burden of someone reading and maintaining the code. This is not performance-critical code -- it only happens once. An extra scan through a list of objects with the non-highest generation I suspect wouldn't even register in any of the performance tests, and files with reused objects are rare. This is the kind of thing I mean when I talk about coding for clarity. I don't know if my original code would have still been wrong, but the live iterator thing wouldn't have happened.

How are we getting negative object IDs? This sounds like a more fundamental issue. Maybe QPDFParser needs to refuse to create negative object IDs. This fix is probably still good for defensive coding purposes.

jberkenbilt (approved):
This is better than what was there before, but I wonder how we're getting negative object IDs in the first place.

PR #1127, opened 2024-01-17
Tighten checks for invalid indirect references in QPDFParser

m-holger:
I have extended the practice of silently replacing 0 0 R with null to all invalid references. While I don't think that the detail of these repairs is of interest to most users, I am not sure that qpdf should report that it succeeded without errors in this case.


jberkenbilt (approved):
Thanks -- this looks like a deeper fix to the other issues.

PR #1128, opened 2024-01-17
Tighten checks for invalid indirect references  during xref reconstru…

m-holger:
…ction

I don't think having a specific warning here is useful - we already know that the file is damaged. I think a message here would just the flood of spurious messages for severely damaged files.

Do we need any specific test cases?


m-holger:
The method only inserts entries for uncompressed objects (including objects with negative object ids found during reconstruction).

jberkenbilt (approved):

PR #1132, opened 2024-01-21
In JSON::parse allow duplicate dictionary keys

m-holger:
If duplicate keys are encountered, overwrite earlier values with the latest value.

Should checkDictionaryKeySeen be deprecated? I can't really see it being used by users, if it is used they can easily replicate it and removing it saves 8 bytes for every JSON dictionary.


jberkenbilt:
I have to think through how this is going to interact with the JSON reactor, and I'm not going to do that until after the pages work settles.

libtests/qtest/json_parse.test
```diff
@@ -120,15 +120,15 @@ my @bad = (
...
+    # 40 "bs char in string"
     "decimal point after minus",# 41
     "e after minus",            # 42
     "missing digit after e",    # 43
     "missing digit after e+/-", # 44
     "tab char in string",       # 45
     "cr char in string",        # 46
     "lf char in string",        # 47
-    "bs char in string",        # 48
+    "bs char in string",        # 40
```
jberkenbilt, line 131: These have to be in order. Otherwise, the labels in the test output won't match up with the actual test file.

jberkenbilt:
I'm not going to accept this change. The problem of how JSON handles duplicated keys is not important. Not detecting them potentially causes the json reactor to be called multiple times for the same key. Duplicated keys in QPDF JSON or QPDFJob JSON are probably mistakes, and I'd rather have them be detected.

m-holger:
Just to make sure, you do appreciate that this check costs 12.5% of the runtime of the JSON input test with the pdf-spec file (clang-15) ?

jberkenbilt:
I didn't know it was 12.5%, but is that actually a problem right now? I don't think it's appropriate for the reactor to be called more than once on a given element.

jberkenbilt:
I'll think about it later.

m-holger:
> is that actually a problem 

I think so

>  right now?

no

>  I don't think it's appropriate for the reactor to be called more than once on a given element.

This may certainly be true for some reactors, and if so, this would need to be catered for. On the other hand, I don't think it matters, for example, when parsing a JSON representation of PDF objects, where the standard JSON rules and QPDFParser treat duplicate keys the same way.

If there are cases where reactors require a guarantee that a given key is seen at most once than this raises three questions:

- should the necessary checks be optional so that the cost only occurs when the checks are necessary? There are two distinct reasons why the checks may be unnecessary:
  - the reactor does not require a uniqueness guarantee; and
  - the production of the JSON and the consumption by the reactor are part of a single system and uniqueness can be guaranteed at the production side - for a example, if the JSON dictionary is produced by serializing a map data structure duplicate keys are impossible.  

- what is the cheapest way to provide the guarantee?

- can  checkDictionaryKeySeen deprecated? Arguments in favor are that 
  - there is no obvious use for it in user code; 
  - if there is a use, it is easily and more cheaply implemented in user code; 
  - providing support for checkDictionaryKeySeen has a cost for all programs, whether they use it or not; and
  - it is not required in order for qpdf to provide a uniqueness guarantee when required.

jberkenbilt:
I don't think deprecating this is sensible. If we don't need it, we should simply remove it and break API to force people to deal with it. I think it's likely that there is no code outside the library that calls it. What I'd be willing to consider is outright removal of duplicate key detection and the function that checks for duplicate keys along with a change to the specification that, if duplicate keys appear in the JSON, the reactor will be called for all of them, and the last one accepted will win. That means that someone who wanted to implement the old behavior could in theory track duplicate keys in the reactor, but I don't even see a reason why we would do that.

If you agree, feel free to adjust this PR to fix the test order issue and to drop `parsed_ keys` and `checkDictionaryKeysSeen` entirely. We can tag this with `ABI` and accept it when we open the qpdf 12 window.

libtests/qtest/json_parse.test
m-holger: I realized that, but managed to comment the wrong entry out. (I think its worth keeping a comment in the original place as the test logically belongs there)

m-holger:
Updated as discussed. This is now ABI breaking.

libtests/qtest/json_parse.test
```diff
@@ -120,15 +120,15 @@ my @bad = (
...
     "decimal point after minus",# 41
     "e after minus",            # 42
     "missing digit after e",    # 43
     "missing digit after e+/-", # 44
     "tab char in string",       # 45
     "cr char in string",        # 46
     "lf char in string",        # 47
-    "bs char in string",        # 48
+    #"bs char in string",       # 40
```
jberkenbilt, line 131: I think you forgot to delete this comment.

m-holger: I left it deliberately as a comment because logically the test belongs there.

jberkenbilt: Please do this instead. It's ugly that I set it up with the tests being in a specific order, but I don't want random renames or comments that would be in the way if we added more tests.

https://github.com/m-holger/qpdf/pull/7

jberkenbilt (changes requested):

jberkenbilt (approved):
I'm good with this as it is now and at this time.

PR #1133, opened 2024-01-21
Add additional test case for fuzz CI tests

m-holger:


jberkenbilt:
The test is not reproducible on oss-fuzz and will be closed automatically on Feburary 3 if it doesn't recur. I think this is a fluke. We can ignore it. It happens sometimes.

PR #1135, opened 2024-01-27
Refactor Pl_TIFFPredictor 

m-holger:
Use standard library instead of memcpy and raw pointers.

For the last 7 days crashes have been running at over 500 per day. Probably worth seeing whether this makes a difference

### DIFF
--- a/libqpdf/Pl_TIFFPredictor.cc
+++ b/libqpdf/Pl_TIFFPredictor.cc
@@ -7,6 +7,7 @@
 #include <climits>
 #include <stdexcept>
 #include <vector>
+#include <iostream>
 
 Pl_TIFFPredictor::Pl_TIFFPredictor(
     char const* identifier,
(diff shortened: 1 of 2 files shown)


m-holger:
Btw, is there a test recipe for performance testing Pl_TIFFPredictor?

m-holger:
Also btw, bits_functions feels overly general.

If n is the number of bits read/written, bits_function does not impose any constraints on n (other then maxima). As far as I can tell, in all cases where the functions are used we have n = 8 * i or n * i  = 8 for some integer i. Only supporting those cases would make the code much easier to reason about, and possibly might be faster.

jberkenbilt:
> Also btw, bits_functions feels overly general.
> 
> If n is the number of bits read/written, bits_function does not impose any constraints on n (other then maxima). As far as I can tell, in most cases where the functions are used we have n = 8 * i or n * i = 8 for some integer i. Only supporting those cases would make the code much easier to reason about, and possibly might be faster.

I took the bits code from earlier code I had written for that image library I've mentioned. That code needed more general bit support. Also, linearization hint tables use it. Maybe they need more functionality. It's been a long time. The bit code isn't heavily used in most cases, I believe, so I don't how much benefit we'd get from optimizing it. That code has been very stable for a long time. I'm not saying one way or the other about whether it can be simplified, only that it was used in a more general context and that I haven't looked at the code for a very long time.

jberkenbilt:
I've scrutinized this, and I can't see any problems with it. However, with your optimization of the 8 bits per sample case, the coverage has changed. I'm going to double check code coverage and modify the fuzzer so that I'm certain we are exercising other than 8 bits per sample in the fuzzer.

jberkenbilt (approved):
Approved with one change to the fuzzer I am making separately and pushing at the same time.

PR #1138, opened 2024-01-29
Allow "n:/pdf-syntax" JSON syntax for dictionary keys

m-holger:


jberkenbilt (commented):
Thanks. I guess I overlooked this case. It would be nice if we didn't have to duplicate the logic of deciding which way to go. I'd probably either create a QPDF_Name as call its getJSON or, if that's too inefficient, tweak QPDF_Name to add something that does that. Currently QPDF_Dctionary already calls QPDF_Name::normalizeName, and these are all internal APIs.

m-holger:
Create a QPDF_Name would be very expensive.

When I get some time I will push a PR with a replacement for getJSON which avoids actually creating JSON objects, saving 16% for the pdf-spec and  7% for abuild-manual on the JSON output test (on clang-15). Probably not worth dealing with the duplication until then.

I 've pushed a PR to deal with the pikepdf test failure.

jberkenbilt:
I feel like you're spending a lot of time on qpdf for someone who doesn't have any time to spend on qpdf right now. :-)

jberkenbilt (approved):

PR #1142, opened 2024-02-03
Fuzz json

jberkenbilt:


jberkenbilt:
@m-holger If you feel like looking at this, feel free, though please don't feel a need to prioritize this over other things. I would wait for your review, but I'd like to get this pushed out so the new fuzzer can start being used and potentially finding issues. I reworked the parser to slightly improve efficiency (though I didn't make a huge effort to optimize heavily) and to remove some weaknesses and complexity. The new implementation keeps state in a simpler way and is much less likely to having its internal state land in a bad place.

I'm merging now so oss-fuzz can start fuzzing with it. It wouldn't surprise me if we start seeing some crashes.

PR #1146, opened 2024-02-10
Refactor writing objects as JSON

m-holger:


libqpdf/QPDF_Name.cc
```diff
@@ -51,20 +54,82 @@ QPDF_Name::unparse()
     return normalizeName(this->name);
 }
 
+std::pair<bool, bool>
+QPDF_Name::analyze(const std::string& name)
```
jberkenbilt, line 58: `analyze` is too broad of a name for this function. It's doing too many things to not have comments explaining what it is doing and why. I'm not looking for blow-by-blow comments that duplicate the code, but some kind of explanation is warranted. It can't return a `std::pair` without explaining what the first and second values mean.

jberkenbilt (changes requested):
Please explain what you're trying to accomplish here, what the benefit is, and briefly what the key aspect of the implementation is. I can mostly tell from the code, but the pull request needs to explain it for future reference and to ensure that I'm not getting it wrong when I read it.

This introduces a lot of duplicated code, and some of it, especially QPDF_Array and QPDF_Dictionary, is complex. Is it really worth it? I appreciate the performance improvement, but I'm not completely convinced that it's worth even a significant performance improvement for writing json if the cost is so much duplication of logic and yet another case of where there are multiple ways to accomplish the same goal. If we are going to accept this, there would need to be good tests that would fail if getJSON and writeJSON are inconsistent with each other. I don't see those tests. Does this change coverage?

m-holger:
The purpose of this PR is to performance tune the writing of a JSON representation of a QPDFObjectHandle. On clang-15 this reduces the runtime of the json-output test by around 17.5% for the pdf-spec file and around 8% for the abuild-manual file.

The main optimizations are:

- Write the JSON representation on the fly without creating an intermediate JSON object. This avoids 3 memory allocations for most objects as well as a number of string copies and a number of function calls.

- Inline a number of small methods.

- Tune the writing of commas, new lines and indentations to reduce the number of function calls and avoid the creation of strings.  

- Tune checking of names for invalid utf-8 characters and characters that require escaping. Carry out the checks in a single pass to avoid the extra iteration over the string as well as in most cases a string copy. Carry out the utf-8 check by inspection only without calculating code points to reduces the calculations required.

m-holger:
> Is there a way to have getJSON and writeJSON share code

Assuming we add a QPDF_Stream::writeJSON, is there a need for getJSON? I cannot think of a reason to call getJSON other than to write the object as JSON.

jberkenbilt:
> Assuming we add a QPDF_Stream::writeJSON, is there a need for getJSON? I cannot think of a reason to call getJSON other than to write the object as JSON.

Well, the ability to get a JSON from a QPDFObjectHandle is part of the public API, but I agree it's not that likely that anyone would need it for another purpose. I was wondering if it might make sense to implement getJSON for the non-trivial cases as calling writeJSON and passing a pipeline that is sent to the JSON parser, perhaps by writing to a Pl_Buffer and parsing with a BufferInputSource. It will probably make getJSON a little worse than it is, but probably not that much worse. Implementing getJSON in terms of writeJSON will reduce or eliminate code duplication and will not negatively impact performance for the common case. Probably getJSON can exist in that form in QPDFObjectHandle and not require each individual value type to have its own getJSON. I think that could be done without breaking ABI. I'm saying all this without studying the code, so I might be missing something.

m-holger:
That should work.

m-holger:
I have implemented  getJSON in terms of writeJSON as suggested. To assess the impact, I ran the json-output test with --json=2 and --json=1 instead of --json-output. For --json=2 the performance improvement was similar to --json-output suggesting that for light usage the performance penalty is acceptable. 

For --json=1 I saw a 12% - 20% degradation, suggesting that for heavy usage there is a definite penalty. I have tweaked QPDFJob::doJSONObjects to use writeJSON. The performance is not quite as good as for JSON v2, but given that this is a legacy format I don't think it is worth spending time on tuning this further.

libqpdf/QPDF_Name.cc
```diff
@@ -51,20 +54,69 @@ QPDF_Name::unparse()
...
+            return {false, false};
+        }
+    }
+    return {tail == 0, !needs_escaping};
+}
+
+void
+QPDF_Name::writeJSON(int json_version, JSON::Writer& p)
 {
     if (json_version == 1) {
```
m-holger, line 109: Reminder to self: Add comment with a reminder to also update QPDF_Dictionary::writeJSON if this is updated.

(I tried to use a common method to use in both places, but haven't so far managed to do it without a significant performance penalty)

jberkenbilt:
> For --json=1 I saw a 12% - 20% degradation, suggesting that for heavy usage there is a definite penalty. I have tweaked QPDFJob::doJSONObjects to use writeJSON. The performance is not quite as good as for JSON v2, but given that this is a legacy format I don't think it is worth spending time on tuning this further.

I agree. It would be nice to to make it a lot worse, but I'm okay with some degradation on the legacy format in exchange for the combination of better code simplicity and performance gain on the current format.

jberkenbilt:
I like this a lot better. I'll try to do a careful review and get this merged today. If you don't get to the "reminder to self" comment above, I can add a comment, or you can do it in a separate PR. I'd like to get this baking for a few days so I can release qpdf 11.9.0 later in the week and get out in front of the Ubuntu feature freeze.

jberkenbilt (approved):
I think this looks right. I'm going to check code coverage before I merge.

jberkenbilt:
When #1152 merges, this will come along for the ride. #1152 includes this and adds some build options to make it easier to do coverage testing.

jberkenbilt:
I'm happy with how this turned out. Thanks for this contribution.

PR #1158, opened 2024-02-20
Add test for QPDFObjectHandle::isDirectNull

m-holger:


jberkenbilt:
Rather than merging these as they come, how about if you let me know when you're gone through and added the tests you're going to add? Since these are just tests, I don't have an issue with merging right before I prepare 11.9.0. If there are any non-obvious fixes, I might push it off a little.

Thank you very much for doing this.

PR #1159, opened 2024-02-21
Exclude test files from coverage checks

m-holger:


jberkenbilt:
I don't want to do this as explained in private email.

PR #1161, opened 2024-02-28
Tune QPDFWriter

m-holger:
The main changes are based on changing the indexing of containers of objects from objgens to object ids based on the fact that only one object with the same id can exist in a pdf file. This than allows the replacement of maps with vectors as well as the elimination of duplicate containers indexed by both objgens and object ids.

A number of vectors have been combined into two tables, obj indexed by original ids and new_obj indexed by  renumbered ids.

In addition a number of intermediate containers have been removed by combining processing loops, e.g. preserveObjectStreams works directly on a const& of the main xref table rather than the map of compressed objects produced by getObjectStreamData, and sets of compressible objects are now directly produced by getCompressibleObjGens.

There are also number of minor tweaks to reduce the processing inside loops.


jberkenbilt (approved):
This is too big and complex for me to spend the time to grok it fully, but the thrust of the changes look correct, and given that there were no test change, I'm happy to accept this as is. I read this through commit by commit as well as looking over it in its entirety, and I view it as an important step toward the single generation per object representation. I'm going to merge this, and if there are changes, we can address them in a future update.

PR #1162, opened 2024-02-28
Remove non-standard string_view<unsigned char> from QPDF_Name

zclifford:
string_view<unsigned char> leads to char_traits<unsigned char> which is not standard C++ (background in https://github.com/qpdf/qpdf/issues/1024).

This triggers compilation failures with certain C++20 compiler configurations.

To avoid this I moved the cast to the loop's range declaration. This should work (for 8-bit characters) despite switching from an explicit cast to an integral conversion, per the following C++17 standards text:

> For each value i of type unsigned char in the range 0 to 255 inclusive, there exists a value j of type char such that the result of an integral conversion from i to char is j, and the result of an integral conversion from j to unsigned char is i.

But if you want me to be more explicit and keep the reinterpret_cast at the cost of an extra line or two I'd be happy to.

https://github.com/qpdf/qpdf/issues/1024

### DIFF
--- a/libqpdf/QPDF_Name.cc
+++ b/libqpdf/QPDF_Name.cc
@@ -3,8 +3,6 @@
 #include <qpdf/JSON_writer.hh>
 #include <qpdf/QUtil.hh>
 
-#include <string_view>
-
 QPDF_Name::QPDF_Name(std::string const& name) :
     QPDFValue(::ot_name, "name"),
     name(name)
@@ -57,14 +55,12 @@ QPDF_Name::unparse()
 std::pair<bool, bool>
 QPDF_Name::analyzeJSONEncoding(const std::string& name)
 {
-    std::basic_string_view<unsigned char> view{
-        reinterpret_cast<const unsigned char*>(name.data()), name.size()};
-
     int tail = 0;       // Number of continuation characters expected.
     bool tail2 = false; // Potential overlong 3 octet utf-8.
     bool tail3 = false; // potential overlong 4 octet
     bool needs_escaping = false;
-    for (auto const& c: view) {
+    for (auto it = name.begin(); it != name.end(); ++it) {
+        unsigned char c = static_cast<unsigned char>(*it);
         if (tail) {
             if ((c & 0xc0) != 0x80) {
                 return {false, false};


zclifford:
OK the checks were unhappy (even though the code is fine):

```
/home/runner/work/qpdf/qpdf/libqpdf/QPDF_Name.cc:64:27: error: conversion to ‘unsigned char’ from ‘char’ may change the sign of the result [-Werror=sign-conversion]
   64 |     for (unsigned char c: name) {
      |                           ^~~~
```

Lemme figure out how to edit a pull request :)

zclifford:
I apparently suck at editing github pull requests

jberkenbilt:
Just edit/commit with --amend locally and force push. :-) I never edit pull requests in any other way.

zclifford:
Yeah I deleted the commit and re-created it, which apparently closed the PR so I had to manually reopen it. Will try your suggestion next time.

Git is still a bit foreign to me (what do you mean the rest of the world doesn't use perforce anymore??), especially when tied to a big UI like githubs.

libqpdf/QPDF_Name.cc
```diff
@@ -57,14 +55,12 @@ QPDF_Name::unparse()
...
 QPDF_Name::analyzeJSONEncoding(const std::string& name)
 {
-    std::basic_string_view<unsigned char> view{
-        reinterpret_cast<const unsigned char*>(name.data()), name.size()};
-
     int tail = 0;       // Number of continuation characters expected.
     bool tail2 = false; // Potential overlong 3 octet utf-8.
     bool tail3 = false; // potential overlong 4 octet
     bool needs_escaping = false;
-    for (auto const& c: view) {
```
m-holger, line 67: We can simply change c -> ch and  view -> name here and stick with a range based for loop.

libqpdf/QPDF_Name.cc
```diff
@@ -57,14 +55,12 @@ QPDF_Name::unparse()
...
+    for (auto it = name.begin(); it != name.end(); ++it) {
```
jberkenbilt, line 62: Can't this also be a range for loop? You can change it, or I can merge and then fix. Thanks.

jberkenbilt (changes requested):

jberkenbilt (changes requested):
Minor nit -- if you don't get to it, I'll merge and fix. Thanks.

libqpdf/QPDF_Name.cc
jberkenbilt: as in `for (auto const& it: name) {` and just adjust the cast to `it` rather than `*it`.

jberkenbilt:
I pushed #1167, which includes this change plus my tweak to it. If it passes CI, I will merge it, bringing this along for the ride. Thanks.

zclifford:
Thanks for getting that for me. I never check work email on weekends as a matter of principle, and I was totally out of commission Monday and Tuesday with a bad cold!

jberkenbilt:
@zclifford No problem. :-) qpdf isn't work mail for me, so I sometimes ignore it during the week ;-)

PR #1169, opened 2024-03-06
Add bad xref table test

m-holger:
I think the test should produce a "accepting invalid xref table entry" warning and exit code 3.

xref table:
```
xref
0 7
0000000000 65535 f don't panick 
0000000009 00000 n life is too short 
0000000063 00000 n 





0000000135 00000 n 
0000000307 00000 n 
0000000403 00000 n 




















0000000438 00000 n 
trailer <<
```


jberkenbilt (approved):
I agree. Let's take this as a reminder.

PR #1170, opened 2024-03-07
Refactor QPDF::parse_xrefEntry

m-holger:
Move reading of the entry from read_xrefTable to parse_xrefEntry.

Split parse_xrefEntry into two new methods read_xrefEntry and read_bad_xrefEntry. read_xrefEntry is optimized for reading correct entries. To handle incorrect entries it calls read_bad_xrefEntry, which is largely unchanged from parse_xrefEntry.

Also remove some string creation/copying in read_xrefTable.

The change improves runtime for the tests using the abuild-manual file by over 5% on clang-15, rising to around 25% on the short "extract first page" test.


jberkenbilt (approved):
Looks good -- merge at will.

PR #1171, opened 2024-03-08
Refactor the creation of unresolved objects

m-holger:
Create unresolved objects (only) during parsing of the xref table. Create dangling references as indirect nulls.


m-holger:
This PR highlights that the object cache potentially holds multiple entries for a single object id. Normally only one of them will be non-null. However, this can lead to multiple non-null objects for the same id as the result of replacing null objects through calling replaceObject or replaceReservedObject. This includes calls by JSONReactor, as at the moment nothing prevents a JSON file to include multiple objects with same id and different  gens.

For objects in the xref table, does it make sense to replace objects with the same id but different gen?

For objects not in the xref table, does it make sense to replace objects with gen != 0?

zenmhui:
how brilliant it is！

jberkenbilt:
This looks like a work in progress. Is it? I'm not seeing exactly what the salient change is, and I see dangling comments and commented out code. That said, as we have discussed, I am in strong support of changing the internal representation so that we can't represent having two objects with the same object ID and different generations since this is actually not valid. The exception to this is supporting incremental update and preserving the ability to retrieve older versions, but I think the way to do this is to actually create the concept of a QPDF object that updates an earlier QPDF object. Then the earlier one would have the old generation, and the newer one would have the new generation.

m-holger:
This was meant to be ready for review, The comments you spotted where oversights.

I have slightly amended the code in view of your email re incremental updates, namely dangling references in pdf files are now immediately changed to direct nulls. This has no impact on incremental updates as those will need to be tracked in separate  versions of the object cache and reference tables anyway.

One issue that occurred to me after making the change is the effect of xref table recovery. If additional objects are found during xref recovery, some of the dangling reference may turn out to be objects after all. It may be necessary to set up a separate table of dangling references to keep track of them without polluting the object cache.

libqpdf/QPDF_json.cc
```diff
@@ -420,12 +414,6 @@ QPDF::JSONReactor::containerEnd(JSON const& value)
         // Handle dangling indirect object references which the PDF spec says to treat as nulls.
```
jberkenbilt, line 414: This comment is slightly confusing now since there's no code left.  Perhaps just changing `Handle` to `Ignore` would be adequate.

jberkenbilt (approved):
I think this seems like a good change. My only request is to fix the wording of one comment. If I understand this correctly, it would cause a behavior change if someone managed to add an object with a specific object ID that previously appeared as a dangling reference, but you'd really have to jump through hoops to do that. This change seems more likely to eliminate some potential surprise than to create one, and it also looks like an efficiency improvement. We should put something in the release notes indicating the change...something like "When the PDF parser encounters an indirect reference to an object that doesn't exist in the file, it converts the reference to a null earlier. This is unlikely to affect any existing code, but if you were relying on the ability to provide a value for a dangling object (which would require overriding the library's default behavior), you may see a difference in behavior." If I've missed something here and the above statement is inaccurate, let me know.

libqpdf/QPDF_json.cc
m-holger: I have taken the whole comment out. I think the comment only makes sense if you know what the code was previously, and does not add anything

jberkenbilt: sounds good

m-holger:
> We should put something in the release notes ...

I agree. I'll have a think about the wording. I think we need to make clear that the parser behaviour only applies when QPDF parses an input file and not when the user calls the parse method, which allows the creation of dangling references.

PR #1179, opened 2024-03-20
In FUTURE, treat uninitialized object handles as null

m-holger:
API breaking

Treating uninitialized objects as null has a number of benefits:

- For performance reasons, qpdf uses shared nulls in a number of places. This is usually safe but can cause problems if they get accidentally mutated, e.g. by replacing an object with one of those shared nulls. Treating default constructed objects as null create a truly  immutable null.

- It removes the odd situation that a default constructed object is considered invalid by most of qpdf.

- It resolves the issue of getTrailer  returning invalid objects and calling getRoot potentially being a logic error for certain input files.

- It makes qpdf safer. Because default constructed object handles are considered invalid by most of qpdf, it is difficult to reach much of the code with them. Using them as shared nulls ensures that all code gets exercised with null pointers.

Using an uninitialized object handle as a shared null is also marginally more efficient.

The change is unlikely to affect library users. The only scenario I can think of would be when an uninitialized object handle is used to indicate a failure when a shared null would be valid which does seem a very odd situation.

### DIFF
--- a/examples/pdf-mod-info.cc
+++ b/examples/pdf-mod-info.cc
@@ -131,7 +131,7 @@ main(int argc, char* argv[])
         QPDFObjectHandle fileinfo;
 
         for (auto const& it: Keys) {
-            if (!fileinfo.isInitialized()) {
+            if (!fileinfo) {
                 if (filetrailer.hasKey("/Info")) {
                     QTC::TC("examples", "pdf-mod-info has info");
                     fileinfo = filetrailer.getKey("/Info");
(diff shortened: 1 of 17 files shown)


m-holger:
CI failure is due to a libgcc library mismatch

jberkenbilt:
Let's close this. Per my email response, I recommend
* Remove isUninitialized() for qpdf 12
* Make the default-constructed QPDFObjectHandle behave as an immutable null
* Document that users relying on isUninitialized and for whom letting the default-constructed QPDFObjectHandle behave as null doesn't work should instead use std::variant or code some other way of detecting that the object was never initialized.

PR #1180, opened 2024-03-23
Fix pikepdf future CI

m-holger:
Ensure pikepdf is build with QPDF_FURURE enabled.


m-holger:
@jbarlow83  thanks!

jberkenbilt (approved):
oops, thanks

PR #1182, opened 2024-03-29
optimize fix-qdf

oltolm:
I noticed that unlike the previous Perl version `fix-qdf` reads the entire file at once. I changed it to read a file line by line instead to reduce memory consumption.

### DIFF
--- a/include/qpdf/QUtil.hh
+++ b/include/qpdf/QUtil.hh
@@ -94,6 +94,11 @@ namespace QUtil
     QPDF_DLL
     FILE* safe_fopen(char const* filename, char const* mode);
 
+#ifdef _WIN32
+    QPDF_DLL
+    std::shared_ptr<wchar_t> win_convert_filename(char const* filename);
+#endif
+
     // The FILE* argument is assumed to be the return of fopen. If null, throw std::runtime_error.
     // Otherwise, return the FILE* argument.
     QPDF_DLL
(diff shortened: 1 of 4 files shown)


qpdf/fix-qdf.cc
```diff
@@ -380,13 +372,20 @@ realmain(int argc, char* argv[])
...
     }
-    std::string input;
-    if (filename == nullptr) {
+    if (filename.empty()) {
         filename = "standard input";
-        QUtil::binary_stdin();
-        input = QUtil::read_file_into_string(stdin);
     } else {
-        input = QUtil::read_file_into_string(filename);
+#ifdef _WIN32
```
jberkenbilt, line 378: I don't want to see any `#ifdef _WIN32` outside of QUtil or code whose specific purpose is encapsulating platform differences. This needs to be encapsulated in a different way. Maybe `QUtil::call_main_from_wmain` can be used. You can see how qpdf.cc does it.

include/qpdf/QUtil.hh
```diff
@@ -94,6 +94,11 @@ namespace QUtil
     QPDF_DLL
     FILE* safe_fopen(char const* filename, char const* mode);
 
+#ifdef _WIN32
```
jberkenbilt, line 97: We can't have anything in the API be influenced by a preprocessor definition or have any public function that is not documented.

jberkenbilt (changes requested):
This seems reasonable minus a few issues. If you can address the issues I raised, I can a deeper look. I'm not opposed to the idea or the general approach.

qpdf/fix-qdf.cc
oltolm: Can you give me some more guidance? I'm not sure how to solve this problem. I need some way to pass a wide string to `open`.

PR #1185, opened 2024-04-07
Update README-maintainer.md for clang-format-18

m-holger:


jberkenbilt (changes requested):
How about "Code is formatted with clang-format. See [format-code](./format-code) for the minimum required version." Then we don't have to keep changing it and letting it get out of date.

m-holger:
I think just shortening clang-format-16 to clang-format should do the trick given the rest of the paragraph.

jberkenbilt (approved):

PR #1187, opened 2024-04-13
Tweak wording on unreferenced objects in qdf.rst

jberkenbilt:
...based on a debian bug report.

This was reported in https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1067196.

### DIFF
--- a/manual/qdf.rst
+++ b/manual/qdf.rst
@@ -84,9 +84,10 @@ nothing generally ever references it by number.
 
 It is not generally practical to remove objects from QDF files without
 messing up object numbering, but if you remove all references to an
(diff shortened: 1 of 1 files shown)


m-holger:
I think it would be easier to change the object to null and remove indirect references in arrays (if any) only. Also. I don't think there is a need to run fix-qdf before running the file through qpdf.

jberkenbilt:
> I think it would be easier to change the object to null and remove indirect references in arrays (if any) only. Also. I don't think there is a need to run fix-qdf before running the file through qpdf.

I'm not sure this would be easier.

PR #1194, opened 2024-04-30
Fix bugs in #1161

m-holger:
Fix bugs detected by fuzzing.


jberkenbilt (approved):
Thanks for jumping right on this.

PR #1195, opened 2024-05-04
Fix bug in QPDFWriter::preserveObjectStreams

m-holger:
Code failed to allow for QPDF::getCompressibleObjSet deleting objects from the object cache in case of multiple entries for the same object id.

Add fuzz test case 68668.


m-holger:
The mingw64 CI failure looks like a bug in mingw to me. I added a work around to QUtil::call_main_from_wmain. The macOS issue seems to have resolved itself too.

jberkenbilt (approved):

PR #1196, opened 2024-05-07
Refine reconstruction of xref tables and trailers

m-holger:
When xref table recovery is triggered by failure to find a valid trailer, accept the last valid trailer rather than the first valid trailer.


m-holger:
xref table creation and reconstruction raises a number of questions, including:

1. QPDF objects without a root object are generally unusable and will eventually throw a damagedPDF exception. Would it make sense to check for the root object early?

2. Should a trailer without /Root key be accepted as a valid trailer (which would necessarily end in a damagedPDF exception)? It would seem to make sense to treat such trailers as damaged and try to find a better candidate.

3. qpdf generates a warning if the highest object number in the xref table does not match the trailer /Size key. The spec states

        Any object in a cross-reference section whose number is greater than this
        value shall be ignored and defined to be missing by a conforming reader.

   Should qpdf remove objects with ids >= /Size from the xref table?

jberkenbilt:
> xref table creation and reconstruction raises a number of questions, including:
> 
>     1. QPDF objects without a root object are generally unusable and will eventually throw a damagedPDF exception. Would it make sense to check for the root object early?

Yes, sounds like a good idea.

>     2. Should a trailer without /Root key be accepted as a valid trailer (which would necessarily end in a damagedPDF exception)? It would seem to make sense to treat such trailers as damaged and try to find a better candidate.

I agree -- look for a better option.

>     3. qpdf generates a warning if the highest object number in the xref table does not match the trailer /Size key. The spec states
>        ```
>         Any object in a cross-reference section whose number is greater than this
>         value shall be ignored and defined to be missing by a conforming reader.
>        ```

I think this makes sense to do.

jberkenbilt:
@m-holger I agree with your suggestions. Feel free to move forward in that direction and let me know when this is ready for review. I haven't reviewed the changes since I assume you will make additional changes based on the answers to the questions.

PR #1200, opened 2024-05-21
Fix QPDF::tableSize

m-holger:
Apply temporary fix to deal with fuzz case 68915. The bug high-lighted does not affect any released version of qpdf.

(Error is an integer overflow which would immediately cause a runtime error as a result of a call to QIntC::to_size.)


jberkenbilt (approved):
Agree with this interim solution. I'd consider a permanent fix low priority as this is a sensible check either way.

PR #1201, opened 2024-05-22
QPDF::processXRefStream

m-holger:
Break processXRefStream into smaller methods.

Other changes:
- simplify some pointer arithmetic
- simplify calculation of object ids / handling of subsection transitions by adding an outer loop of subsections around the loop of entries
- change the processed Index array into an array of pairs <first object id, number of entries> to simplify
- simplify int casting

### DIFF
--- a/include/qpdf/QPDF.hh
+++ b/include/qpdf/QPDF.hh
@@ -1032,7 +1032,7 @@ class QPDF
     processXRefW(QPDFObjectHandle& dict, std::function<QPDFExc(std::string_view)> damaged);
     int processXRefSize(
         QPDFObjectHandle& dict, int entry_size, std::function<QPDFExc(std::string_view)> damaged);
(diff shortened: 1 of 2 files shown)


m-holger:
The PR raises a couple of issues:

- the PDF spec states a number of requirements, eg ascending object ids and non-overlapping sections, that are not checked by qpdf. Should qpdf check and warn if input files are non-coforming?

- qpdf's handling of objects with id >= /Size is non-conforming. Presumably that should be fixed.

See comments in code for further detail.

jberkenbilt (approved):
I'm fine with these changes. I looked through it with medium carefulness...enough to get the gist, but not enough to be 100% certain that memory/boundary errors weren't introduced. But I didn't really see anything that looked potentially dangerous.

jberkenbilt:
Feel free to merge when ready.

PR #1205, opened 2024-06-06
Iterative update object maps

jberkenbilt:
@m-holger Would you mind taking a look? If you don't have time, I can merge and you can look at it after the fact. My goal isn't to be super-optimal, but I don't want to be foolish about it. This is intended to be a surgical fix. It is in response to a long-time user of qpdf who uses it on Windows with the DLL. Certain files were failing to linearize because very deep structures in the Outline hierarchy were causing a stack overflow. There was nothing technically wrong with the files. This switches a recursive function. This commit also adds some IDE configuration that enabled me to build and debug this in Windows with CLion, which is a game changer in my ability to debug Windows issues.

### DIFF
--- a/libqpdf/QPDF_optimization.cc
+++ b/libqpdf/QPDF_optimization.cc
@@ -54,6 +54,14 @@ QPDF::ObjUser::operator<(ObjUser const& rhs) const
     return false;
 }
 
+QPDF::UpdateObjectMapsFrame::UpdateObjectMapsFrame(
+    QPDF::ObjUser const& ou, QPDFObjectHandle oh, bool top) :
+    ou(ou),
+    oh(oh),
+    top(top)
+{
+}
+
 void
 QPDF::optimize(
     std::map<int, int> const& object_stream_data,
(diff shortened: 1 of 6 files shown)


jberkenbilt:
once I fix the memory issues...

libqpdf/QPDF_optimization.cc
```diff
@@ -277,78 +285,70 @@ QPDF::pushInheritedAttributesToPageInternal(
 
 void
 QPDF::updateObjectMaps(
-    ObjUser const& ou,
-    QPDFObjectHandle oh,
+    ObjUser const& first_ou,
+    QPDFObjectHandle first_oh,
     std::function<int(QPDFObjectHandle&)> skip_stream_parameters)
 {
+    std::vector<std::shared_ptr<ObjUser>> heap;
```
m-holger, line 292: std::vector<ObjUser> should work ...

libqpdf/QPDF_optimization.cc
```diff
@@ -277,78 +285,70 @@ QPDF::pushInheritedAttributesToPageInternal(
...
-                // Don't traverse into stream parameters that we are not going to write.
-            } else {
-                updateObjectMapsInternal(
-                    ou, dict.getKey(key), skip_stream_parameters, visited, false);
+            for (auto const& key: dict.getKeys()) {
+                if (is_page_node && (key == "/Thumb")) {
+                    // Traverse page thumbnail dictionaries as a special case.
+                    auto ou = std::make_shared<ObjUser>(ObjUser::ou_thumb, cur.ou.pageno);
+                    pending.emplace_front(*ou, dict.getKey(key), false);
+                    heap.emplace_back(ou);
```
m-holger, line 341: ... with 
```c++
heap.emplace_back(ObjUser::ou_thumb, cur.ou.pageno);
pending.emplace_back(heap.back(());
```
We may not even need a vector. A scalar ObjUser variable may be sufficient.

jberkenbilt: I actually figure out how to do it without seeing your example, suggesting my brain is starting to extend around emplace_back. :-) I think I need a vector. I don't want to assume there isn't some invalid data that would make there be more than one of these.

libqpdf/QPDF_optimization.cc
jberkenbilt: You are right. I'll fix.

libqpdf/QPDF_optimization.cc
jberkenbilt: I don't need a vector. I do need `top` though, but I can't find your comment so maybe you realized and removed it.

m-holger: The ```if(is_page_node) ``` should ensure that this is a top-level key, preventing the possibility of any other `/Thumb` keys being processed.

m-holger: I can't spot why `top` is needed.

jberkenbilt: > The `if(is_page_node) ` should ensure that this is a top-level key, preventing the possibility of any other `/Thumb` keys being processed.

Right, I explained that in my comment. I think `top` is needed because pending could become empty at any time, such as when traversing the last key of the dictionary.

m-holger: > Right, I explained that in my comment. 

I only saw your update after my comment.

>  I think top is needed because pending could become empty at any time

Good point. I think this needs a test case. Probably also for  copyForeignObject which could similarly get caught out.

jberkenbilt: > I only saw your update after my comment.

We each did that to each other multiple times in this PR, which I guess shows we're thinking along the same path -- a good sign.



> Good point. I think this needs a test case. Probably also for copyForeignObject which could similarly get caught out.

Perhaps. We'd have to do it with a single-key dictionary to make it deterministic, and this would be a good case for a QTC coverage case since it's easy for the test to pass without exercising the boundary condition in the event of a logic change. This is the kind of thing that is what prevents me from doing too much optimization. With `top`, the intent is clear, and there is no need for tricky or subtle reasoning or fragility with edge cases, and there's no real performance cost other than the extra few bytes.

PR #1207, opened 2024-06-07
Disable PointerHolder by default

jberkenbilt:
I intend to merge this when we are ready to start merging qpdf 12 ABI-breaking changes. This is not itself ABI breaking, but it is something I don't want to introduce until qpdf 12.


jberkenbilt:
Adding `abi` label so I remember not to merge this before we open the window for qpdf 12.

PR #1214, opened 2024-06-18
Fuzz future

jberkenbilt:
Build fuzzers with FUTURE=ON and FUTURE=OFF so we fuzz future changes.


jberkenbilt:
@m-holger FYI. I think this will do it. I'll look at CI and, if all seems right, I will merge this. It will take us a day to be sure the new fuzzers are being run. Note that the corpuses for the new fuzzers won't be as rich as they will not include what's accumulated over time,  but it should still be valuable, and the seed corpus contains all cases that triggered errors in the past.

m-holger:
Thanks

PR #1215, opened 2024-06-21
Add a new example of QPDFJob

capric0rn78:
A new example demonstrating the use of `QPDFJob` to remove all Javascript & Action object in a PDF file.

### DIFF
--- a/examples/qpdfjob-remove-js-action.cc
+++ b/examples/qpdfjob-remove-js-action.cc
@@ -27,14 +27,7 @@ main(int argc, char* argv[])
 
         auto names = pdf.getRoot().getKey("/Names");
         if (names.isDictionary()) {
-            auto js = names.getKey("/JavaScript");
-            if (js.isDictionary()) {
-                auto nested_names = js.getKey("/Names");
-                if (nested_names.isArray()) {
-                    js.replaceKey("/Names", "null"_qpdf);
-                }
-                names.replaceKey("/JavaScript", "null"_qpdf);
-            }
+            names.removeKey("/JavaScript");
         }
 
         for (auto obj: pdf.getAllObjects()) {
--- a/examples/qpdfjob-remove-js-action.cc
+++ b/examples/qpdfjob-remove-js-action.cc
@@ -1,0 +1,63 @@
+// This is an example how to use QPDFJob for removing all Javascript and Action in a PDF file
+
+#include <qpdf/QPDFJob.hh>
+#include <qpdf/QPDFUsage.hh>
+#include <qpdf/QUtil.hh>
+
+#include <cstdio>
+#include <cstdlib>
+#include <iostream>
+
+int
+main(int argc, char* argv[])
+{
+    if (argc != 3) {
+        std::cerr << "Usage: " << argv[0] << " <input file> <output file>\n";
+        return 1;
+    }
+
+    const char* inputFilePath = argv[1];
+    const char* outputFilePath = argv[2];
+
+    try {
+        QPDFJob j;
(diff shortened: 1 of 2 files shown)


examples/qpdfjob-remove-js-action.cc
```diff
@@ -0,0 +1,63 @@
...
+                    obj.replaceKey("/S", "null"_qpdf);
+                }
+
+                auto js = obj.getKey("/JS");
+                if (js.isString()) {
+                    obj.replaceKey("/JS", "null"_qpdf);
+                }
+
+                obj.replaceKey("/Action", "null"_qpdf);
+                pdf.getRoot().replaceKey("/Action", "null"_qpdf);
```
m-holger, line 53: I don't think either `obj` or the root object have an /Action entry

examples/qpdfjob-remove-js-action.cc
```diff
@@ -0,0 +1,63 @@
...
+                names.replaceKey("/JavaScript", "null"_qpdf);
+            }
+        }
+
+        for (auto obj: pdf.getAllObjects()) {
+            if (obj.isDictionaryOfType("/Action")) {
+                auto s = obj.getKey("/S");
+                if (s.isName() && s.getName() == "/JavaScript") {
+                    obj.replaceKey("/S", "null"_qpdf);
+                }
```
m-holger, line 45: This can be written more compactly as
```cpp
std::string name;
for (auto obj: pdf.getAllObjects()) {
   if (obj.isDictionaryOfType("/Action") && obj.getKey("/S").getValueAsName(name) && 
       name == "/JavaScript") {
          // do whatever you have to
   }
}
```
I assume what you want to do is disable the action. Assuming that an empty string is valid javascript, setting the /JS entry to an empty string may work. Setting the /S entry to null is not valid as this a required entry for action dictionaries, and if you leave this as /JavaScript, /JS is also required. 

You could try to delete the action completely, but then you would have to make sure that the action is removed everywhere where the action is used and a null object is not acceptable, which feel a little bit tricky. My first choice would be to try an empty string.

examples/qpdfjob-remove-js-action.cc
```diff
@@ -0,0 +1,63 @@
...
+
+    try {
+        QPDFJob j;
+        j.config()->inputFile(inputFilePath)->outputFile(outputFilePath)->checkConfiguration();
+        auto qpdf_sp = j.createQPDF();
+        auto& pdf = *qpdf_sp;
+
+        auto names = pdf.getRoot().getKey("/Names");
+        if (names.isDictionary()) {
+            auto js = names.getKey("/JavaScript");
```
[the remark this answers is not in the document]
capric0rn78: You're right, we don't need to check it first. `removeKey(...)` also delete the entire key.

examples/qpdfjob-remove-js-action.cc
capric0rn78: Code removed.

examples/qpdfjob-remove-js-action.cc
capric0rn78: I've updated the code as suggested. But replace /JS with empty string eventually will delete /Action too.

[input.json](https://github.com/user-attachments/files/16321199/input.json)
[output.json](https://github.com/user-attachments/files/16321200/output.json)

capric0rn78:
@jberkenbilt nothing specific, just want to learn how to use QPDFJob for this example. Should I change it to use QPDF instead?

m-holger:
I think it is fine to use QPDFJob as a high-level API to deal with the reading and writing of the PDF file. For C++ it is neither here nor there, but from other languages it is likely to be more convenient as the whole job config can be passed as a single JSON string.

PR #1221, opened 2024-06-25
Refine handling of severely damaged files

m-holger:
Check for presence of page tree early and apply a sanity check on object ids during xref table recovery.

This should fix fuzz case 69857.

As a result of the new checks we are loosing the test case for self-referential object streams. Do be need to construct a new test file for this?


jberkenbilt:
> As a result of the new checks we are loosing the test case for self-referential object streams. Do be need to construct a new test file for this?

Yes, we should construct a new test case if have lost something.

PR #1222, opened 2024-06-26
In PL_DCT add option to limit the size of uncompressed corrupt data 

m-holger:
This addresses fuzzing issues, where corrupt JPEG data can produce high memory usage or time-outs. The solution adopted is to not decompress corrupt data if their expanded size is large.

For qpdf, it is felt that it is better to accept the impact and to attempt to decompress.

However, should we capture JPEG warning messages and redirect them to the qpdf logger in order to ensure that qpdf's return code reflects the warnings?


include/qpdf/Pl_DCT.hh
```diff
@@ -34,6 +34,11 @@ class QPDF_DLL_CLASS Pl_DCT: public Pipeline
     QPDF_DLL
     Pl_DCT(char const* identifier, Pipeline* next);
 
+    // Constructor for decompressing image data. If corrupt_data_limit is non-zero and the data is
+    // corrupt, only attempt to uncompress if the uncompressed size is less than corrupt_data_limit.
+    QPDF_DLL
+    Pl_DCT(char const* identifier, Pipeline* next, size_t corrupt_data_limit);
```
jberkenbilt, line 40: Usually when I add a new constructor whose first arguments overlap an existing one, I add an ABI comment reminding me to make the new argument optional for the next ABI window, but it doesn't matter.

jberkenbilt:
> However, should we capture JPEG warning messages and redirect them to the qpdf logger in order to ensure that qpdf's return code reflects the warnings?

If not too much trouble, perhaps, though qpdf's default mode doesn't decompress jpegs, so I'm not sure it's that important.

PR #1229, opened 2024-07-02
Clean dct fuzz changes

jberkenbilt:
@m-holger Please take a look at this. If you agree, feel free to merge (if CI passes). If not, please push back or make additional changes.

I feel strongly against using conditional compilation in qpdf without a very good reason, and in particular, building the library differently for fuzzing is a slippery slope. Conditional compilation is appropriate when there are real OS/Compiler differences that can't be handled any other way or when some part of the system may need to be disabled because of OS limitations, such as avoiding certain file system operations on embedded hardware or avoiding wchar_t on minimal distributions. Otherwise, I prefer configuration.

I added setMaxWarnings to QPDF but did not expose it via a command-line flag. I could do that if we think it's useful. I'm not sure why you added it exactly since warnings are not shown in the fuzz tests, but I wonder whether having a cap on the number of warnings might be sensible. Anyway, I don't think it's that important, but I guess the question is still there about whether we should add a command-line flag like --max-warnings. Adding command-line flags is tricky but documented. I don't think we need it.


jberkenbilt:
Also, I would like you to get into the habit of updating ChangeLog. The ChangeLog format conforms to the gnu format, including the particular line break and indentation style, so let's not reformat ChangeLog. My goal for ChangeLog is that it's a level of granularity between the release notes and the git history. It gives someone a chance to scan the list of changes from one release to another and see what might be relevant. Also, ChangeLogs are chronological, while the release notes are organized by category. I try to do ChangeLog as I go. When working on features, I try to do release notes as I go as well, but I'm also likely to build the release notes by looking at the ChangeLog and diffing public headers.

jberkenbilt:
Closing in favor of #1230

PR #1230, opened 2024-07-04
Alternative clean dct fuzz changes

m-holger:
Alternative to #1229. See https://github.com/qpdf/qpdf-dev/discussions/5 for detail.


jberkenbilt (approved):
This looks good to merge to me. It's better than what we have now.

PR #1248, opened 2024-07-20
Update TODO.md and .clang-format

m-holger:


.clang-format
```diff
@@ -41,7 +41,7 @@ IndentCaseBlocks: true
 IndentExternBlock: Indent
 IndentPPDirectives: AfterHash
 IndentWidth: 4
-InsertBraces: true
```
jberkenbilt, line 44: Actually, if we could leave this as `true`, I'd appreciate it. This is one of the few that I actually care about. I find it visually difficult when I'm scanning the code if braces are missing around single statements.

jberkenbilt (changes requested):
It would be great if we can leave InsertBraces alone. Do you have a strong feeling about it? This is probably the one that masters most to me.

jberkenbilt:
If you really care, I can live with the change as long as we don't remove braces where they are already there. I will always use them. Interestingly, in my new job, this is the one single code formatting issue that will get your pull-request rejected in C++, perhaps because there's rust and go code, and you can't omit braces in those languages.

m-holger:
This is probably one that I feel least strongly about, although I think not requiring unnecessary braces is more idiomatic C++ / K&R style and I personally prefer not to require unnecessary braces as long as it doesn't interfere with readability. ( I think llvm at https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements got it right in this respect).

I'll bat this back to you for final decision.

jberkenbilt:
If you don't feel strongly about it, let's leave it as is. It's a little bit of cognitive load for me every time I see one without braces. If you remove the .clang-format change, then go ahead and merge. Thanks.

PR #1253, opened 2024-07-27
Refactor Pl_QPDFTokenizer

m-holger:
Reduce unnecessary copying for performance reasons

### DIFF
--- a/include/qpdf/QPDFTokenizer.hh
+++ b/include/qpdf/QPDFTokenizer.hh
@@ -191,6 +191,8 @@ class QPDFTokenizer
     // returns a tt_inline_image token.
     QPDF_DLL
     void expectInlineImage(std::shared_ptr<InputSource> input);
+    QPDF_DLL
+    void expectInlineImage(InputSource& input);
 
   private:
     friend class QPDFParser;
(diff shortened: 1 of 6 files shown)


m-holger:
@jberkenbilt  , is there any reason for the content normalizer to preserve trailing spaces and blank lines?

jberkenbilt:
That's a tough one. My impulse is that it should leave them alone. I feel like I kind of want to know if the original file has trailing whitespace and blank lines. It's been a while, but I have encountered files with some uncompressed content that uses trailing spaces for padding purposes for naive, text-based form-filling. Back years ago, there was online greeting card thing we used to use that generated PDFs that way, and I might not have known if qdf stripped that stuff. Of course, people can use regular uncompress without qdf, but qdf has behaved a certain way for a long time, I don't think we should gratuitously change it.

m-holger:
Ok

jberkenbilt (approved):
These kinds of changes are hard to review....it's easy to insert subtle errors. I reviewed this pretty carefully and didn't find anything wrong. Hopefully if there are any memory errors (which I don't think there are), the fuzzer will catch them.

PR #1262, opened 2024-08-05
Fix writing reals with trailing '.' as JSON (fixes #1261)

### LINKED ISSUE #1261: JSON output can be invalid when number in PDF contains trailing dot
First, thanks for this amazing library! It works great.

We use the JSON output to read the MediaBox/CropBox values for each page of user-submitted PDFs and noticed that if the MediaBox/CropBox values contain a float in the format `123.` (no trailing digit), qpdf will pass that value through to the output JSON. Unfortunately, not having a trailing decimal on a number with a dot is invalid JSON, so it causes the parser to crash. We've worked around this by stripping trailing dots on numerics before parsing the JSON, but thought I'd bring it up in case a future fix is possible.

To summarize, given this PDF value

```
/MediaBox [ 0 0 612. 792. ]
```

I'd expect a JSON representation of

```
{ "/MediaBox": [ 0, 0, 612, 702 ] }
```

but instead get

```
{ "/MediaBox": [ 0, 0, 612., 702. ] }
```

which doesn't parse.

Thanks again!

m-holger:


jberkenbilt:
Thanks for jumping on it. Do you think it's better to append `0` than to strip the trailing `.`? I guess it probably is...Not that you need it, but I'll "approve" this fix since I think it's a perfectly acceptable one.

jberkenbilt (approved):

m-holger:
I see pros and cons, but on balance I lean towards appending a 0.

PR #1269, opened 2024-08-16
Fix handling of hybrid reference files in QPDF::read_xrefTable

m-holger:
QPDF::read_xrefTable ignores type 0 entries for objects in a section if an associates XRefStm has an entry for the same object.

The spec states:

When the conforming reader searches for an object, if an entry is not found in any given standard cross-reference section, the search shall proceed to a cross-reference stream specified by the XRefStm entry before looking in the previous cross-reference section,

If a deleted entry is found in a section, the XRefStm is not searched according to the standard.


jberkenbilt (approved):
I think this looks like a correct change, though I am not certain. Anyway, I think my original implementation was incorrect and this one looks better.

jberkenbilt:
I see this is also part of #1272, which I am reviewing.

PR #1271, opened 2024-08-20
Fix QPDF::recoverStreamLength

m-holger:
Ensure the the recovered stream end is not part of a different object.

Test file is bad24.pdf with stream 4 'endstream' corrupted.

The test file raises an interesting issue - in this case /Length is actually valid. Should we in this situation fall back to /Length rather than length 0 ?


jberkenbilt (commented):
I think this is okay. Rather than falling back to the previous length or 0, you could consider just bumping up to the next object (basically where `end` is). Perhaps a more robust solution would be to use the `/Length` that's there if it falls within the allowable region. If not, then assume everything up to the next object is in the stream but ignore `endobj` if you find it.

There's a pathological corner case of someone having one PDF as an embedded file in another PDF and having the embedded PDF be uncompressed, but I'm reasonably certain that neither qpdf nor anything else would do a good job of recovering in that situation, so I think we can ignore the case of where the contents of a stream legitimately include `endstream` AND the xref table is messed up.

jberkenbilt:
I see this is also part of #1272, which I am reviewing.

m-holger:
For convenience, I think I am going to merge as is and add a separate PR after refactoring the object cache as follows:

- if recovered stream length == 0, check whether /Length lies in stream object 
   - if  /Length lies in stream object, recovered stream length = /Length

Once the handling of deleted objects has been refactored this can be further refined based on the start of the next object (including deleted objects).

jberkenbilt:
> For convenience, I think I am going to merge as is and add a separate PR after refactoring the object cache as foll

Yup, sounds good.

PR #1272, opened 2024-08-23
Refactor QPDF xref table 

m-holger:
Encapsulate the xref table and trailer into a new class and use a vector indexed by object id as the underlying data structure with the aims of ensuring:
- there is a single entry for each object id,
- deleted objects are properly represented in the table, and
- access is better controlled to ease maintenance. 

The PR is made up of parts:

1. Move QPDF inner class definitions to a new private header file

     commit 'Move QPDF inner class definitions to new QPDF_private.hh'

2. Create new inner class Xref_table to encapsulate the xref table and trailer

     commits 'Add new class QPDF::Xref_table' to 'Move QPDF::readTrailer to QPDF::Xref_table'

     This involves moving relevant members of Members and QPDF methods to the new class. These changes are largely 
 mechanical and routine, with the exception of commit 'Move QPDF::reconstruct_xref to QPDF::Xref_table', where when a trailer is recovered from a xref stream the last valid trailer is used rather than the first, as a result of not callind setTrailer. 

3. Add accessor methods to Xref_table and restrict access to data members

    commits 'Add QPDF::Xref_table methods type, offset, stream_number ...' to 'Make all QPDF::Xref_table data members private'

    This includes adding  a temporary method 'as_map' to provide a map view to the xref table for code that can't be refactored until later.

4. Refactor QPDF::emptyPDF and createFromJSON

    This is mainly to avoid the complications arising from 'createFromJSON' creating an invalid xref table, i.e. a table that does not contain the required root object. 

5. Refactor Xref_table methods

    commits 'Refactor Xref_table::parse_first' to 'Refactor Xref_table::reconstruct'

    The main aims are to read the trailer before reading the xref table in order to be able to ignore entries with for objects with id > /Size as required by the PDF spec and to process reconstructed object last to first in order to avoid the special case for insertion into the xref table.

6. Change the underlying table to a vector

   commits 'Change Xref_table::table to std::vector' to 'Change Xref_table::table to std::vector' and 'Remove Xref_table::deleted_objects'

7. Change the entry type for efficiency

8. Refactor code that uses a map-based copy of the xref table

    commits 'Refactor QPDFWriter::preserveObjectStreams' to 'Refactor QPDF::getXRefTable'

9. Refactor Xref_table::subsections for efficiency in the normal case of encountering a valid table.

10. Move 'end_before/after_space' from ObjCache to Xref_table

    Both items are only relevant for objects read from a PDF file, i.e. objects in the xref table.

The main question raised by the PR is whether there are circumstances where we would want to ignore the /Size key of an apparently valid trailer / xref table combination, i.e. are there validations that we want to add for the /Size value and/or do we want to provide an option to override the /Size value.


libqpdf/QPDF.cc
```diff
@@ -1355,26 +1355,30 @@ QPDF::Xref_table::insert_reconstructed(int obj, qpdf_offset_t f1, int f2)
...
         QPDFXRefEntry const& entry = iter.second;
         cout << og.unparse('/') << ": ";
         switch (entry.getType()) {
         case 1:
-            cout << "uncompressed; offset = " << entry.getOffset();
+            cout << "uncompressed; offset = " << entry.getOffset() << "\n";
             break;
 
         case 2:
-            *m->log->getInfo() << "compressed; stream = " << entry.getObjStreamNumber()
```
jberkenbilt, line 1376: Interesting that the old code was inconsistent. I wonder which way it should be. Anyway, changing it to cout is okay.

jberkenbilt (approved):
I actually went through every commit. Some I read thoroughly, and some I skimmed. I think these look like significant improvements to the code. After this, it will definitely no longer feel like my own code to me, and I don't know how comfortable I will be working in the internals. That said, I don't think there will be any reason for me to be in this part of the code. I'm not sure whether this is done, but I'm on board with the approach.

I appreciate the way you broke this into commits. I read it commit by commit but did not look at the aggregation, which is too big to deal with in a meaningful way. I noticed some temporary things that disappeared and at least one that didn't (duplicated warnings), and I noticed a few lines that weren't covered, though most of them look like it's okay for them not to be covered.

I'm comfortable with your merging these changes whenever you feel that they are ready to go.

PR #1274, opened 2024-08-25
Add new commands --remove-metadata and --remove-info

### LINKED ISSUE #1145: Show and or / Clean PDF & Remove Meta Data Etc
ExifTool is a great Meta Data tool for all sorts of file formats (I even got them to show meta data on text files)..  https://exiftool.org/

It would be nice however if ExifTool would show all the same if not more Meta Data in the PDF and other structural information on the PDF..   --SHOWMETADATA

The second and bigger request would be a clean one step process for removing any and all meta data tied to privacy etc.

Some people have a process where they use both ExifTool then QPDF to remove the meta data.

https://www.reddit.com/r/libgen/comments/12ai3mz/how_to_remove_all_metadata_identifiers_when/

Having a PDF Cleaning Process in one simple "clean" process would be nice.

m-holger:
Partially fixes #1145


m-holger:
qpdf probably should not preserve any existing /ModDate value.

mborsetti:
Thanks for the great and fast work!

> qpdf probably should not preserve any existing /ModDate value, but instead update it with the current date.

I completely agree that qpdf should not preserve any existing /ModDate value. But probably it should be removed as well, since it's part of /Info, in order to adhere to the [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

Even better, could `--remove-info` mimic the syntax of  [`--add-attachment`](https://qpdf.readthedocs.io/en/stable/cli.html#option-add-attachment) and allow users to set/modify the dates with [`--creationdate`](https://qpdf.readthedocs.io/en/stable/cli.html#option-creationdate) and [`--modifieddate`](https://qpdf.readthedocs.io/en/stable/cli.html#option-modifieddate)?  That would be the best solution IMHO, as it's the most flexible and transparent.

m-holger:
> However, it probably should be removed,

Under certain conditions /ModDate is a required entry. Adding the necessary checks to see whether it can be removed safely would be nice, but there are higher priority items on the list.

> allow users to set/modify the dates with [--creationdate](https://qpdf.readthedocs.io/en/stable/cli.html#option-creationdate) and [--modifieddate](https://qpdf.readthedocs.io/en/stable/cli.html#option-modifieddate)

Given that the /Info dictionary is deprecated in PDF 2.0 this would feel lower priority than extracting / replacing the metadata and converting /Info to metadata.

mborsetti:
> Under certain conditions /ModDate is a required entry. Adding the necessary checks to see whether it can be removed safely would be nice, but there are higher priority items on the list.

Got it

> Given that the /Info dictionary is deprecated in PDF 2.0 this would feel lower priority than extracting / replacing the metadata and converting /Info to metadata.

Ah, wasn't aware.  Makes sense.

Thanks so very much! ❤️

jberkenbilt:
I know pikepdf has code that deals with metadata. I haven't touched it in qpdf for not wanting to add dependencies to work with xml or to hand-create xml textually. I finally broke down and added a JSON parser to qpdf a couple of years ago, but a JSON parser can be written in a day. XML is a lot harder. Back in the olden days, when I started working on qpdf as a closed-source, internal tool for the company I worked for at the time, it was part of a larger framework that included parser generators as well as an image library with support for a subset of tiff, jpeg, and png, an XML library that wrapped xerces-C, a database library with support for sqlite and Oracle, and a system for working with Unicode that wrapped ICU. When I left that job, I had permission to extract the PDF parts and make them open source, but there are better answers to all the other things my framework did, or they are proprietary to what that company did at the time (and no longer does).

I have often considered whether qpdf should split into a core library with minimal dependencies (perhaps just zlib and openssl/gnutls) and one or more libraries layered on top. A big part of qpdf's success has been that it works and builds anywhere. Then an outer library could add support for the image filters, xml metadata, and maybe some other things. I'm just musing. Maybe it's something @m-holger and I can discuss if it feels like a good idea.

Regarding moddate and attachments, when qpdf is ready to require C++20 (no earlier than when Ubuntu 20.04 goes out of security support near April 2025), then we will have access to the filesystem bits of the standard library, which will enable us to more easily copy dates from files without lots of conditional code. That's not full relevant to this discussion, but I mention it because the topic came up.

I had actually forgotten that `/Info` was deprecated in PDF 2.0.

jberkenbilt (approved):
I think this is clearly a desired feature, and it looks like your dealing with the mechanics of adding the CLI argument is correct. I didn't carefully study the test files, but it looks like it's the right tests.

jberkenbilt:
This is probably a case where I would go ahead and include a release notes entry, but you've got it in the ChangeLog, so reviewing that for release notes would catch it.

m-holger:
> I have often considered whether qpdf should split into a core library with minimal dependencies (perhaps just zlib and openssl/gnutls) and one or more libraries layered on top. A big part of qpdf's success has been that it works and builds anywhere. Then an outer library could add support for the image filters, xml metadata, and maybe some other things.

continued at https://github.com/qpdf/qpdf-dev/discussions/14

PR #1284, opened 2024-09-17
Use NAMES keyword in find_library call

cbwoyak:
Updates the find_library call for zlib to use the NAMES keyword.

This fixes an issue encountered when building qpdf from source on Windows with MSVC 2022 and targeting zlib also built from source.


jberkenbilt:
Thanks -- I'll make sure this is in for the next stable release.

PR #1304, opened 2024-11-01
Refactor Objects::recover_stream_length

m-holger:
Change how the maximum stream length is calculated. For streams not in the xref table, instead of returning 0 length return the maximum length that does not overlap with a known object or xref table. For further detail see #1271.

PR is last 3 commits only, remainder is /1302 and earlier.


jberkenbilt (approved):
I have only skimmed the code, but I think returning the longest length that doesn't overlap is definitely better than 0, so I am in favor of the change.

PR #1305, opened 2024-11-02
Refactor InputSource

m-holger:


m-holger:
@jberkenbilt , could you review the two bug fix commits in case I am missing something.

jberkenbilt:
Can you let me know what your doubts are? I haven't read this part of the spec for ages, and this is a bit out of context.

m-holger:
It seems to me that there are 2 bugs in isLinearized:

1. If the first object found does not start with a dictionary open token, the method continues searching for the next object. The spec says that the linearization parameter dictionary should be the first object, not the first dictionary object.

>  Following the header, the first object in the body of the file (part 2) shall be an indirect dictionary object, the
>  linearization parameter dictionary, 


2. If the /L is not an integer, the method returns that the file is linearized. According to the spec:

>  (Required) The length of the entire file in bytes. It shall be exactly
>  equal to the actual length of the PDF file. A mismatch indicates that
>  the file is not linearized and shall be treated as ordinary PDF

and the comment in the code

 >  // If the first object in the file is a dictionary with a suitable /Linearized key and has an /L
 >  // key that accurately indicates the file size, initialize m->lindict and return true.

I am just wondering whether I am missing some subtle point.

jberkenbilt:
You're probably not missing any subtle point. I wrote the linearization code a really long time ago based on the 1.3 spec, and lots of empirical evidence and trial and error and probably made some mistakes. It strikes me as unlikely that there's a ton of code out there that makes decisions based on whether or not qpdf says a file is linearized. Anyway, these are probably good to fix.

m-holger:
>  lots of empirical evidence and trial and error

Some of the empirical evidence have turned into the law in 2.0, e.g. your guess about offsets if '%PDF-' is not at the start of the file has now been codified into the rule that offsets are counted  from the '%' char of the header.

jberkenbilt:
So much of how PDF works ends up being codification of things that Adobe reader happened to accept, causing numerous PDF files to appear in the wild. For so long, implementing good PDF readers involved a lot of reverse engineering of what Adobe did.

PR #1310, opened 2024-11-19
DOC: Fix some typos in cli.rst

StefRe:


jberkenbilt:
Thanks for taking the time to submit this. I'll take a look at make sure I agree with the changes.

@m-holger In case you end up taking this, you probably realize that changes to cli.rst fail CI without regeneration of automatic files. Rather than approving CI workflows for cases like this, my typical approach is to pull the change down locally, regenerate automatic files, add that as a second commit, and push the result to main, which carries along the doc commit and passes CI.

jberkenbilt:
My intention is to merge this with the auto job generation stuff after we resolve the `next` branch issue.

PR #1311, opened 2024-11-21
Add emscripten build

j3k0:
Integrates the necessary options to build a Javascript library, with tests and documentation.

Tested with emscripten 3.1.72

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -404,3 +404,63 @@ message(STATUS "")
 if(NOT (MULTI_CONFIG OR CMAKE_BUILD_TYPE))
   message(WARNING "  CMAKE_BUILD_TYPE is not set; using default settings")
 endif()
+
+option(EMSCRIPTEN "Build with Emscripten" OFF)
+
+if(EMSCRIPTEN)
+  set(CMAKE_C_COMPILER emcc)
+  set(CMAKE_CXX_COMPILER em++)
+  set(CMAKE_EXECUTABLE_SUFFIX ".js")
+
+  # Define runtime methods to export
+  set(EXPORTED_RUNTIME_METHODS
+    "callMain"
+    "FS"
+  )
+
+  # Enhanced debugging and safety flags
+  set(EMSCRIPTEN_LINK_FLAGS "\
+    -s WASM=1 \
+    -s ABORTING_MALLOC=1 \
+    -s SUPPORT_LONGJMP=1 \
+    -s EXPORTED_RUNTIME_METHODS='[${EXPORTED_RUNTIME_METHODS}]' \
+    -s EXPORT_ALL=1 \
+    -s FORCE_FILESYSTEM=1 \
+    -s ALLOW_MEMORY_GROWTH=1 \
+    -s MODULARIZE=1 \
+    -s USE_ZLIB=1 \
+    -s USE_LIBJPEG=1 \
+    -s BUILD_AS_WORKER=1 \
+    -s PROXY_TO_WORKER=1 \
+    -s ENVIRONMENT='worker' \
+    -s INVOKE_RUN=0 \
+    -s EXIT_RUNTIME=0 \
+    -s MAIN_MODULE=1 \
+    -s EXPORT_NAME='QPDF' \
+    -s ASSERTIONS=0 \
+    -s STACK_OVERFLOW_CHECK=0 \
+    -s DEMANGLE_SUPPORT=0 \
+    -s ERROR_ON_UNDEFINED_SYMBOLS=0 \
+    -s DISABLE_EXCEPTION_CATCHING=1 \
+    -s SAFE_HEAP=0 \
+    -s SAFE_HEAP_LOG=0 \
+    -s EXCEPTION_DEBUG=0 \
+    -s EXCEPTION_STACK_TRACES=0 \
+    -O2"
+  )
+
+  # Add sanitizer flags for additional runtime checks
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${EMSCRIPTEN_LINK_FLAGS} -fsanitize=address -fsanitize=undefined")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${EMSCRIPTEN_LINK_FLAGS} -fsanitize=address -fsanitize=undefined")
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${EMSCRIPTEN_LINK_FLAGS}")
+  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} ${EMSCRIPTEN_LINK_FLAGS}")
+
+  # Use native crypto
+  set(REQUIRE_CRYPTO_NATIVE ON)
+  set(REQUIRE_CRYPTO_OPENSSL OFF)
+  set(USE_IMPLICIT_CRYPTO OFF)
+
+  # Static linking
+  set(BUILD_SHARED_LIBS OFF)
+  set(BUILD_STATIC_LIBS ON)
+endif()
--- a/README-emscripten.md
+++ b/README-emscripten.md
@@ -1,0 +1,89 @@
+# Building QPDF with Emscripten
+
+This guide explains how to compile QPDF to JavaScript/WebAssembly using Emscripten.
+
+> **Note**: For pre-built bundles and a convenient wrapper around the WebAssembly build, check out [qpdf.js](https://github.com/j3k0/qpdf.js).
+
+## Prerequisites
+
+Install Emscripten by following the [official installation instructions](https://emscripten.org/docs/getting_started/downloads.html).
+
+## Quick Start
+
+1. Create and enter build directory:
+```bash
+mkdir build-emscripten && cd build-emscripten
+```
+
+2. Build and copy to examples:
+```bash
+# Configure and build
+emcmake cmake .. \
+    -DEMSCRIPTEN=ON \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DBUILD_SHARED_LIBS=OFF && \
+emmake make -j8 qpdf && \
+cp qpdf/qpdf.* ../examples/qpdf-js/lib/
+```
+
+3. Test the build:
+```bash
+cd ../examples/qpdf-js
+python3 serve.py
+```
+
+4. Open `http://localhost:8523/test.html` in your browser
+
+## Build Output
+
+The build produces these files in `build-emscripten/qpdf/`:
+- `qpdf.js`: JavaScript glue code
+- `qpdf.wasm`: WebAssembly binary
+
+## Browser Usage Example
+
+Here's a minimal example to get started:
+
+```html
+<!DOCTYPE html>
+<html>
+<head>
+    <title>QPDF Test</title>
+</head>
+<body>
+    <pre id="output"></pre>
+    
+    <script src="qpdf.js"></script>
+    <script>
+        function print(text) {
+            document.getElementById('output').textContent += text + '\n';
+            console.log(text); // Also log to console for debugging
+        }
+
+        QPDF({
+            keepAlive: true,
+            logger: print,
+            ready: function(qpdf) {
+                // Test QPDF is working
+                qpdf.execute(['--copyright'], function(err) {
+                    if (err) {
+                        print('Error: ' + err.message);
+                        return;
+                    }
+                    print('QPDF loaded successfully!');
+                });
+            }
+        });
+    </script>
+</body>
+</html>
+```
+
+For a complete working example with PDF operations like checking and encryption, see [examples/qpdf-js/test.html](examples/qpdf-js/test.html) in the repository.
+
+## Additional Resources
+
+- [qpdf.js](https://github.com/j3k0/qpdf.js) - Pre-built WebAssembly bundles with JavaScript wrapper
+- [Emscripten Documentation](https://emscripten.org/docs/getting_started/index.html)
+- [WebAssembly](https://webassembly.org/)
+- [QPDF Documentation](https://qpdf.readthedocs.io/) 
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ appimage/build
 .cache
 /html
 Doxyfile
+/examples/qpdf-js/lib/
--- a/README.md
+++ b/README.md
@@ -142,6 +142,10 @@ qpdf is known to build and pass its test suite with mingw and Microsoft Visual C
 work. In addition to the manual, see [README-windows.md](README-windows.md) for more details on how to build under
 Windows.
 
+## WebAssembly / JavaScript Support
+
+QPDF can be compiled to WebAssembly for use in browsers and JavaScript environments. For most users, we recommend using [qpdf.js](https://github.com/j3k0/qpdf.js) which provides pre-built bundles and a convenient wrapper around the WebAssembly build. If you need to build the WebAssembly version yourself, see [README-emscripten.md](README-emscripten.md) for instructions.
+
 # Building Documentation
 
 The qpdf manual is written in reStructured Text format and is build with [sphinx](https://www.sphinx-doc.org). The
--- a/examples/qpdf-js/qpdf-worker.js
+++ b/examples/qpdf-js/qpdf-worker.js
@@ -1,0 +1,385 @@
+/* eslint-env worker */
+/* global FS, callMain */
+
+const scriptURL = self.location.href;
+const basePath = scriptURL.substring(0, scriptURL.lastIndexOf('/') + 1);
+let errorOrWarning = '';
+
+function stdout(txt) {
+  postMessage({
+    type: 'stdout',
+    line: txt
+  });
+}
+
+function debugLog(type, ...args) {
+  const msg = args.map(arg => {
+    if (arg instanceof Error) {
+      return arg.message + '\n' + arg.stack;
+    }
+    if (typeof arg === 'object') {
+      return JSON.stringify(arg, null, 2);
+    }
+    return String(arg);
+  }).join(' ');
+  // stdout(`[DEBUG:${type}] ${msg}`);
+}
+
+
+var Module = {
+  thisProgram: 'qpdf',
+  noInitialRun: true,
+  print: function(text) {
+    stdout(text);
+  },
+  printErr: function(text) {
+    if (text.startsWith('WARNING: ')) {
+      errorOrWarning = text.slice(8);
+    }
+    stdout('[stderr] ' + text);
+  },
+  onRuntimeInitialized: function() {
+    debugLog('init', 'QPDF runtime initialized');
+    postMessage({
+      type: 'ready'
+    });
+  },
+  locateFile: function(path, prefix) {
+    if (path.endsWith('.wasm')) {
+      return basePath + 'lib/' + path;
+    }
+    return prefix + path;
+  },
+  quit: function(status, toThrow) {
+    debugLog('quit', `Program quit with status: ${status}`, toThrow);
+    if (toThrow) {
+      throw status;
+    }
+  }
+};
+
+importScripts(basePath + 'lib/qpdf.js');
+
+function getFileData(fileName) {
+  try {
+    if (!FS.analyzePath(fileName).exists) {
+      debugLog('getFileData', `File ${fileName} does not exist`);
+      return null;
+    }
+    const file = FS.root.contents[fileName];
+    if (!file) {
+      debugLog('getFileData', `Could not access ${fileName} in FS`);
+      return null;
+    }
+    debugLog('getFileData', `Successfully read ${fileName}, size: ${file.contents.length}`);
+    return file.contents;
+  } catch (error) {
+    debugLog('getFileData', 'Error reading file:', error);
+    return null;
+  }
+}
+
+onmessage = function(event) {
+  var message = event.data;
+
+  switch (message.type) {
+    case 'save': {
+      const filename = message.filename;
+      const arrayBuffer = message.arrayBuffer;
+      
+      try {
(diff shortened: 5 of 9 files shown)


jberkenbilt:
This is a significant change to drop in without any discussion. I think the goal of being able to compile qpdf to web assembly is a good goal, but I think we should discuss the best way to achieve it. My general approach has been not to add additional complexity to qpdf's build but rather to have external layers on top of it and to enhance qpdf as needed to make it possible to extend it without additional modifications to qpdf. Coincidentally, I may be working on an unrelated web assembly thing for my day job and may be learning more about this, but I have the fundamentals now.

What would help is if you can provide some details about what your approach was, how you expect this to be used, and why you've chosen this particular approach. We will also need to decide how this will be exercised in CI so it doesn't break. At a glance, I'm not seeing why any of these changes need to be part of qpdf itself, but I'll admit to having just skimmed through this.

j3k0:
Yes there could be an external tool or repository that builds QPDF in wasm (or just a documentation?). The same could be said for every platform I guess. The main question regarding that is discoverability, maintenance and usefulness.

I'll let you decide on whether an officially supported JS version of QPDF makes sense or not, I can help with maintaining it.

Regarding the approach, the use case is to bring the power of QPDF to browser-like environments. For NodeJS, there are native wrappers already, so a JS only version isn't really useful. With that in mind, the build settings are tweaked to build a version that runs within a web-worker as a single self-contained pair of files (`.js` and `.wasm`). I'm providing example high-level wrappers for the generated files in the `examples/qpdf-js` directory:

**examples/qpdf-js/qpdf-worker.js** - https://github.com/qpdf/qpdf/pull/1311/files#diff-e3d8e5b875868b5ea51a5d1d4fc84d458613b8b0f51a3e1142dc223417c2de30
It's the code for the web worker, it loads and interacts with the compiled qpdf binary directly, provides helpers to save and load files in the virtual filesystem, report execution logs and errors to the upper layer. Big part of that file tries to extract error details from the address of thrown exceptions, https://github.com/qpdf/qpdf/pull/1311/files#diff-e3d8e5b875868b5ea51a5d1d4fc84d458613b8b0f51a3e1142dc223417c2de30R200 -- this was a pain point that could be improved by patching qpdf native `main()` function to try catch and format exceptions in a way that could be parsed by the javascript -- JSON maybe.

**examples/qpdf-js/qpdf.js** - https://github.com/qpdf/qpdf/pull/1311/files#diff-13fad4a5122c3869dcc33605469acd3b9971ea3b427ba022e5fe518d7b88930c
The side user code interacts with. It loads and terminate the web worker (after a period of inactivity). It exposes method to interact with it. Calls are converted to worker messages, then success or failures are reported in callbacks. At the most basic level, it can be used as an API to call QPDF cli with given command line arguments, also to create and load files. Then there's a helper function for encrypting a PDF file, that demos what could be a potentially more user friendly interface. Eventually all possible QPDF modes should be method calls. For example, we'd rather use `qpdf.check(fileContent, callback)` instead of `qpdf.save("input.pdf", fileContent); qpdf.exec(["--check", "input.pdf"], callback)`.

Regarding CI. It would be relatively easy to make sure JS version builds. Concerning he test suit, it is built with emscripten, but it will need some tool-chain to run those test in a headless browser, which isn't part of that PR. Minimally, we could also just make sure the operations performed in "test.html" work (which would also require a headless browser environment).

roihitvishwakarma (approved):

jberkenbilt:
Sorry for the delay. I'm just trying to find some time to dig into this. I'm interested in accepting this in some form and maybe even trying qpdf wasm myself. I'm hoping that I'll find some time over the holidays to dig into this deeply enough, but please don't take the delay as anything other than a sign of a maintainer with a lot going on. :-)

jberkenbilt:
To be more specific, before I accept this, I need some CI for it, and I need to understand the JavaScript. I can create the CI, and doing so will be a good exercise for me to test the docs you've provided and to make sure I understand how it works. Understanding the JavaScript is something I just have to find time to do. I'm going to try to find a few hours in the next couple of weeks, at which point I will either accept this with some additional commits I add or will make comments with specific requests. I haven't "approved" the workflows for CI because there's nothing here that would impact any existing CI. There will also need to be some addition to the manual. Most likely I'd write a brief section in the part about using qpdf from other languages. It doesn't have to repeat the README you provided, but it should refer to it, and it would be fine to link to same resource of your prebuilt version.

Questions for you:

Are there any additional artifacts that should be included in the official qpdf release that would be useful, or is your intention just to update your wrapper when qpdf is released?

You mentioned that you could help maintain it. What do you see as the lifecycle of https://github.com/j3k0/qpdf.js? One scenario I could imagine would be that the qpdf releases page include the wasm build and possibly the glue code. A quick glance through the glue suggests that it's mainly a wrapper around QPDFJob and, as such, wouldn't need to change very much. Specific feature-based things would live in your repo. That would be consistent with how other wrappers work. I'm thinking you've set up this PR so that someone who knows what they're doing could use wasm/js for qpdf directly from this repo without requiring your external stuff, but your external stuff adds extra value and capability. That seems sensible to me, and I'm okay with drawing the line in that spot. If you think there will be churn in the bit that you've included (or you want to update/release it more often than qpdf releases), it might make sense for some of it to live in your repo. At a minimum, the build enhancements should be in the qpdf repo. That said, if I'm going to produce a wasm artifact, I need a way to test it, so I think maybe this is right as is.

As we shift around some of the maintainership responsibilities of qpdf, it has temporarily drifted away from a state where I feel that I could release at any moment. I'm not sure when the next release will be, but probably not for at least a couple of months.

So, to recap:
* I will work to understand the JavaScript part of this. The build part looks fine as is.
* I will work on CI and documentation updates but would also take input in the form of additional commits on this PR. That said, I suspect it will take me just a few minutes to do the doc and CI. The vast majority of my time will on understanding the JS and thinking through the lifecycle.

jberkenbilt:
I've had a chance to do a deeper dive. Here's what I'd like to do:
* take the cmake changes as they are or perhaps refined a little in a way I can consider
* create a new top-level executable that takes a JSON string as its sole argument (probably something in QPDFJob JSON format) and generates a newly defined JSON structure as its output. This should be a well-defined, documented structure that contains warning and error information. I think all informational output qpdf gives (or nearly all) can come from qpdf's JSON output. This new output format would have to consider that as well. My hope would be that the WASM wrapper could interface with that tool rather than the main of qpdf without any loss of functionality and that that would enable you to drop all exception parsing code
* Add something to qpdf's CI to exercise invoking this tool in some small list of specified ways...similar to your smoke test but maybe with something like running a tool that gives json-formatted output with possible warnings or errors.
* Create an install target in cmake that creates a bundle that includes the WASM file containing the qpdf-whatever CLI tool that exists for this purpose and including that as an item in the qpdf official release area
* Updating qpdf's documentation and website to point to your JS wrapper.

The idea would be that qpdf would include as parts of its "contract" this purely JSON-based way of interacting and a working WASM build but would leave the vast majority of the javascript and html code you've provided up to your repo. Your job there should be simple -- you (or your users) should be able to consume a binary artifact from qpdf's release page that will be loosely coupled with whatever wrapper layer you add.

I can help with implementation of the build code and the C++ main program, which will be short because most of the work is already done in QPDFJob. What I would need would be your input on this and help testing it. Mostly the part I don't know is what you would ideally like the bundle to contain and if you have any particular input on the JSON output of this tool. I'd think it should have something like `{"output": ..., "warnings": [...], "error": ...}` perhaps inside of something `{"qpdf-output-vn": ...}` or something else we can standardize on. Or I could add additional fields to qpdf's existing JSON output. I haven't looked at what's there for a while and don't necessarily keep it my head, so I'd have to refresh myself.

Given this, I think it would only be a few hours of work on my side to create something that would simplify your work and increase its discoverability.

What do you think of this approach? Please let me know if you have any specific input on the details of any of this. Otherwise, I will try to glean it from what you've already created.

As for timing, I can't predict when I will have some time for this, but I think it sounds interesting and useful, and I would like to prioritize it soonish, like next time I have a decent chunk of qpdf time. I am hoping to get another qpdf release out in the next few months as there have been a lot of accumulated changes. Though as this is a side project, I can't commit to any specific timeframe.

j3k0:
Before I could say what I think of this approach, I had to do a bit of reading (thanks for your responses, that gave me a starting point).

Let's see if I understand correctly: the `qpdf-job-json-whatever` CLI tool wouldn't work with command line arguments like the human consumable `qpdf`. Instead it would get its configuration as a JSON document in [QPDFJob JSON format](https://github.com/qpdf/qpdf/blob/main/libqpdf/qpdf/auto_job_schema.hh) and output a single JSON document (with a format to define). A formatted format like JSON should be superior to interface with.

That schema file is generated by [generate_auto_job](https://github.com/qpdf/qpdf/blob/main/generate_auto_job) from [job.yml](https://github.com/qpdf/qpdf/blob/main/job.yml) with the documentation for that schema being extracted from [cli.rst](https://github.com/qpdf/qpdf/blob/main/manual/cli.rst).

With that new knowledge, I figured it would be possible and interesting to automatically generate the javascript interface (and associated typings). I could eventually tap into `generate_auto_job`, because it contains all the parsing logic and knowledge about special types of arguments. This script could use a bit of cleanup, but it looks scary to refactor. A quick option would generate an additional file: `main/qpdf_job.schema.json` in the [JSON Schema](https://json-schema.org) format (a standard JSON schema definition format). *JSON Schema* opens the door to using [existing tools](https://json-schema.org/tools) for validation, documentation generation and even code generation, [quicktype for example](https://github.com/glideapps/quicktype).

`qpdf_job.schema.json` can be then used to generate various languages bindings to the format and documentation (which code, in my case, would be in the qpdf-js repository). Admittedly, it's a bit backward to have the *JSON Schema* file generated when it could be the source of truth and have all the other files generated from it, but could be a good first step to refactoring 

The output JSON schema can also be defined in that format.

A pain point specific to the Javascript interface is to identify command line arguments that are filenames as they require special treatment (file content need to be passed as a buffer, blob, base64 or something). If they could be tagged in some way. I could keep a list of those in qpdf-js, but tagging them at the source feels better.

Even better, could we pass the file content in the input JSON either:
- by using [data-url](https://developer.mozilla.org/en-US/docs/Web/URI/Schemes/data) were filenames are expected.
- by having a dedicated JSON field, like `"inputFileBase64": "<base64....>"`.
- by having a "filesystem" field in the input JSON (in which case file reads will use this instead of the system):

```json
{
  "...": "...",
  "filesystem": {
    "/tmp/input.pdf": "<base64>",
  }
}
  ```

In that same spirit, the output file(s) could be written to the output JSON document instead of the filesystem.

Obviously, I'll let you judge if that makes sense to have that options in `qpdf-job-json`, but I'm mentioning the idea. That probably requires to have file reads and writes quite encapsulated in the C++ code to be able to overload them.

To sum up:

- I like the idea to interface with a `qpdf-job-json` executable.
- The `generate_auto_job` script could output a complete *JSON Schema* file, this file can be used by the various bindings to that executable (including qpdf-js).
- Add a *JSON Schema* file for the `qpdf-job-json` output as well.
- A large portion of `generate_auto_job` could later be removed, by using existing code generation tools with that *JSON Schema* as an input, same for documentation maybe.
- Somehow tag command line arguments that are filenames (in `job.yml` or `cli.rst`)
- Have the ability to provide file content to QPDFJob in the JSON instead of filenames (data-url, virtual filesystem) / and to tell `qpdf-job-json` to write files in the output JSON instead of filesystem.

**Comment by jberkenbilt** [comment]
> Let's see if I understand correctly: the `qpdf-job-json-whatever` CLI tool wouldn't work with command line arguments like the human consumable `qpdf`. Instead it would get its configuration as a JSON document in [QPDFJob JSON format](https://github.com/qpdf/qpdf/blob/main/libqpdf/qpdf/auto_job_schema.hh) and output a single JSON document (with a format to define). A formatted format like JSON should be superior to interface with.

Yes, that's basically it. There are some nuances we'd have to work through as some commands generate textual output and it is possible to write a PDF to standard output. What the qpdf CLI does is write certain things to stderr if PDF output is written to stdout. This is not really a big issue because you don't _have_ to run qpdf in a way that generates output to standard output. So what you said is my intention with a few avoidable caveats.

Your remarks about `generate_auto_job` needing some cleanup and being scary to refactor are true. Even I find it scary, and I wrote the original version. Your observation about using JSON schema and so forth are also good. My original hope was to have something that you could use to generate a wrapper interface automatically with. I had even envisioned being able to create a qpdf gui...imagine something with various buttons or drop downs or whatever that you could select that would construct a qpdf CLI and/or job JSON format. I know some have asked for something like this, but I don't think there would be a ton of demand for it. I think it would be a useful proof of concept that some kind of job JSON schema contained enough information though. The path you've outlined is pretty consistent with my original thinking though perhaps a little further thought out.

I'll provide a little background. Prior to the creation of QPDFJob JSON, the qpdf executable contained way too much logic, and there was a lot of functionality, like image optimization, and page splitting, that were not exposed to the library API. I made QPDFJob JSON as a step toward providing more useful higher-level APIs for those things, and I did this as a step toward the "pages epic" that I've been talking about for years and may or may not ever actually finish. It depends on whether I or any other potential qpdf maintainers have the time to devote to it.

The implementation of QPDFJob allows specification of a qpdf "job" using command-line arguments or a JSON file specification. I've been quite extreme over the years in never breaking the qpdf CLI. On one hand, this is good for my users in that scripts that use qpdf should continue to work. On the other hand, the qpdf CLI is a train wreck, and some of qpdf job JSON reflects that. I did some normalization and cleanup and have gradually introduced more normalized command-line arguments, but it's hard to fully move away without breaking stuff. If you look at the implementation of QPDFJob, there's a big pile of common code, a big pile of automatically generated code, and then a bunch of CLI-specific code and a bunch of JSON-specific code that implement the same logic for some of the more complicated bits. _Most_ new features/arguments/etc. fall into the simple case of just adding a simple argument and a corresponding JSON key, and there's a blurb in README-maintainer.md that explains the steps to add a new command-line flag. To date, only two people have ever done this. Adding something with a subparser is more involved, and even if I have to do it, I takes me more than one try and I have to refer back to old commits as a reminder. This is all just part of qpdf showing its age and accumulated baggage from being a side project for its whole life. This isn't to say I think there's a problem with the code quality or anything...just that I fully acknowledge that this aspect of things is less polished than I wish it were.

So.....pragmatically, we'd have to make a decision. Going all the way to the goal of being able to use real JSON schema instead of my kind of bespoke self-documenting-ish JSON documents, or going to something that would fully describe what an interface wrapper would need to know is a pretty heavy lift that requires more effort than I have to put into it. If someone else felt really adventurous and wanted to take this on, I'd be happy to advise, but it will be a slow process as my qpdf time comes in bursts and it takes me a long time to trust someone enough to give them free reign. :-)

BUT I think we can get far enough without going all the way there. Something that uses QPDFJob in its current form with all its warts will require special-case code for some operations but will still be a lot easier to deal with than trying to invoke the qpdf CLI directly. Depending on how this goes, we could consider a larger project that creates a QPDFJob v2 that uses schema, quicktype, or some other technology as long as it is done in a way that doesn't break the ability to build qpdf basically everywhere, now possible because it has so few external dependencies.

I'd suggest a program called `qpdf-job-json` that takes, as an argument, either a single JSON document or something like `--stdin`, which will read the document from standard input. It could also take `--help`, `--verison`, `--copyright`, or similar if we wanted it to. Its output would be some JSON document perhaps containing `output`, `warnings`, `error` key along with some kind of version field. I could probably create this in a couple of hours without touching `generate_auto_job` or really any other existing code and could work through the nuances around stdout/stderr. You could then move forward with that and see where we end up. Once we learn what we need to know from your writing a better JavaScript/Typescript API, we can iterate.

I'll mention that, in my day job, I recently went through the exercise of creating a nodejs extension using neon in rust to wrap a C API that was in turn wrapping a rust API. The original C API had some patterns that were not conducive to writing an API wrapper in a garbage-collected language like JavaScript, but as I had control over the API, I was able to extend it. The QPDFJob API only works with files and doesn't suffer from this problem, but if someone ever wanted to write a wrapper around the qpdf C API in a garbage-collected language, they would run into certain problems. In particular, there is memory created by the qpdf library that is all freed at once when you free the main qpdf handle, which means that some pointers' lifetimes are tied to operations performed on other pointers. This is tricky when connecting to a garbage collector, but in practice, it's probably not a big deal for qpdf. But you're only talking about wrapping functionality qpdf provides from the CLI, and given that there are already fully functional native JavaScript interfaces for working PDF files, I'm not sure we need to go any farther down that path.

Does this help? In summary, I like how you're thinking about it, and I also think we'll have to start with something more easily reachable if we want to get this done. If you want, I can take a crack at the qpdf-job-json command-line tool, and you can see whether it's good enough to do what you want for now. Note that there is a `--job-json-help` flag that may give some information that may be useful for autogenerating some things, though it was not really intended for that purpose and may be a little ugly.

**Comment by jberkenbilt** [comment]
Maybe some of the ideas, like marking arguments that are file names, can be done without going all the way to schema, etc.

File: CMakeLists.txt [hunk 1]
```diff
@@ -404,3 +404,63 @@ message(STATUS "")
 if(NOT (MULTI_CONFIG OR CMAKE_BUILD_TYPE))
   message(WARNING "  CMAKE_BUILD_TYPE is not set; using default settings")
 endif()
+
+option(EMSCRIPTEN "Build with Emscripten" OFF)
```
jberkenbilt, line 408: All options should be up at the top -- see the comment. This line should be right below the definition of option(CXX_NEXT). Also, all configure options must be documented. There are comments in the top-level CMakeLists.txt -- search for `*** OPTIONS ***`.

CMakeLists.txt
```diff
@@ -404,3 +404,63 @@ message(STATUS "")
...
+
+if(EMSCRIPTEN)
```
jberkenbilt, line 410: I think the right place for this block is after `if(ENABLE_COVERAGE)` and before `include(CPack)`.

CMakeLists.txt
```diff
@@ -404,3 +404,63 @@ message(STATUS "")
...
+    -s DISABLE_EXCEPTION_CATCHING=1 \
+    -s SAFE_HEAP=0 \
+    -s SAFE_HEAP_LOG=0 \
+    -s EXCEPTION_DEBUG=0 \
+    -s EXCEPTION_STACK_TRACES=0 \
+    -O2"
+  )
+
+  # Add sanitizer flags for additional runtime checks
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${EMSCRIPTEN_LINK_FLAGS} -fsanitize=address -fsanitize=undefined")
```
jberkenbilt, line 453: We should not enable sanitizers by default. They have a big performance impact, and qpdf's CI runs with sanitizers, so I don't think we need them to be here for wasm.

CMakeLists.txt
```diff
@@ -404,3 +404,63 @@ message(STATUS "")
...
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${EMSCRIPTEN_LINK_FLAGS} -fsanitize=address -fsanitize=undefined")
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${EMSCRIPTEN_LINK_FLAGS}")
+  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} ${EMSCRIPTEN_LINK_FLAGS}")
+
+  # Use native crypto
+  set(REQUIRE_CRYPTO_NATIVE ON)
```
jberkenbilt, line 459: There may come a day when one of the crypto libraries is required. I'd think openssl at least must work with emscripten?

README-emscripten.md
```diff
@@ -0,0 +1,89 @@
...
+
+1. Create and enter build directory:
+```bash
+mkdir build-emscripten && cd build-emscripten
+```
+
+2. Build and copy to examples:
+```bash
+# Configure and build
+emcmake cmake .. \
```
jberkenbilt, line 21: This didn't work for me. I got errors about zlib, openssl, and jpeg, even though your wasm flags seem to not require them. I'm not sure why...a little quick tinkering didn't reveal it.

README-emscripten.md
```diff
@@ -0,0 +1,89 @@
...
+    -DEMSCRIPTEN=ON \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DBUILD_SHARED_LIBS=OFF && \
```
jberkenbilt, line 24: You explicitly set this inside the `if(EMSCRIPTEN)` block, so why do we also need it here?

jberkenbilt (changes requested):
As discussed, I don't want to include all the javascript examples. I'll make some additional comments outside the review comments.

jberkenbilt:
After giving this some more thought, I think rather than having a whole separate executable, it would make sense to have a flag for qpdf that would tell it to generate JSON-friendly output. It's getting to be a bit much with all the different json flags that mean different things. Perhaps a flag like `--json-diagnostics` could be used, but I'm open to ideas on what it should be called. When run in this way, any standard output _other than --json_ would appear base64-encoded in an `"output"` key, and there would be `"warnings"` and `"errors"` keys that would have a more predictable structure like QPDFExc. If you ran `qpdf --json --json-diagnostics` and there were no warnings or errors, it would behave exactly like `--json` behaves now. This way, you could more or less count on valid JSON coming out of qpdf, even in the event of an error. It's possible that a qpdf bug could cause other output to be generated, so programs would have to handle it robustly. We could also have qpdf write everything to a file or file descriptor which would be less likely to get intermixed with other things. This is basically the same as the earlier idea, but it would just be a command-line flag, not a whole new executable. This also means it could be specified as part of the job JSON if someone wanted to go that route.

So, for this to go in, I would want only the build support and required documentation, leaving the sample code and glue to an external repo, which I'm happy to have mentioned in the docs. The build instructions should be carefully exercised. We will also need some CI for this. I pulled this locally and did a little experimentation with the toolchain. I think it would work in CI to pull down the emsdk repository, run the "install", do the build, and then run `node qpdf.js --version` or something to make sure we got a good build. Perhaps some other commands, like running with the above options on a file with warnings and a file with an error might fit. This would be quite easy to add.

PR #1325, opened 2024-12-29
[WIP] Add JSON Schema generation to generate_auto_job

j3k0:
Add JSON Schema generation to generate_auto_job to provide a standard, machine-readable description of the QPDFJob JSON format. This enables:

- Validation in JSON Schema aware tools and IDEs
- Automated code generation for language bindings
- Better documentation of the JSON interface
- Proper identification of file paths and their content types

Key changes:

- Generate draft-2020-12 compliant JSON Schema
- Extract and format help text from CLI documentation
- Track file paths and their content types in job.yml
- Add proper schema validation for choice options
- Ensure consistent text formatting and capitalization

The schema is built from multiple sources:

1. job.yml "json" section: defines structure and relationships
3. job.yml "file_paths": identifies file paths and content types
2. cli.rst: provides detailed help text for fields

File paths are marked with uri-reference format and contentMediaType following JSON Schema best practices, with validation handled at the application level.

This is hopefully a step toward better tooling support for QPDFJob's JSON interface, making it easier to create language bindings and validate configuration files.

The generated file should probably be part of the repository, for now I didn't add it. So I it pasted below.

The output schema doesn't specify required fields and field dependencies, not sure how to handle that yet (have to decipher the "options tables" in job.yml maybe)

Ref #1311 for the initial discussion on the idea to generate a JSON Schema for use in qpdf-js

```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "https://github.com/qpdf/qpdf/blob/main/qpdf-job-json.schema.json",
  "title": "QPDF Job JSON Schema",
  "description": "A QPDF Job configuration",
  "type": "object",
  "properties": {
    "inputFile": {
      "type": "string",
      "description": "Input filename.",
      "format": "uri-reference",
      "contentMediaType": "application/pdf"
    },
    "password": {
      "type": "string",
      "description": "Password for encrypted file.\n\nSpecify a password for an encrypted, password-protected file. Not needed for encrypted files without a password."
    },
    "passwordFile": {
      "type": "string",
      "description": "Read password from a file.\n\nThe first line of the specified file is used as the password. This is used in place of the \"password\" option.",
      "format": "uri-reference",
      "contentMediaType": "text/plain"
    },
    "empty": {
      "type": "string",
      "description": "Use empty file as input.\n\nUse in place of infile for an empty input. Especially useful with \"pages\"."
    },
    "jsonInput": {
      "type": "string",
      "description": "Input file is qpdf json.\n\nTreat the input file as a json file in qpdf json format. See the \"qpdf json format\" section of the manual for information about how to use this option."
    },
    "outputFile": {
      "type": "string",
      "description": "Output filename.",
      "format": "uri-reference",
      "contentMediaType": "application/pdf"
    },
    "replaceInput": {
      "type": "string",
      "description": "Overwrite input with output.\n\nUse in place of outfile to overwrite the input file with the output."
    },
    "qdf": {
      "type": "string",
      "description": "Enable viewing pdf code in a text editor.\n\nCreate a pdf file suitable for viewing in a text editor and even editing. This is for editing the pdf code, not the page contents. All streams that can be uncompressed are uncompressed, and content streams are normalized, among other changes. The companion tool \"fix-qdf\" can be used to repair hand-edited qdf files. Qdf is a feature specific to the qpdf tool. Please see the \"qdf mode\" chapter in the manual."
    },
    "preserveUnreferenced": {
      "type": "string",
      "description": "Preserve unreferenced objects.\n\nPreserve all objects from the input even if not referenced."
    },
    "newlineBeforeEndstream": {
      "type": "string",
      "description": "Force a newline before endstream.\n\nFor an extra newline before endstream. Using this option enables qpdf to preserve pdf/a when rewriting such files."
    },
    "normalizeContent": {
      "type": "string",
      "description": "Fix newlines in content streams.\n\nNormalize newlines to unix-style newlines in pdf content streams, which is useful for viewing them in a programmer's text editor across multiple platforms. This is also turned on by \"qdf\".",
      "enum": [
        "y",
        "n"
      ]
    },
    "streamData": {
      "type": "string",
      "description": "Control stream compression.\n\nThis option controls how streams are compressed in the output. It is less granular than the newer options, \"compressstreams\" and \"decodelevel\".  parameters: - compress: same as \"compressstreams\":\"y\" \"decodelevel\":\"generalized\" - preserve: same as \"compressstreams\":\"n\" \"decodelevel\":\"none\" - uncompress: same as \"compressstreams\":\"n\" \"decodelevel\":\"generalized\"",
      "enum": [
        "compress",
        "preserve",
        "uncompress"
      ]
    },
    "compressStreams": {
      "type": "string",
      "description": "Compress uncompressed streams.\n\nSetting \"compressstreams\":\"n\" prevents qpdf from compressing uncompressed streams. This can be useful if you are leaving some streams uncompressed intentionally.",
      "enum": [
        "y",
        "n"
      ]
    },
    "recompressFlate": {
      "type": "string",
      "description": "Uncompress and recompress flate.\n\nThe default generalized compression scheme used by pdf is flate, which is the same as used by zip and gzip. Usually qpdf just leaves these alone. This option tells qpdf to uncompress and recompress streams compressed with flate. This can be useful when combined with \"compressionlevel\"."
    },
    "decodeLevel": {
      "type": "string",
      "description": "Control which streams to uncompress.\n\nWhen uncompressing streams, control which types of compression schemes should be uncompressed: - none: don't uncompress anything. This is the default with   \"jsonoutput\". - generalized: uncompress streams compressed with a   general-purpose compression algorithm. This is the default   except when \"jsonoutput\" is given. - specialized: in addition to generalized, also uncompress   streams compressed with a special-purpose but non-lossy   compression scheme - all: in addition to specialized, uncompress streams compressed   with lossy compression schemes like jpeg (dct) qpdf does not know how to uncompress all compression schemes.",
      "enum": [
        "none",
        "generalized",
        "specialized",
        "all"
      ]
    },
    "decrypt": {
      "type": "string",
      "description": "Remove encryption from input file.\n\nCreate an unencrypted output file even if the input file was encrypted. Normally qpdf preserves whatever encryption was present on the input file. This option overrides that behavior."
    },
    "deterministicId": {
      "type": "string",
      "description": "Generate id deterministically.\n\nGenerate a secure, random document id only using static information, such as the page contents. Does not use the file's name or attributes or the current time."
    },
    "staticAesIv": {
      "type": "string",
      "description": "Use a fixed aes vector.\n\nUse a static initialization vector for aes-cbc. This is intended for testing only so that output files can be reproducible. Never use it for production files. This option is not secure since it significantly weakens the encryption."
    },
    "staticId": {
      "type": "string",
      "description": "Use a fixed document id.\n\nUse a fixed value for the document id. This is intended for testing only. Never use it for production files. See also qpdf --help=--deterministic-id."
    },
    "noOriginalObjectIds": {
      "type": "string",
      "description": "Omit original object ids in qdf.\n\nOmit comments in a qdf file indicating the object id an object had in the original file."
    },
    "copyEncryption": {
      "type": "string",
      "description": "Copy another file's encryption details.\n\nCopy encryption details from the specified file instead of preserving the input file's encryption. Use \"encryptionfilepassword\" to specify the encryption file's password.",
      "format": "uri-reference",
      "contentMediaType": "application/pdf"
    },
    "encryptionFilePassword": {
      "type": "string",
      "description": "Supply password for copyencryption.\n\nIf the file named in \"copyencryption\" requires a password, use this option to supply the password."
    },
    "linearize": {
      "type": "string",
      "description": "Linearize (web-optimize) output.\n\nCreate linearized (web-optimized) output files."
    },
    "linearizePass1": {
      "type": "string",
      "description": "Save pass 1 of linearization.\n\nWrite the first pass of linearization to the named file. The resulting file is not a valid pdf file. This option is useful only for debugging qpdf.",
      "format": "uri-reference",
      "contentMediaType": "application/octet-stream"
    },
    "objectStreams": {
      "type": "string",
      "description": "Control use of object streams.\n\nControl what qpdf does regarding object streams. Options: - preserve: preserve original object streams, if any (the default) - disable: create output files with no object streams - generate: create object streams, and compress objects when possible",
      "enum": [
        "disable",
        "preserve",
        "generate"
      ]
    },
    "minVersion": {
      "type": "string",
      "description": "Set minimum pdf version.\n\nForce the pdf version of the output to be at least the specified version. The version number format is \"major.minor[.extension-level]\", which sets the version header to \"major.minor\" and the extension level, if specified, to \"extension-level\"."
    },
    "forceVersion": {
      "type": "string",
      "description": "Set output pdf version.\n\nModification options make systematic changes to certain parts of the pdf, causing the pdf to render differently from the original."
    },
    "progress": {
      "type": "string",
      "description": "Show progress when writing.\n\nIndicate progress when writing files."
    },
    "splitPages": {
      "type": "string",
      "description": "Write pages to separate files.\n\nThis option causes qpdf to create separate output files for each page or group of pages rather than a single output file.  file names are generated from the specified output file as follows:  - if the string %d appears in the output file name, it is replaced with a   zero-padded page range starting from 1 - otherwise, if the output file name ends in .pdf (case insensitive), a   zero-padded page range, preceded by a dash, is inserted before the file   extension - otherwise, the file name is appended with a zero-padded page range   preceded by a dash.  page ranges are single page numbers for single-page groups or first-last for multi-page groups.",
      "format": "uri-reference",
      "contentMediaType": "application/pdf"
    },
    "jsonOutput": {
      "type": "string",
      "description": "Apply defaults for json serialization.\n\nImplies \"json\":\"version\". Changes default values for certain options so that the json output written is the most faithful representation of the original pdf and contains no additional json keys. See also \"jsonstreamdata\", \"jsonstreamprefix\", and \"decodelevel\".",
      "enum": [
        "2",
        "latest"
      ]
    },
    "removeRestrictions": {
      "type": "string",
      "description": "Remove security restrictions from input file.\n\nRemove restrictions associated with digitally signed pdf files. This may be combined with \"decrypt\" to allow free editing of previously signed/encrypted files. This option invalidates and disables any digital signatures but leaves their visual appearances intact."
    },
    "encrypt": {
      "type": "object",
      "properties": {
        "userPassword": {
          "type": "string",
          "description": "Specify user password.\n\nrun qpdf --help=encryption for details.\n\nSet the user password of the encrypted file."
        },
        "ownerPassword": {
          "type": "string",
          "description": "Specify owner password.\n\nrun qpdf --help=encryption for details.\n\nSet the owner password of the encrypted file."
        },
        "40bit": {
          "type": "object",
          "properties": {
            "annotate": {
              "type": "string",
              "description": "Restrict document annotation.\n\nrun qpdf --help=encryption for details.\n\nEnable/disable modifying annotations including making comments and filling in form fields. For 128-bit and 256-bit encryption, this also enables editing, creating, and deleting form fields unless \"modifyother\":\"n\" or \"modify\":\"none\" is also specified."
            },
            "extract": {
              "type": "string",
              "description": "Restrict text/graphic extraction.\n\nrun qpdf --help=encryption for details.\n\nEnable/disable text/graphic extraction for purposes other than accessibility."
            },
            "modify": {
              "type": "string",
              "description": "Restrict document modification.\n\nrun qpdf --help=encryption for details.\n\nFor 40-bit files, modify-opt may only be y or n and controls all aspects of document modification.  for 128-bit and 256-bit encryption, modify-opt values allow enabling and disabling levels of restriction in a manner similar to how some pdf creation tools do it. Modify-opt values map to other combinations of options as follows:  all: allow full modification (the default) annotate: \"modifyother\":\"n\" form: \"modifyother\":\"n\" \"annotate\":\"n\" assembly: \"modifyother\":\"n\" \"annotate\":\"n\" \"form\":\"n\" none: \"modifyother\":\"n\" \"annotate\":\"n\" \"form\":\"n\" \"assemble\":\"n\""
            },
            "print": {
              "type": "string",
              "description": "Restrict printing.\n\nrun qpdf --help=encryption for details.\n\nControl what kind of printing is allowed. For 40-bit encryption, print-opt may only be y or n and enables or disables all printing. For 128-bit and 256-bit encryption, print-opt may have the following values:  none: disallow printing low: allow low-resolution printing only full: allow full printing (the default)"
            }
          }
        },
        "128bit": {
          "type": "object",
          "properties": {
            "accessibility": {
              "type": "string",
              "description": "Restrict document accessibility.\n\nrun qpdf --help=encryption for details.\n\nThis option is ignored except with very old encryption formats. The current pdf specification does not allow restriction of document accessibility. This option is not available with 40-bit encryption."
            },
            "annotate": {
              "type": "string",
              "description": "Restrict document annotation.\n\nrun qpdf --help=encryption for details.\n\nEnable/disable modifying annotations including making comments and filling in form fields. For 128-bit and 256-bit encryption, this also enables editing, creating, and deleting form fields unless \"modifyother\":\"n\" or \"modify\":\"none\" is also specified."
            },
            "assemble": {
              "type": "string",
              "description": "Restrict document assembly.\n\nrun qpdf --help=encryption for details.\n\nEnable/disable document assembly (rotation and reordering of pages). This option is not available with 40-bit encryption."
            },
            "cleartextMetadata": {
              "type": "string",
              "description": "Don't encrypt metadata.\n\nrun qpdf --help=encryption for details.\n\nIf specified, don't encrypt document metadata even when encrypting the rest of the document. This option is not available with 40-bit encryption."
            },
            "extract": {
              "type": "string",
              "description": "Restrict text/graphic extraction.\n\nrun qpdf --help=encryption for details.\n\nEnable/disable text/graphic extraction for purposes other than accessibility."
            },
            "form": {
              "type": "string",
              "description": "Restrict form filling.\n\nrun qpdf --help=encryption for details.\n\nEnable/disable whether filling form fields is allowed even if modification of annotations is disabled. This option is not available with 40-bit encryption."
            },
            "modifyOther": {
              "type": "string",
              "description": "Restrict other modifications.\n\nrun qpdf --help=encryption for details.\n\nEnable/disable modifications not controlled by \"assemble\", \"annotate\", or \"form\". \"modifyother\":\"n\" is implied by any of the other \"modify\" options. This option is not available with 40-bit encryption."
            },
            "modify": {
              "type": "string",
              "description": "Restrict document modification.\n\nrun qpdf --help=encryption for details.\n\nFor 40-bit files, modify-opt may only be y or n and controls all aspects of document modification.  for 128-bit and 256-bit encryption, modify-opt values allow enabling and disabling levels of restriction in a manner similar to how some pdf creation tools do it. Modify-opt values map to other combinations of options as follows:  all: allow full modification (the default) annotate: \"modifyother\":\"n\" form: \"modifyother\":\"n\" \"annotate\":\"n\" assembly: \"modifyother\":\"n\" \"annotate\":\"n\" \"form\":\"n\" none: \"modifyother\":\"n\" \"annotate\":\"n\" \"form\":\"n\" \"assemble\":\"n\""
            },
            "print": {
              "type": "string",
              "description": "Restrict printing.\n\nrun qpdf --help=encryption for details.\n\nControl what kind of printing is allowed. For 40-bit encryption, print-opt may only be y or n and enables or disables all printing. For 128-bit and 256-bit encryption, print-opt may have the following values:  none: disallow printing low: allow low-resolution printing only full: allow full printing (the default)"
            },
            "forceV4": {
              "type": "string",
              "description": "Force v=4 in encryption dictionary.\n\nrun qpdf --help=encryption for details.\n\nThis option is for testing and is never needed in practice since qpdf does this automatically when needed."
            },
            "useAes": {
              "type": "string",
              "description": "Use aes with 128-bit encryption.\n\nrun qpdf --help=encryption for details.\n\nEnables/disables use of the more secure aes encryption with 128-bit encryption. Specifying \"useaes\":\"y\" forces the pdf version to be at least 1.6. This option is only available with 128-bit encryption. The default is \"n\" for compatibility reasons. Use 256-bit encryption instead."
            }
          }
        },
        "256bit": {
          "type": "object",
          "properties": {
            "accessibility": {
              "type": "string",
              "description": "Restrict document accessibility.\n\nrun qpdf --help=encryption for details.\n\nThis option is ignored except with very old encryption formats. The current pdf specification does not allow restriction of document accessibility. This option is not available with 40-bit encryption."
            },
            "annotate": {
              "type": "string",
              "description": "Restrict document annotation.\n\nrun qpdf --help=encryption for details.\n\nEnable/disable modifying annotations including making comments and filling in form fields. For 128-bit and 256-bit encryption, this also enables editing, creating, and deleting form fields unless \"modifyother\":\"n\" or \"modify\":\"none\" is also specified."
            },
            "assemble": {
              "type": "string",
              "description": "Restrict document assembly.\n\nrun qpdf --help=encryption for details.\n\nEnable/disable document assembly (rotation and reordering of pages). This option is not available with 40-bit encryption."
            },
            "cleartextMetadata": {
              "type": "string",
              "description": "Don't encrypt metadata.\n\nrun qpdf --help=encryption for details.\n\nIf specified, don't encrypt document metadata even when encrypting the rest of the document. This option is not available with 40-bit encryption."
            },
            "extract": {
              "type": "string",
              "description": "Restrict text/graphic extraction.\n\nrun qpdf --help=encryption for details.\n\nEnable/disable text/graphic extraction for purposes other than accessibility."
            },
            "form": {
              "type": "string",
              "description": "Restrict form filling.\n\nrun qpdf --help=encryption for details.\n\nEnable/disable whether filling form fields is allowed even if modification of annotations is disabled. This option is not available with 40-bit encryption."
            },
            "modifyOther": {
              "type": "string",
              "description": "Restrict other modifications.\n\nrun qpdf --help=encryption for details.\n\nEnable/disable modifications not controlled by \"assemble\", \"annotate\", or \"form\". \"modifyother\":\"n\" is implied by any of the other \"modify\" options. This option is not available with 40-bit encryption."
            },
            "modify": {
              "type": "string",
              "description": "Restrict document modification.\n\nrun qpdf --help=encryption for details.\n\nFor 40-bit files, modify-opt may only be y or n and controls all aspects of document modification.  for 128-bit and 256-bit encryption, modify-opt values allow enabling and disabling levels of restriction in a manner similar to how some pdf creation tools do it. Modify-opt values map to other combinations of options as follows:  all: allow full modification (the default) annotate: \"modifyother\":\"n\" form: \"modifyother\":\"n\" \"annotate\":\"n\" assembly: \"modifyother\":\"n\" \"annotate\":\"n\" \"form\":\"n\" none: \"modifyother\":\"n\" \"annotate\":\"n\" \"form\":\"n\" \"assemble\":\"n\""
            },
            "print": {
              "type": "string",
              "description": "Restrict printing.\n\nrun qpdf --help=encryption for details.\n\nControl what kind of printing is allowed. For 40-bit encryption, print-opt may only be y or n and enables or disables all printing. For 128-bit and 256-bit encryption, print-opt may have the following values:  none: disallow printing low: allow low-resolution printing only full: allow full printing (the default)"
            },
            "allowInsecure": {
              "type": "string",
              "description": "Allow empty owner passwords.\n\nrun qpdf --help=encryption for details.\n\nAllow creation of pdf files with empty owner passwords and non-empty user passwords when using 256-bit encryption."
            },
            "forceR5": {
              "type": "string",
              "description": "Use unsupported r=5 encryption.\n\nrun qpdf --help=encryption for details.\n\nThese options allow pages from another file to be overlaid or underlaid on the primary output. Overlaid pages are drawn on top of the destination page and may obscure the page. Underlaid pages are drawn below the destination page. Usage:  {\"overlay\"|\"underlay\"} [\"file\":\"]file\"       [\"password\":\"password]\"       [\"to\":\"page-range]\"       [\"from\":\"[page-range]]\"       [\"repeat\":\"page-range]\"       --  note the use of \"--\" by itself to terminate overlay/underlay options.  for overlay and underlay, a file and optional password are specified, along with a series of optional page ranges. The default behavior is that each page of the overlay or underlay file is imposed on the corresponding page of the primary output until it runs out of pages, and any extra pages are ignored. You can also give a page range with \"repeat\" to cause those pages to be repeated after the original pages are exhausted.  this options are repeatable. Pages will be stacked in order of appearance: first underlays, then the original page, then overlays.  run qpdf --help=page-ranges for help with page ranges."
            }
          }
        }
      }
    },
    "check": {
      "type": "string",
      "description": "Partially check whether pdf is valid.\n\nCheck the structure of the pdf file as well as a number of other aspects of the file, and write information about the file to standard output. Note that qpdf does not perform any validation of the actual pdf page content or semantic correctness of the pdf file. It merely checks that the pdf file is syntactically valid. See also qpdf --help=exit-status."
    },
    "checkLinearization": {
      "type": "string",
      "description": "Check linearization tables.\n\nCheck to see whether a file is linearized and, if so, whether the linearization hint tables are correct."
    },
    "filteredStreamData": {
      "type": "string",
      "description": "Show filtered stream data.\n\nWhen used with \"showobject\", if the object is a stream, write the filtered (uncompressed, potentially binary) stream data to standard output instead of the object's contents. See also \"rawstreamdata\"."
    },
    "rawStreamData": {
      "type": "string",
      "description": "Show raw stream data.\n\nWhen used with \"showobject\", if the object is a stream, write the raw (compressed) binary stream data to standard output instead of the object's contents. See also \"filteredstreamdata\"."
    },
    "showEncryption": {
      "type": "string",
      "description": "Information about encrypted files.\n\nShow document encryption parameters. Also show the document's user password if the owner password is given and the file was encrypted using older encryption formats that allow user password recovery."
    },
    "showEncryptionKey": {
      "type": "string",
      "description": "Show key with showencryption.\n\nWhen used with \"showencryption\" or \"check\", causes the underlying encryption key to be displayed."
    },
    "showLinearization": {
      "type": "string",
      "description": "Show linearization hint tables.\n\nCheck and display all data in the linearization hint tables."
    },
    "showNpages": {
      "type": "string",
      "description": "Show number of pages.\n\nPrint the number of pages in the input file on a line by itself. Useful for scripts."
    },
    "showObject": {
      "type": "string",
      "description": "Show contents of an object.\n\nShow the contents of the given object. This is especially useful for inspecting objects that are inside of object streams (also known as \"compressed objects\")."
    },
    "showPages": {
      "type": "string",
      "description": "Display page dictionary information.\n\nShow the object and generation number for each page dictionary object and for each content stream associated with the page."
    },
    "showXref": {
      "type": "string",
      "description": "Show cross reference data.\n\nShow the contents of the cross-reference table or stream (object locations in the file) in a human-readable form. This is especially useful for files with cross-reference streams, which are stored in a binary format."
    },
    "withImages": {
      "type": "string",
      "description": "Include image details with showpages.\n\nWhen used with \"showpages\", also shows the object and generation numbers for the image objects on each page."
    },
    "listAttachments": {
      "type": "string",
      "description": "List embedded files.\n\nShow the key and stream number for each embedded file. Combine with \"verbose\" for more detailed information."
    },
    "showAttachment": {
      "type": "string",
      "description": "Export an embedded file.\n\nShow information about the pdf file in json format. Please see the json chapter in the qpdf manual for details."
    },
    "json": {
      "type": "string",
      "description": "Show file in json format.\n\nGenerate a json representation of the file. This is described in depth in the json section of the manual. \"version\" may be a specific version or \"latest\" (the default). Run qpdf \"jsonhelp\" for a description of the generated json object.",
      "enum": [
        "1",
        "2",
        "latest"
      ]
    },
    "jsonKey": {
      "type": "array",
      "items": {
        "type": "string",
        "description": "Limit which keys are in json output.\n\nThis option is repeatable. If given, only the specified top-level keys will be included in the json output. Otherwise, all keys will be included. With \"jsonoutput\", when not given, only the \"qpdf\" key will appear in the output.",
        "enum": [
          "acroform",
          "attachments",
          "encrypt",
          "objectinfo",
          "objects",
          "outlines",
          "pagelabels",
          "pages",
          "qpdf"
        ]
      }
    },
    "jsonObject": {
      "type": "array",
      "items": {
        "type": "string",
        "description": "Limit which objects are in json.\n\nThis option is repeatable. If given, only specified objects will be shown in the \"objects\" key of the json output. Otherwise, all objects will be shown."
      }
    },
    "jsonStreamData": {
      "type": "string",
      "description": "How to handle streams in json output.\n\nWhen used with \"json\", this option controls whether streams in json output should be omitted, written inline (base64-encoded) or written to a file. If \"file\" is chosen, the file will be the name of the output file appended with -nnn where nnn is the object number. The prefix can be overridden with \"jsonstreamprefix\". The default is \"none\", except when \"jsonoutput\" is specified, in which case the default is \"inline\".",
      "enum": [
        "none",
        "inline",
        "file"
      ]
    },
    "jsonStreamPrefix": {
      "type": "string",
      "description": "Prefix for json stream data files.\n\nWhen used with \"jsonstreamdata\":\"file\", \"jsonstreamdata\":\"file-prefix\" sets the prefix for stream data files, overriding the default, which is to use the output file name. Whatever is given here will be appended with -nnn to create the name of the file that will contain the data for the stream stream in object nnn."
    },
    "updateFromJson": {
      "type": "string",
      "description": "Update a pdf from qpdf json.\n\nThe options below are useful when writing automated test code that includes files created by qpdf or when testing qpdf itself.",
      "format": "uri-reference",
      "contentMediaType": "application/json"
    },
    "allowWeakCrypto": {
      "type": "string",
      "description": "Allow insecure cryptographic algorithms.\n\nAllow creation of files with weak cryptographic algorithms. This option is necessary to create 40-bit files or 128-bit files that use rc4 encryption."
    },
    "keepFilesOpen": {
      "type": "string",
      "description": "Manage keeping multiple files open.\n\nWhen qpdf needs to work with many files, as when merging large numbers of files, explicitly indicate whether files should be kept open. The default behavior is to determine this based on the number of files.",
      "enum": [
        "y",
        "n"
      ]
    },
    "keepFilesOpenThreshold": {
      "type": "string",
      "description": "Set threshold for keepfilesopen.\n\nAdvanced control options control qpdf's behavior in ways that would normally never be needed by a user but that may be useful to developers or people investigating problems with specific files."
    },
    "noWarn": {
      "type": "string",
      "description": "Suppress printing of warning messages.\n\nSuppress printing of warning messages. If warnings were encountered, qpdf still exits with exit status 3. Use \"warningexit0\" with \"nowarn\" to completely ignore warnings."
    },
    "verbose": {
      "type": "string",
      "description": "Print additional information.\n\nOutput additional information about various things qpdf is doing, including information about files created and operations performed."
    },
    "testJsonSchema": {
      "type": "string",
      "description": "Test generated json against schema.\n\nThis is used by qpdf's test suite to check consistency between the output of qpdf \"json\" and the output of qpdf \"jsonhelp\"."
    },
    "ignoreXrefStreams": {
      "type": "string",
      "description": "Use xref tables rather than streams.\n\nThe options below tell qpdf to apply transformations that change the structure without changing the content."
    },
    "passwordIsHexKey": {
      "type": "string",
      "description": "Provide hex-encoded encryption key.\n\nProvide the underlying file encryption key as a hex-encoded string rather than supplying a password. This is an expert option."
    },
    "passwordMode": {
      "type": "string",
      "description": "Tweak how qpdf encodes passwords.\n\nFine-tune how qpdf controls encoding of unicode passwords. Valid options are auto, bytes, hex-bytes, and unicode.",
      "enum": [
        "bytes",
        "hex-bytes",
        "unicode",
        "auto"
      ]
    },
    "suppressPasswordRecovery": {
      "type": "string",
      "description": "Don't try different password encodings.\n\nSuppress qpdf's usual behavior of attempting different encodings of a password that contains non-ascii unicode characters if the first attempt doesn't succeed."
    },
    "suppressRecovery": {
      "type": "string",
      "description": "Suppress error recovery.\n\nAvoid attempting to recover when errors are found in a file's cross reference table or stream lengths."
    },
    "coalesceContents": {
      "type": "string",
      "description": "Combine content streams.\n\nIf a page has an array of content streams, concatenate them into a single content stream."
    },
    "compressionLevel": {
      "type": "string",
      "description": "Set compression level for flate.\n\nSet a compression level from 1 (least, fastest) to 9 (most, slowest) when compressing files with flate (used in zip and gzip), which is the default compression for most pdf files. You need \"recompressflate\" with this option if you want to change already compressed streams."
    },
    "externalizeInlineImages": {
      "type": "string",
      "description": "Convert inline to regular images.\n\nConvert inline images to regular images."
    },
    "iiMinBytes": {
      "type": "string",
      "description": "Set minimum size for externalizeinlineimages.\n\nDon't externalize inline images smaller than this size. The default is 1,024. Use 0 for no minimum."
    },
    "removeUnreferencedResources": {
      "type": "string",
      "description": "Remove unreferenced page resources.\n\nRemove from a page's resource dictionary any resources that are not referenced in the page's contents. Parameters: \"auto\" (default), \"yes\", \"no\".",
      "enum": [
        "auto",
        "yes",
        "no"
      ]
    },
    "addAttachment": {
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "file": {
            "type": "string",
            "description": "File to attach.\n\nthe \"addattachment\" flag and its options may be repeated to add multiple attachments. Run qpdf --help=add-attachment for details.\n\nSpecify the file for the current page operation. This is used with \"pages\", \"overlay\", and \"underlay\" and appears between the option and the terminating --. Run qpdf --help=page-selection for details.",
            "format": "uri-reference",
            "contentMediaType": "application/octet-stream"
          },
          "creationdate": {
            "type": "string",
            "description": "Set attachment's creation date.\n\nthe \"addattachment\" flag and its options may be repeated to add multiple attachments. Run qpdf --help=add-attachment for details.\n\nSpecify the attachment's creation date in pdf format; defaults to the current time. Run qpdf --help=pdf-dates for information about the date format."
          },
          "description": {
            "type": "string",
            "description": "Set attachment's description.\n\nthe \"addattachment\" flag and its options may be repeated to add multiple attachments. Run qpdf --help=add-attachment for details.\n\nSupply descriptive text for the attachment, displayed by some pdf viewers."
          },
          "filename": {
            "type": "string",
            "description": "Set attachment's displayed filename.\n\nthe \"addattachment\" flag and its options may be repeated to add multiple attachments. Run qpdf --help=add-attachment for details.\n\nSpecify the filename to be used for the attachment. This is what is usually displayed to the user and is the name most graphical pdf viewers will use when saving a file. It defaults to the last element (basename) of the attached file's filename."
          },
          "key": {
            "type": "string",
            "description": "Specify attachment key.\n\nthe \"addattachment\" flag and its options may be repeated to add multiple attachments. Run qpdf --help=add-attachment for details.\n\nSpecify the key to use for the attachment in the embedded files table. It defaults to the last element (basename) of the attached file's filename."
          },
          "mimetype": {
            "type": "string",
            "description": "Attachment mime type, e.g. Application/pdf.\n\nthe \"addattachment\" flag and its options may be repeated to add multiple attachments. Run qpdf --help=add-attachment for details.\n\nSpecify the mime type for the attachment, such as text/plain, application/pdf, image/png, etc."
          },
          "moddate": {
            "type": "string",
            "description": "Set attachment's modification date.\n\nthe \"addattachment\" flag and its options may be repeated to add multiple attachments. Run qpdf --help=add-attachment for details.\n\nSpecify the attachment's modification date in pdf format; defaults to the current time. Run qpdf --help=pdf-dates for information about the date format."
          },
          "replace": {
            "type": "string",
            "description": "Replace attachment with same key.\n\nthe \"addattachment\" flag and its options may be repeated to add multiple attachments. Run qpdf --help=add-attachment for details.\n\nThe options listed below appear between \"copyattachmentsfrom\" and its terminating \"--\".  to copy attachments from a password-protected file, use the \"password\" option after the file name."
          }
        }
      }
    },
    "removeAttachment": {
      "type": "array",
      "items": {
        "type": "string",
        "description": "Remove an embedded file.\n\nThe options listed below appear between \"addattachment\" and its terminating \"--\"."
      }
    },
    "copyAttachmentsFrom": {
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "file": {
            "type": "string",
            "description": "File to copy attachments from.\n\nthe \"copyattachmentsfrom\" flag and its options may be repeated to copy attachments from multiple files. Run qpdf --help=copy-attachments for details.\n\nSpecify the file for the current page operation. This is used with \"pages\", \"overlay\", and \"underlay\" and appears between the option and the terminating --. Run qpdf --help=page-selection for details.",
            "format": "uri-reference",
            "contentMediaType": "application/pdf"
          },
          "password": {
            "type": "string",
            "description": "Password for encrypted file.\n\nthe \"copyattachmentsfrom\" flag and its options may be repeated to copy attachments from multiple files. Run qpdf --help=copy-attachments for details.\n\nSpecify a password for an encrypted, password-protected file. Not needed for encrypted files without a password."
          },
          "prefix": {
            "type": "string",
            "description": "Key prefix for copying attachments.\n\nthe \"copyattachmentsfrom\" flag and its options may be repeated to copy attachments from multiple files. Run qpdf --help=copy-attachments for details.\n\nThese options provide tools for inspecting pdf files. When any of the options in this section are specified, no output file may be given."
          }
        }
      }
    },
    "collate": {
      "type": "string",
      "description": "Collate with pages.\n\nCollate rather than concatenate pages specified with \"pages\". With a numeric parameter, collate in groups of n. The default is 1. With comma-separated numeric parameters, take n from the first file, m from the second, etc. Run qpdf --help=page-selection for additional details."
    },
    "flattenAnnotations": {
      "type": "string",
      "description": "Push annotations into content.\n\nPush page annotations into the content streams. This may be necessary in some case when printing or splitting files. Parameters: \"all\", \"print\", \"screen\".",
      "enum": [
        "all",
        "print",
        "screen"
      ]
    },
    "flattenRotation": {
      "type": "string",
      "description": "Remove rotation from page dictionary.\n\nFor each page that is rotated using the /rotate key in the page's dictionary, remove the /rotate key and implement the identical rotation semantics by modifying the page's contents. This can be useful if a broken pdf viewer fails to properly consider page rotation metadata."
    },
    "generateAppearances": {
      "type": "string",
      "description": "Generate appearances for form fields.\n\nPdf form fields consist of values and appearances, which may be inconsistent with each other if a form field value has been modified without updating its appearance. This option tells qpdf to generate new appearance streams. There are some limitations, which are discussed in the manual."
    },
    "keepInlineImages": {
      "type": "string",
      "description": "Exclude inline images from optimization.\n\nPrevent inline images from being considered by \"optimizeimages\"."
    },
    "oiMinArea": {
      "type": "string",
      "description": "Minimum area for optimizeimages.\n\nDon't optimize images whose area in pixels is below the specified value."
    },
    "oiMinHeight": {
      "type": "string",
      "description": "Minimum height for optimizeimages.\n\nDon't optimize images whose height is below the specified value."
    },
    "oiMinWidth": {
      "type": "string",
      "description": "Minimum width for optimizeimages.\n\nDon't optimize images whose width is below the specified value."
    },
    "optimizeImages": {
      "type": "string",
      "description": "Use efficient compression for images.\n\nAttempt to use dct (jpeg) compression for images that fall within certain constraints as long as doing so decreases the size in bytes of the image. See also help for the following options:   \"oiminwidth\"   \"oiminheight\"   \"oiminarea\"   \"keepinlineimages\""
    },
    "pages": {
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "file": {
            "type": "string",
            "description": "Source for pages.\n\nrun qpdf --help=page-selection for details.\n\nSpecify the file for the current page operation. This is used with \"pages\", \"overlay\", and \"underlay\" and appears between the option and the terminating --. Run qpdf --help=page-selection for details.",
            "format": "uri-reference",
            "contentMediaType": "application/pdf"
          },
          "password": {
            "type": "string",
            "description": "Password for encrypted file.\n\nrun qpdf --help=page-selection for details.\n\nSpecify a password for an encrypted, password-protected file. Not needed for encrypted files without a password."
          },
          "range": {
            "type": "string",
            "description": "Page range.\n\nrun qpdf --help=page-selection for details.\n\nSpecify the page range for the current page operation with \"pages\". If omitted, all pages are selected. This is used with \"pages\" and appears between \"pages\" and --. Run qpdf --help=page-selection for details."
          }
        }
      }
    },
    "removeInfo": {
      "type": "string",
      "description": "Remove file information.\n\nExclude file information (except modification date) from the output file."
    },
    "removeMetadata": {
      "type": "string",
      "description": "Remove metadata.\n\nExclude metadata from the output file."
    },
    "removePageLabels": {
      "type": "string",
      "description": "Remove explicit page numbers.\n\nExclude page labels (explicit page numbers) from the output file."
    },
    "reportMemoryUsage": {
      "type": "string",
      "description": "Best effort report of memory usage.\n\nThis is used by qpdf's performance test suite to report the maximum amount of memory used in supported environments."
    },
    "rotate": {
      "type": "string",
      "description": "Rotate pages.\n\nRotate specified pages by multiples of 90 degrees specifying either absolute or relative angles. \"angle\" may be 0, 90, 180, or 270. You almost always want to use +angle or -angle rather than just angle, as discussed in the manual. Run qpdf --help=page-ranges for help with page ranges."
    },
    "setPageLabels": {
      "type": "array",
      "items": {
        "type": "string",
        "description": "Number pages for the entire document.\n\nCreate encrypted files. Usage:  \"encrypt\" \\   [\"userpassword\":\"user-password]\" \\   [\"ownerpassword\":\"owner-password]\" \\   \"bits\":\"key-length\" [options] --  or  \"encrypt\" user-password owner-password key-length [options] --  the first form, with flags for the passwords and bit length, was introduced in qpdf 11.7.0. Only the \"bits\" option is is mandatory. This form allows you to use any text as the password. If passwords are specified, they must be given before the \"bits\" option.  the second form has been in qpdf since the beginning and wil continue to be supported. Either or both of user-password and owner-password may be empty strings.  the key-length parameter must be either 40, 128, or 256. The user and/or owner password may be omitted. Omitting either password enables the pdf file to be opened without a password. Specifying the same value for the user and owner password and specifying an empty owner password are both considered insecure.  encryption options are terminated by \"--\" by itself.  40-bit encryption is insecure, as is 128-bit encryption without aes. Use 256-bit encryption unless you have a specific reason to use an insecure format, such as testing or compatibility with very old viewers. You must use the \"allowweakcrypto\" to create encrypted files that use insecure cryptographic algorithms. The \"allowweakcrypto\" flag appears outside of \"encrypt\" ... -- (before \"encrypt\" or after --).  available options vary by key length. Not all readers respect all restrictions. Different pdf readers respond differently to various combinations of options. Sometimes a pdf viewer may show you restrictions that differ from what you selected. This is probably not a bug in qpdf.  options for 40-bit only:   \"annotate\":\"[y|n]\"         restrict comments, filling forms, and signing   \"extract\":\"[y|n]\"          restrict text/graphic extraction   \"modify\":\"[y|n]\"           restrict document modification   \"print\":\"[y|n]\"            restrict printing  options for 128-bit or 256-bit:   \"accessibility\":\"[y|n]\"    restrict accessibility (usually ignored)   \"annotate\":\"[y|n]\"         restrict commenting/filling form fields   \"assemble\":\"[y|n]\"         restrict document assembly   \"extract\":\"[y|n]\"          restrict text/graphic extraction   \"form\":\"[y|n]\"             restrict filling form fields   \"modifyother\":\"[y|n]\"     restrict other modifications   \"modify\":\"modify-opt\"      control modify access by level   \"print\":\"print-opt\"        control printing access   \"cleartextmetadata\"     prevent encryption of metadata  for 128-bit only:   \"useaes\":\"[y|n]\"          indicates whether to use aes encryption   \"forcev4\"               forces use of v=4 encryption handler  for 256-bit only:   \"forcer5\"               forces use of deprecated r=5 encryption   \"allowinsecure\"         allow user password with empty owner password  values for print-opt:   none                     disallow printing   low                      allow only low-resolution printing   full                     allow full printing  values for modify-opt:   none                     allow no modifications   assembly                 allow document assembly only   form                     assembly + filling in form fields and signing   annotate                 form + commenting and modifying forms   all                      allow full document modification"
      }
    },
    "overlay": {
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "file": {
            "type": "string",
            "description": "Source for pages.\n\noverlay pages from another file on the output. Run qpdf --help=overlay-underlay for details.\n\nSpecify the file for the current page operation. This is used with \"pages\", \"overlay\", and \"underlay\" and appears between the option and the terminating --. Run qpdf --help=page-selection for details.",
            "format": "uri-reference",
            "contentMediaType": "application/pdf"
          },
          "password": {
            "type": "string",
            "description": "Password for encrypted file.\n\noverlay pages from another file on the output. Run qpdf --help=overlay-underlay for details.\n\nSpecify a password for an encrypted, password-protected file. Not needed for encrypted files without a password."
          },
          "from": {
            "type": "string",
            "description": "Source pages for underlay/overlay.\n\noverlay pages from another file on the output. Run qpdf --help=overlay-underlay for details.\n\nSpecify pages from the overlay/underlay file that are applied to the destination pages. See qpdf --help=page-ranges for help with the page range syntax. The page range may be omitted if \"repeat\" is used."
          },
          "repeat": {
            "type": "string",
            "description": "Overlay/underlay pages to repeat.\n\noverlay pages from another file on the output. Run qpdf --help=overlay-underlay for details.\n\nIt is possible to list, add, or delete embedded files (also known as attachments) and to copy attachments from other files. See help on individual options for details. Run qpdf --help=add-attachment for additional details about adding attachments. See also --help=--list-attachments and --help=--show-attachment."
          },
          "to": {
            "type": "string",
            "description": "Destination pages for underlay/overlay.\n\noverlay pages from another file on the output. Run qpdf --help=overlay-underlay for details.\n\nSpecify the range of pages in the primary output to apply overlay/underlay to. See qpdf --help=page-ranges for help with the page range syntax."
          }
        }
      }
    },
    "underlay": {
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "file": {
            "type": "string",
            "description": "Source for pages.\n\nunderlay pages from another file on the output. Run qpdf --help=overlay-underlay for details.\n\nSpecify the file for the current page operation. This is used with \"pages\", \"overlay\", and \"underlay\" and appears between the option and the terminating --. Run qpdf --help=page-selection for details.",
            "format": "uri-reference",
            "contentMediaType": "application/pdf"
          },
          "password": {
            "type": "string",
            "description": "Password for encrypted file.\n\nunderlay pages from another file on the output. Run qpdf --help=overlay-underlay for details.\n\nSpecify a password for an encrypted, password-protected file. Not needed for encrypted files without a password."
          },
          "from": {
            "type": "string",
            "description": "Source pages for underlay/overlay.\n\nunderlay pages from another file on the output. Run qpdf --help=overlay-underlay for details.\n\nSpecify pages from the overlay/underlay file that are applied to the destination pages. See qpdf --help=page-ranges for help with the page range syntax. The page range may be omitted if \"repeat\" is used."
          },
          "repeat": {
            "type": "string",
            "description": "Overlay/underlay pages to repeat.\n\nunderlay pages from another file on the output. Run qpdf --help=overlay-underlay for details.\n\nIt is possible to list, add, or delete embedded files (also known as attachments) and to copy attachments from other files. See help on individual options for details. Run qpdf --help=add-attachment for additional details about adding attachments. See also --help=--list-attachments and --help=--show-attachment."
          },
          "to": {
            "type": "string",
            "description": "Destination pages for underlay/overlay.\n\nunderlay pages from another file on the output. Run qpdf --help=overlay-underlay for details.\n\nSpecify the range of pages in the primary output to apply overlay/underlay to. See qpdf --help=page-ranges for help with the page range syntax."
          }
        }
      }
    },
    "warningExit0": {
      "type": "string",
      "description": "Exit 0 even with warnings.\n\nShell completion is supported with bash and zsh. Use eval $(qpdf \"completionbash\") or eval $(qpdf \"completionzsh\") to enable. The qpdf_executable environment variable overrides the path to qpdf that these commands output."
    },
    "jobJsonFile": {
      "type": "string",
      "description": "Job json file.\n\nMeaning of exit codes:  - 0: no errors or warnings - 1: not used by qpdf but may be used by the shell if unable to invoke qpdf - 2: errors detected - 3: warnings detected, unless \"warningexit0\" is given",
      "format": "uri-reference",
      "contentMediaType": "application/json"
    },
    "preserveUnreferencedResources": {
      "type": "string",
      "description": "Use removeunreferencedresources=no.\n\nSynonym for \"removeunreferencedresources\":\"no\". Use that instead."
    },
    "requiresPassword": {
      "type": "string",
      "description": "Silently test a file's password.\n\nSilently exit with a code indicating the file's password status:  0: a password, other than as supplied, is required 1: not used 2: the file is not encrypted 3: the file is encrypted, and correct password (if any) has been supplied"
    },
    "isEncrypted": {
      "type": "string",
      "description": "Silently test whether a file is encrypted.\n\nSilently exit with a code indicating the file's encryption status:  0: the file is encrypted 1: not used 2: the file is not encrypted  this can be used with password-protected files even if you don't know the password."
    }
  }
}
```


jberkenbilt:
I won't have time to look at this for a bit because of holiday-related activities, and I see it's still WIP, but I'm really excited to see this effort. BTW, I was thinking about the file:// URL thing, and I think modifying QUtil::safe_fopen to accept that format in addition to local files would probably a safe change and would cover almost all cases. I'm saying that from my head and not from the code, so I may retract it. Don't feel like you have to do that. When you're ready for this to be reviewed, let me know, and if you have any questions in the meantime, feel free to ask. I may have a moment here and there.

j3k0:
I'm working on improving JSON Schema types with information on option tables (bare, optional, ...). I see that "bare" options are required to be empty strings.

```json
{
  "linearize": ""
}
```

As per the config handler:

```c++
void
Handlers::addBare(bare_handler_t fn)
{
    jh->addStringHandler([this, fn](std::string const& path, std::string const& parameter) {
        if (!parameter.empty()) {
            QTC::TC("qpdf", "QPDFJob json bare not empty");
            usage(path + ": value must be the empty string");
        } else {
            fn();
        }
    });
}
```

I wasn't really willing to change the spec, but I feel like a boolean would be more appropriate:

```json
{
  "linearize": true
}
```

As for example in user code:

```js
{
   ...,
   "linerize": doesMyUserWantToLinearize()
}
```

`doesMyUserWantToLinearize` returning either and empty string or undefined doesn't feel right.

We could allow specify non-empty string like "y" | "yes" | "true"|"n"|"no"|"false" (like yn choices or some options that accept "yes" or "no" already do), the empty string defaulting to "true" would keep backward compatibility.

That's outside the original scope I had in mind, but the change seems easy:

```c++
oid
Handlers::addBare(bare_handler_t fn)
{
    jh->addStringHandler([this, fn](std::string const& path, std::string const& parameter) {
        if (parameter.empty() || parameter == "y" || parameter == "yes" || parameter == "true") {
            fn();
        } else if (parameter == "n" || parameter == "no" || parameter == "false") {
            // nothing
        } else {
            QTC::TC("qpdf", "QPDFJob json bare not empty | y | yes | true | n | no | false");
            usage(path + ": value must be either the empty string, \"y\", \"yes\", \"true\", \"n\", "
                          "\"no\", or \"false\"");
        }
    });
    jh->addBoolHandler([this, fn](std::string const& path, bool parameter) {
        if (parameter) {
            fn();
        }
    });
}
```

I guess that requires some unit tests too, thoughts? If you like I can make a different PR for that (but the JSON Schema would be impacted, so maybe just a different commit in the same PR)

jberkenbilt:
I look forward to doing a deep review. One thing to note: in qpdf's pseudo JSON schema-ish thing, any field that can be an array can also be a single item, which is intentional and intended to allow previously non-repeatable arguments to be later made repeatable without breaking the JSON contract. As such, I would expect the generated schema to use some kind of `anyOf` to allow either an array of T or a T (probably always string) in those cases. I didn't see that with a quick glance. I may have time this weekend to dig into this, but if not, I probably will the following weekend.

jberkenbilt:
Don't worry about the Linux build failure in CI. I will fix it at some point...there is a problem with the image comparison tests because of some change in ghostscript as far as I can tell. I will find time to implement a workaround.

j3k0:
> allow either an array of T or a T

Implemented with the latest commit. Didn't know about that.

m-holger:
I have been out of action and haven't had a chance to look at this in detail yet, but at a first glance this looks extremely promising.

One trivial point  though: we have standardised on `qpdf` rather than `QPDF` everywhere, including the start of sentences (except for the QPDF class).

j3k0:
> We have standardised on `qpdf` rather than `QPDF`

Do you mean in comments and documentation?

For type names ("title" in JSON Schema), I used `QPDF_type_name` so code generators converts that to classes or interfaces named `QPDFTypeName` -- if using `qpdf_type_name` as an output I get `QpdfTypeName`.

I guess I'll wait for your complete feedback and fix everything at once.

m-holger:
> Do you mean in comments and documentation?

Yes.

jberkenbilt:
I approved CI for this PR but have not had a chance to review it. I fixed the CI issue that was causing the "Linux" job to fail, so if you rebase off current main, it should pass.

jberkenbilt:
Also, another thing I'm thinking about is how we should use JSON Schema for qpdf --json-help and qpdf --job-json-help. My initial thinking is to have qpdf --json-schema and qpdf --job-json-schema, to have the JSON schema be the authoritative document, and to have the help generated from the schema or both generated from the same source. I haven't looked at the code recently and also haven't looked at your PR yet. I'm not asking you to rework anything at this time.

I'm eager to give this a deep look and appreciate the thoroughness. I will try to squeeze it in soon. I'm afraid I'm out of time for this weekend.

jberkenbilt:
I have to give some thought to the bare option thing. I feel like I'm probably okay with `true` and `false` (boolean) and probably not okay with any string other than the empty string for the reason that sometimes bare options later become options that take arguments. With the CLI, `--asdf` is the same as `--asdf=`, hence the empty string. I'm noticing just now that `qpdf --linearize=asdf` actual works. Funny how you can still notice things about your own code. Tentatively, I think I'd accept the empty string or the boolean `true` to mean yes and `null` or `false` to mean no and reject anything else. If you want to put that as a separate commit in the same PR, that would be fine. Right now, `qpdf --linearize=n` has the non-intuitive behavior of linearizing because the argument value is simply ignored. Addressing that would be a non-compatible change but one I might consider making...but I'm not prepared to say whether I'd accept a bunch of yes and no values without taking a comprehensive look at all the flags.

I have sometimes thought it would be good to make a clean break from the existing args and do something like detecting `--argv2` as the first argument and going to something completely different, but I ultimately think this would be high effort and low value. There are tons of examples floating around with qpdf syntax, and I don't think it's broken enough to try to make a radical change to it. Sometimes there is more value to stability than to things being perfectly clean.

m-holger:
> I have to give some thought to the bare option thing. I feel like I'm probably okay with true and false (boolean) and probably not okay with any string other than the empty string for the reason that sometimes bare options later become options that take arguments. With the CLI, --asdf is the same as --asdf=, hence the empty string. I'm noticing just now that qpdf --linearize=asdf actual works. Funny how you can still notice things about your own code. Tentatively, I think I'd accept the empty string or the boolean true to mean yes and null or false to mean no and reject anything else. 

I am not sure the CLI should accept `--linearize=yes` as an alternative to `--linearize`.

It makes perfect sense to me that job JSON should accept a boolean instead of an empty string / missing key to encode  the presence/absence of a flag. This is likely to be much more convenient and less error prone for programmers generating job JSON.

 For the CLI  this makes much less sense:

- it is an  idiosyncratic interpretation of a flag.

- if we accept the string `yes` in the CLI, do we also accept the string `"yes"` in job JSON as an alternative to the empty string or the boolean `true`?

- if we accept `false` as an alternative to the absence of a flag, to we also use it as an alternative to the absence of an option, i.e. is `--decode-level=false` as an alternative to `--decode-level=generalized` or `--decode-level=none` depending on context? What about if `--decode-level` had started out as a flag before becoming an option? What about if `no` is a valid option, but not the default?

To me, `--linearize= ` feels acceptable as an user error that is not traped by qpdf for convenience, but `--linearize` is the only strictly correct use of the CLI.

m-holger:
Also, we should keep consistency with existing boolean options such as `--compress-streams=[y|n]` in mind.

jberkenbilt:
> I am not sure the CLI should accept `--linearize=yes` as an alternative to `--linearize`.

It shouldn't. I consider it a bug. Sorry if I wasn't clear about that. I don't think fixing it would be hard. I created https://github.com/qpdf/qpdf/issues/1329

To be a little more clear, in the event that a previously bare argument starts accepting options (this has happened several times before, e.g. `--split-pages` used to be bare, and now it takes an optional parameter specifying the side of the page groups), the empty string would still make sense as the value that indicates the default. So the rationale behind empty string being the thing that specifies a bare argument is that it is future-compatible. In order to maintain that, we'd need the following behavior if we are going to support booleans:

* For bare options, the empty string and `true` (the boolean, not the string) select the option. `null` and `false` are equivalent to not having the option.
* For options with optional parameters, `true` would be the default, and `false` or `null` would be the same as not specifying it.

When you look at future-compatibility allowing for things moving from bare to optional parameter, it makes accepting booleans less of a sure thing and rules out the whole yes|y thing as I see it.

Over the years, I have developed habits that increase my flexibility to make changes without breaking backward compatibility. Some might argue that I place too much emphasis on backward compatibility, but I have been privately thanked many times for not breaking stuff with qpdf, and I think it's nice when a script that works one year still works a few years later. While I agree that booleans for bare options is more intuitive, I'm not sure it's intuitive for optional parameters.

Given the above, do you still feel that accepting `true` as the same as having the default value of an optional parameter is sensible and worth it to support it with bare options? If so, I'd consider that; if not, I think we need to stick with the empty string. As I write this, I feel like I'm leaning toward continuing to support the empty string and fixing the bug that the CLI accepts any value without validating it, at least on the CLI. Yes, I realize that technically breaks backward compatibility, but I'm not too worried about breaking something that's goofy and undocumented. https://xkcd.com/1172/

I would say that it's worth a note in the docs about the rationale behind the empty string being used for bare options.

m-holger:
> Given the above, do you still feel that accepting true as the same as having the default value of an optional parameter is sensible and worth it to support it with bare options? 

It only makes sense for flags or where for backward compatibility we have to accept an empty string as a valid option. I think it would be useful for users, but at the moment wouldn't be high on my priorities. Perhaps something for 2027.

jberkenbilt:
@j3k0 Let's skip the boolean and stay with the empty string. We can document the rationale, explaining that job JSON is an encapsulation of command-line arguments, and that the value corresponds to the parameter after `=`. When we update the docs, we should be consistent with the language that I use in the docs regarding option, argument, parameter, flag, etc. There is a comment at the top of manual/cli.rst that summarizes the editorial conventions.

j3k0:
As you say, the issue comes from the fact that this format is just a direct encapsulation of command line options.

In particular, for bare options:

1. Some are operators / filters / modifiers to add to qpdf pipeline (like `--split-pages`) 
2. Some are actual flags (like `--password-is-hex-key`) which could (should?) be have been `y|n` choices.

So you're right, maybe only for flags a boolean would make sense. But not all bare options are flags.

Operators may eventually later take a "value" (a unnamed argument to that operation). Although, if that happens, adding another CLI options to provide the value for that argument is a possibility that offers some advantages:
- it's clear what that value is: `--split-pages --pages-per-file=3` vs. `--split-pages=3`
- if the operation later accepts more than one parameter, it can be done without requiring a custom syntax to provide multiple values `--do-something=3:left:12` vs. `--do-something --something-x=3 --something-side=left --something-limit=12`)
- but it's more verbose.

The way operators are added to JSON is probably not how it would have been if you were to design that JSON Schema for a tool that didn't already exist. I can imagine a way to specify a pipeline with each operation and their optional arguments, like the overly long example below. Someone hard to express with command line arguments.

```js
{
  "pipeline": [
    { "action": "loadPDF", "filename": "my-file.pdf" },
    { "action": "splitPages", "pagesPerFile": 2, "pattern": "pages-%d.pdf" },
    { "action": "showPages", "withImages": true, "inFile": "pages.txt" },
    { "action": "writeQDF", "filename": "my-file.qdf" },
    { "action": "encrypt", "username": "xxx", "password": "yyy" },
    { "action": "linearize" },
    { "action": "writePDF": "filename": "linear.pdf" }
  ]
}
```

Sorry for the maybe unproductive random thoughts, to just conclude that I'll change the schema to only accept empty strings for bare options. For the above, if I ever need a high level "cleaner looking" JSON interface, then it might have to be a new layer on top of that one -- which might require more than 1 invocation to qpdf anyways, to overcome some limitations.

jberkenbilt:
> Sorry for the maybe unproductive random thoughts, to just conclude that I'll change the schema to only accept empty strings for bare options. For the above, if I ever need a high level "cleaner looking" JSON interface, then it might have to be a new layer on top of that one -- which might require more than 1 invocation to qpdf anyways, to overcome some limitations.

These are not unproductive thoughts. They are good, helpful collaboration. :-) QPDFJob was a first iteration and was basically as far as I could go in the amount of time I had when I built that feature. I had originally wanted to be able to chain jobs together, and, as I have mentioned, I would like to do something about qpdf's CLI, but it's too low value compared to the many other things people would like to have and I (and/or @m-holger and/or anyone else who steps up) would like to build. Software development is always full of trade-offs, and open-source development, especially on projects with a small number of volunteer maintainers, is even more so. I think we're settling on a good balance of pragmatism and functionality. Let's get this out and see what we learn by using it. I basically agree with all your points. Perhaps it's a weakness, but I have become quite conservative about what I let in because I feel like I and/or other maintainers have to support it indefinitely. @m-holger can attest to how long it took before I let go enough to let him start making decisions at some level.

Anyway, I like where we're going, and I'm certain that this will get merged in some form. I like perfection, but, with fear of sounding trite, we have to not let perfect be the enemy of good. Let's make incremental improvements to what we have and do so in a way that doesn't box us in, then see what we learn. Anyway, sounds like we're all aligned on this now. Thanks for your meaningful and helpful participation.

j3k0:
Notice an "anonymous titles" option is implemented, not sure if I like that.

```patch
diff --git a/generate_auto_job b/generate_auto_job
index c782690e..ac931686 100755
--- a/generate_auto_job
+++ b/generate_auto_job
@@ -244,6 +244,9 @@ class Main:
         mxg.add_argument('--generate',
                          help='generate files from sources',
                          action='store_true', default=False)
+        parser.add_argument('--json-schema-with-anonymous-enums',
+                         help='generate JSON schema with anonymous enums',
+                         action='store_true', default=False)
         return parser.parse_args(args)
 
     def top(self, options):
@@ -275,7 +278,7 @@ class Main:
[...]
```

jberkenbilt:
I'm hoping this will be the weekend I can dive into this. Sorry it takes me so long to get time for things.

jberkenbilt:
> I'm hoping this will be the weekend I can dive into this. Sorry it takes me so long to get time for things.

Alas, out of time. Sorry for the noise. I just want you to know I haven't forgotten. Turns out I had a whole lot of things I had to get done this weekend.

generate_auto_job
```diff
@@ -943,6 +1000,457 @@ A complete manual can be found at https://qpdf.readthedocs.io.
...
+            "type": "object",
+            "$defs": defs,
+            **obj
+        }
+
+        # Write schema to file
+        with write_file(self.DESTS["schema_json"]) as f:
+            json.dump(json_schema, f, indent=2)
+
+    def _is_file_path(self, job_data, full_key, key):
```
jberkenbilt, line 1452: Is this needed?

generate_auto_job
```diff
@@ -943,6 +1000,457 @@ A complete manual can be found at https://qpdf.readthedocs.io.
...
+            if obj[0] is None:
+                ret['type'] = ['array', 'string']
+                ret['items'] = { "type": "string" }
+            else:
+                object_type = self._convert_to_schema(self._concat_keys(parent_key, key), '[]', obj[0], data, defs, help_texts)
+                if object_type.get('description'):
+                    object_type.pop('description')
+                def_key = f"QPDF_{key}".replace('-', '_')
+                while def_key in defs:
+                    print(f"WARNING: {def_key} already defined")
```
jberkenbilt, line 1233: should be an error?

generate_auto_job
```diff
@@ -943,6 +1000,457 @@ A complete manual can be found at https://qpdf.readthedocs.io.
...
+        Fallback to the long help extracted from cli.rst.
+
+        The cleanup consists in:
+        - Replace command line options with JSON-style notation.
+          (except for --help[=value] which should not be replaced)
+        - Remove the option syntax part if it exists.
+        - Format lists.
+        - Ensure proper capitalization and a final dot.
+        """
+        if not long_text is None:
```
jberkenbilt, line 1018: `if long_text is not None`

generate_auto_job
```diff
@@ -943,6 +1000,457 @@ A complete manual can be found at https://qpdf.readthedocs.io.
...
+                "is_bare": is_bare,
+                "is_opt_param": is_opt_param,
+                "is_opt_choice": is_opt_choice,
+                "is_req_choice": is_req_choice
+            }
+    
+        kind = field_metadata(key, value, data, help_texts)
+        desc = help_texts.get(self._json_schema_key(key))
+        full_key = self._concat_keys(parent_key, key)
+        # comment = f"full_key:{full_key} key:{key} json_key:{self._json_schema_key(key)}"
```
jberkenbilt, line 1301: Should this `$comment` (rather than `comment`)? Also, I'm thinking these should either be included, or the commented-out code should be removed.

jberkenbilt (changes requested):
This is very good initial work. I have a few nitpick comments. Also, there are a number of PEP-8 warnings around indentation and line length, and there is trailing white space. Since this is not primarily a Python project,  I don't have anything automated around this. It's not a deal-breaker, but it would be better if the Python code were closer to PEP-8 clean.

I'm inclined to take this after these issues are corrected and the commits are squashed to one commit. I have some misgivings, but I think having this as it is is better than not having it. My misgivings are:

* If I could go back in time, maybe I would just use JSON schema instead of rolling my own validation format. I'm tempted to say I'd rather have the schema be authoritative and the other parts be derived from the schema, but that would be a lot of work, and I'm not sure it would be practical, so I think I'm good with the pragmatic choices of generating the schema from the other sources as you have done.
* There are a lot of special cases and workarounds for formatting and so forth, and it seems fragile. It has the mark of being an external contribution that views what's there is fixed. That said, I appreciate the light touch -- only modifying this one file makes it a lot easier to review and accept. If there are specific assumptions that the schema generation code makes that would break things if they were violated, it might be good to add checks. Don't go overboard though.
* There needs to be some validation of the schema after it's generated. It doesn't have to be rigorous. Perhaps there is a simple Python way to validate the schema? I would want a situation where generating an invalid schema would cause the build to fail. Otherwise, we run the risk of doing something like releasing with a broken schema. This _could_ be a test in CI, but more rapid feedback would be better. If doing this requires additional Python modules that are not in the standard library, it could be possible to change generate_auto_job to a bash script that detects whether the module is available and, if not, loads them. This could be done with a wrapper like below. We'd rename generate_auto_job to generate_auto_job.py, and make a generate_auto_job bash script like this, where yaml/pyyaml would be replaced with whatever we need to do the schema validation.  

```bash
#!/bin/bash
python=python3
if ! $python -c 'import yaml' >/dev/null 2>&1; then
    $python -mvenv venv
    python=venv/bin/python3
    venv/bin/pip install --upgrade pip
    venv/bin/pip install pyyaml
fi
exec $python $(dirname $0)/generate_auto_job.py ${1+"$@"}
```

If we did all that, it would largely mitigate my concerns about broken json schema generation. Ideally, we could create a little job in CI that uses the schema in a non-trivial way, but if the generation validates it, I think I would let that go unless/until we had a problem that it would have prevented. Mostly I think this would be pretty stable, and we don't need to over-engineer it.

PR #1334, opened 2025-01-20
Reset main for qpdf 11

jberkenbilt:
@m-holger This is the contents of your qpdf-11 branch as a single commit relative to main. It reverts main's contents without reverting the PRs. I have also pushed `next` to my fork.

If you agree with this approach, we can fast-forward main to this PR and directly push jberkenbilt/next to qpdf/next.

If you would still prefer to revert your PRs and reintroduce the PRs as individual commits, we can proceed with that approach. I think this approach has the advantage of being simpler and duplicating less history, but it has the disadvantage of being temporarily misleading about which features are in main. I'm not as concerned about this though since it's just your internal work and not specific feature work that changes functionality in a way people are likely to specifically look for. We can have one last discussion about approach before deciding whether to revert your PRs so we can release qpdf 11.10 from main. There are a few things I'd still like to get in for 11.10, but I might consider releasing without those and doing 11.10.1 or 11.11.0 with the other changes (such as WASM/json schema).


jberkenbilt:
Wait, I didn't do this right. I will try again.

jberkenbilt:
There, I did it again but i a way that maintained the individual commits on your qpdf-11 branch. I think this creates the clearest git history. I'm not sure why readthedocs is failing. qpdf does what the readthedocs blog says it's supposed to do, and the build failure says a key is missing from readthedocs.yaml and then proceeds to show that the key is present. I am assuming it's a problem on their side, but I will investigate if it doesn't fix itself by the time we're ready to move forward.

jberkenbilt:
If we go this way, you can change the target of your unmerged pull requests to the `next` branch once I push jberkenbilt/next to qpdf/next.

jberkenbilt:
@m-holger I plan on merging this in front of other work. I don't think I'll have anything else to merge today, but either way, I will plan on merging this soon unless you have any major objection to this approach. My plan will be to merge this and then push my next branch to the main repo's next branch where it will get CI. Then you can retarget your PRs to it or come up with some other strategy to stage changes into main, etc. We may need a more comprehensive plan. Some of what's planned for 12 may slip to 13 as there are some ABI-breaking or changes I want to get in. Also PointerHolder which is not technically ABI breaking but may require people to change their builds, so I would not want to do it on a minor version. We can have enthusiastic testers test off next at specified increments or prepare a moving nightly release off next or similar....we can discuss options.

m-holger:
@jberkenbilt , I think reverting PRs as in my qpdf-11-revert branch gives better results for git blame, but maybe I am missing something.  

My qpdf-11-11 branch  reapplies the PRs. If i look at QPDF_objects.cc, git blame tracks back all the way back to QPDF.cc to the beginning of qpdf. If I try the same with the next branch, QPDF_objects.cc is treated as an entirely new file with no link back to the history in QPDF.cc.

jberkenbilt:
Please take care of it in your preferred way. I don't think we will lose git blame my way, but I think it's more important that you are comfortable with it, and I think your way will probably create clearer history. I'll close this and delete my `next` branch. I would like to get `main` releasable though. As an alternative, I am also willing to do remaining releases directly from the qpdf-11 branch and not try to get main back as the release branch until your pending changes are done. We have way too many changes in flight simultaneously though. There are something like 25 open PRs. I personally have a rule to keep the number of simultaneously open work branches to <= 2.

PR #1338, opened 2025-01-27
Exclude #1272, #1289, #1297 and #1301 from qpdf 11.10 

m-holger:


jberkenbilt (commented):
Looks good to me -- it matches my previous branch except for a few fuzz changes that are in yours and not mine and should be there, so this is good to go if you want. Thanks for jumping on this.

PR #1341, opened 2025-01-31
Update copyrights

m-holger:


jberkenbilt:
Just a quick sanity check...does the `update copyright notices` in `RELEASE PREPARATION` in README-maintainer.md still work? Probably. Not that worried about it.

m-holger:
I didn't try it, but I did consider this when I was doing the updates and didn't spot anything that would require the instructions to be changed.

jberkenbilt:
> I didn't try it, but I did consider this when I was doing the updates and didn't spot anything that would require the instructions to be changed.

Good enough. We'll try it in 2026.

PR #1342, opened 2025-02-01
CLI reject flags with parameters (fixes #1329)

### LINKED ISSUE #1329: bare CLI arguments should not allow values
* `--linearize=asdf` etc. should be an error
* make sure other than the empty string isn't accepted in job JSON either

m-holger:


libqpdf/QPDFArgParser.cc
```diff
@@ -9,6 +9,8 @@
 #include <cstring>
 #include <iostream>
 
+using namespace std::literals;
```
jberkenbilt, line 12: New to me. Interesting.

jberkenbilt (approved):
This is the fix I was expecting. Thanks for taking care of it.

PR #1345, opened 2025-02-02
Fix qdf output file

jberkenbilt:
Allow fix-qdf to accept a second argument as the output file helps in environments like PowerShell 5 where binary standard output is mangled even with proper handling by the writing executable. This also fixes documentation, which was never updated when fix-qdf was changed to accept an input file as a parameter.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2025-02-02  Jay Berkenbilt  <ejb@ql.org>
+
+	* Have fix-qdf accept a second argument, interpreted as the output
+	file. Fixes #1330.
+
 2024-02-01  M Holger  <m.holger@qpdf.org>
 
 	* Bug fix: in qpdf CLI / QPDFJob throw a QPDFUsage exception if a
--- a/manual/fix-qdf.1.in
+++ b/manual/fix-qdf.1.in
@@ -3,9 +3,11 @@
 fix-qdf \- repair PDF files in QDF form after editing
 .SH SYNOPSIS
 .B fix-qdf
-< \fIinfilename\fR > \fIoutfilename\fR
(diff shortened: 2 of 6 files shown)


jberkenbilt:
@m-holger Feel like reviewing in case I did something goofy with my C++?

m-holger:
Looks fine to me.

My only very minor comment would be that as a habit we should avoid std::endl in favour of "...\n" or '\n'. It makes no difference in this case, but generally avoiding std::endl makes it less likely that we accidentally use it where it hurts performance.

jberkenbilt:
Interesting. I almost always favor endl when potentially writing to a terminal since it flushes, which makes it easier if people tail -f their output or worry about interleaving stdout and stderr. Anyway, thanks.

m-holger:
std::endl should not be necessary in those situations. std::cerr is unbuffered, and any reads from std::cin or writes to std::cerr cause std::cout to be flushed.

jberkenbilt:
Based on my research, you are correct, and it has always been this way. I stand corrected. I will favor "\n" going forward. Maybe we should remove std::endl in places where it appears since it hurts readability. I guess this is low-priority, but now that I know this, I wish we didn't have std::endl all over the place. I guess I've just had this wrong for all these years.

In an ironic twist, I will be working on a C++ project at work for a while. It will be the first time I've coded in C++ for more than ancillary purposes in my day job since around 2007. Perhaps this will help me get myself back up to date so I can stop coding C++ in an outdated fashion.

m-holger:
> I guess I've just had this wrong for all these years

Probably a hang-over from C. My advantage is that I hate C and C++ and have always tried to avoid it like the plague - that leaves me with a lot less historical baggage.

> I guess this is low-priority

Agree, something for new code or significant rewrites, but not worthwhile on its own.

PR #1346, opened 2025-02-03
Zopfli

jberkenbilt:


jberkenbilt:
Also: consider having the zopfli tests separate from quick tests since they're not particularly quick...check successful run for timing.

jberkenbilt:
Re code coverage, the zopfli lines are covered in tests, but the CodeCov test runs without zopfli, so it doesn't count toward coverage.

jberkenbilt:
For the record, I have consciously deviated from a policy I have to avoid compile-time driven behavior changes. This is not unprecedented as we also have a compile-time flag for avoiding file handles as well as checks for various crypto implementations. Adding "feature flags" like this can create an explosion in the number of ways qpdf can be configured.

I think this is justified in this case because use of zopfli is fairly specialized but has been requested many times over the years. Also, the changes are very localized. Other than documentation and sanity checks on runtime configuration, the changes to the code are confined to `Pl_Flate`. Additionally, I have advised against packagers building for zopfli, which means that this, like other "feature flags", is something that would be used by people with specialized needs who are willing to build qpdf themselves.

jberkenbilt:
@m-holger Interested in looking? If this passes CI, it is ready to go. I'd be interested in your feedback on the feature and the way I decided to implement it. There have been multiple requests for this over several years.

m-holger:
A couple of comments.

- I think it is easier and more efficient to implement zopfli_buf as a std::string rather than Pl_Buffer. Under the hood Pl_Buffer is a std::string anyway, and the write method is just a string append.

- Its probably best to call zopfli_active in the constructor to fail as early as possible.

- zopfli_supported can be made a constexpr bool method. Given that qpdf will normally be build without zopfli support and that the write method is called very frequently, it is probably worth surrounding the zopfli code in it with a if constexpr (zopfli_supported()) { ... }

- zopfli-active always returns the same result. It is probably worthwhile to cache the result.

m-holger:
Thinking about it a bit more, do we actually need the environment variable `QPDF_ZOPFLI`? `--zopfli` could simply enable zopfli by calling a static Pl_Flate method if compiled in and available, or deal with the zopfii unavailability, either through an error or by issuing a warning a falling back to the default implementation.

jberkenbilt:
> * I think it is easier and more efficient to implement zopfli_buf as a std::string rather than Pl_Buffer. Under the hood Pl_Buffer is a std::string anyway, and the write method is just a string append.

Okay. That's one of the things you rewrote, so I didn't have that in my head, though I remember now that you mention it.

> * Its probably best to call zopfli_active in the constructor to fail as early as possible.

You must be looking at an old version. There's no more zopfli_active. I could probably move it back to the constructor. I originally had it in the constructor, but it was failing sanitizer tests, though an earlier implementation had the possibility of that throwing an exception. The way it is now, the exception is not thrown from Pl_Flate's normal flow, so I can probably change it back. That said, in practice, I don't think it really makes much of a difference. But I coded it in the constructor originally, and I agree it's better there. I'll try again.

> * zopfli_supported can be made a constexpr bool method. Given that qpdf will normally be build without zopfli support and that the write method is called very frequently, it is probably worth surrounding the zopfli code in it with a if constexpr (zopfli_supported()) { ... }

It can't. I originally did this true, but that makes it have to be inline, and since Pl_Flate.hh is part of the public API, it can't access compile-time constants from config.h.

> * zopfli-active always returns the same result. It is probably worthwhile to cache the result.

To me, that's an over-optimization, but that's why I asked for your input. I can definitely do that.

> Thinking about it a bit more, do we actually need the environment variable `QPDF_ZOPFLI`? `--zopfli` could simply enable zopfli by calling a static Pl_Flate method if compiled in and available, or deal with the zopfli unavailability, either through an error or by issuing a warning a falling back to the default implementation.

With zopfli enabled, qpdf is VERY SLOW. I think this is something that needs to be explicitly requested, even if compiled in. Building qpdf should zopfli support should not make qpdf use zopfli by default. Per their docs, zopfli is 100x slower and produces 5% better output than the best compression otherwise. It's okay for archival use, but not for normal use. I definitely feel that it has to be turned on with an environment variable.

jberkenbilt:
@m-holger I pushed up a new commit. If you like this better and it passes CI, I will squash and merge. I kept it separate for now until it passes CI and to make it easier for you to see what I changed, considering that you may have made your comments before I pushed up changes that I made this morning.

jberkenbilt:
@m-holger Thanks for the review. Much appreciated.

m-holger:
>  can't access compile-time constants from config.h

missed that

> I definitely feel that it has to be turned on with an environment variable

agree, wise choice

Good to go as far as I'm concerned.

PR #1347, opened 2025-02-03
Update cli.rst adding vital "PDF" words

jidanni:
(So folks working offline won't go for hours dreaming they can overlay *.txt files, until after the finish writing the txt file, and try it, and it doesn't work.)

### DIFF
--- a/manual/cli.rst
+++ b/manual/cli.rst
@@ -1522,10 +1522,10 @@ Related Options
 
    .. help: begin overlay options
 
-      Overlay pages from another file on the output.
+      Overlay pages from another PDF file on the output.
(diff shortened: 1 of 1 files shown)


m-holger:
Looks like the manual could have been clearer.

@jidanni thanks for helping to improve the manual.

jberkenbilt:
Changes to cli.rst without updating jobs.sum won't pass CI @m-holger  -- when I get these, I run generate_auto_jobs locally and merge.

jberkenbilt:
@m-holger Thanks for merging -- I agree, this is a good update.

PR #1350, opened 2025-02-05
Apply sanity checks on SF_FlateLzwDecode parameters

m-holger:


m-holger:
@jberkenbilt , could you sense-check the 3rd commit for me

PR #1359, opened 2025-02-10
Fix build against a standard zopfli install

mgorny:
Fix the logic to accept a top-level `zopfli.h` header, as that is the location used by upstream's build system.


jberkenbilt:
Thanks for this -- I'll give a few days for other issues to come in, then will release 11.10.1 with this fix.

mgorny:
Thanks!

PR #1363, opened 2025-02-14
Test mingw CI

m-holger:


m-holger:
@jberkenbilt ,

I didn't expect the CIFuzz failure - the failure occurs building zlib deflate. From first glance, version is 1.3.1.1, which is the same version successfully build for 1359.

The mingw build looks entirely normal up to the running of the tests, where we get
```

test 2
    Start 2: compare-for-test

2: Test command: C:\msys64\usr\bin\perl.exe "D:/a/qpdf/qpdf/run-qtest" "--disable-tc" "--top" "D:/a/qpdf/qpdf" "--bin" "D:/a/qpdf/qpdf/build/compare-for-test" "--bin" "D:/a/qpdf/qpdf/build/libqpdf" "--code" "D:/a/qpdf/qpdf/compare-for-test" "--color" "ON" "--show-on-failure" "ON" "--tc" "D:/a/qpdf/qpdf/compare-for-test/*.cc"
2: Working Directory: D:/a/qpdf/qpdf/build/compare-for-test
2: Test timeout computed to be: 10000000
2: qtest-driver: can't canonicalize path to bindir D: No such file or directory
2/7 Test #2: compare-for-test .................***Failed    1.38 sec
```
in line 2763, with similar results for all remaining tests. I don't have a Windows system set-up to run CI locally to investigate further.

jberkenbilt:
Windows failed on main during its scheduled build. This is on my radar. Looks like the github runner has the tools installed with a different drive letter from the source, and there's no way to construct a relative path across drive letters in Windows. I'll have to tweak qtest's perl code to work around it.

jberkenbilt:
In other words, disregard the Windows failures for now.

PR #1367, opened 2025-02-19
Discontinue ChangeLog; update release notes as you go

jberkenbilt:
@m-holger What do you think? If you agree with this, feel free to merge. I think it's a waste of time and energy at this point to keep multiple ways of tracking changes. This means the release notes may be slightly more detailed. I think we should get into the habit of updating the release notes as we go. It makes main more useful, and it makes it easier at release time.


m-holger:
Agreed.

PR #1369, opened 2025-02-20
Deprecate QPDFTokenizer push mode

m-holger:
Push mode is not used by qpdf and is not tested. It is difficult to see a legitimate use for it.

Continuing to support it creates unnecessary constraints and effort when adapting the tokenizer.


jberkenbilt (approved):
agreed

PR #1372, opened 2025-02-23
Refactor QPDFObject / QPDFObjectHandle

m-holger:
The aims of this PR are:

1. use std::variant in QPDFObject instead of std::share_pointer for efficiency (both runtime and memory consumption)

2. restrict the use of plain std::shared_pointer<QPDFObject>  and direct use of QPDFObject or the various QPDF_Type classes for a cleaner internal API, without significant performance impact

3. allow the development of typed object handles without breaking ABI or unnecessary code duplication     

**Part 1**

To support (3), I have added a new class BaseHandle as a common base class for all object-handle-like classes, including the current object helper classes and future typed object handles. The class is only usable as a base class and reduces code duplication.

The aim is to move relevant private QPDFObjectHandle methods and make them available as BaseHandle protected methods. Similarly, QPDFObject methods and QPDFValue methods common to all value types will be moved to BaseHandle and exposed either as protected methods or private API methods, where access is controlled by implementing them in the private QPDFObjectHandle_private.hh file. Ultimately , direct access to QPDFObject and the QPDFValue types should be restricted BaseHandle, typed object handles, QPDFParser and the future QPDF::Objects, with QPDFObject and the QPDFValue types treated as pure data without methods.

Public API BaseHandle methods are operators to convert derived classes to QPDFObjectHandle, QPDFObjGen and bool,  as well as a new method null. All types derived from BaseHandle have an underlying object handle and require frequent access to it. They are also frequently passed to functions by reference to their obj_gen. Providing conversion operators reduces boiler-plate code, the need for convenience function / method overloads and most importantly, the frequent need to maintain 3 different variables for the same object, which makes code more difficult to understand and maintain.  The new null methods supports the interpretation of uninitialised objects as immutable nulls to act effectively as a replacement for isNull. On reflection, I think we should cancel the deprecation of isInitialized and maintain isInitialized and isNull indefinitely as is. There is no real cost to this and it keeps existing programs using qpdflib working without need for updates.

The public API BaseHandle methods all are (or will be) trivial inline  functions. I have not marked them with QPDF_DLL for export. Is there any reason to do this? 

All remaining BaseHandle methods are part of the private API. My intention is to only selectively expose them once the API is fully stable and proven. Given that BaseHandle is predominantly private, I have placed it in a separate header file included from QPDFObjectHandle. This will need to be reviewed when substantial parts become part of the public API.
 
As part of deriving QPDFObjectHandle from BaseHandle I have made QPDFObjectHandle final. Given that QPDFObjectHandle does not have any virtual members and no virtual destructor I cannot see how it could be subclassed in a meaningful and safe way, and therefore it should be made explicit that this is not supported. I think the same applies to other similiar classes.

I have also derived QPDFObjectHelper from BaseHandle, and to support this I have replaced the oh member with a pair of accessors. However, I think more substantial change is justified:

- I don't think there is ever any reason to upcast an object helper to QPDFObjectHelper. If this is correct, the destructor should be made non-virtual protected, and there is no need to export the class.

- I am not sure that most object helpers can be sensibly sub-classed, in which case they should be made final with non-virtual destructors. 

**Part 2**

The second part of the PR moves type-specific methods of QPDFValue subclasses  into new BaseHandle subclasses Array, BaseDictionary, Dictionary, Name and Stream. The primary purpose of this change is to prepare for merging QPDFValue into QPDFObject and using std::variant to hold type-specific data rather than using a shared pointer to QPDFValue. However, even without this driver the change makes sense. Logically, Arrays, Dictionaries, etc, are instances of object handles and therefore would normally be implemented as subclasses of QPDFObjectHandle. Methods to manipulate these objects require access to methods common to all object handles, and implementing them at a level below object handle requires them to be broken down in a way that requires multiple function calls across compilation unit boundaries. More importantly, the container classes require access to object information such as the owning QPDF object which is not not available to e.g. a plain  QPDF_Dictionary object.  
 
In order to give compilers maximum opportunity for optimisation, I have shifted method implementations around, grouping them by value type rather than by C++ class. For container types, related methods have been moved to QPDF_Array.cc, QPDF_Dictionary.cc and QPDF_Stream.cc . Scalar type methods and methods applicable to QPDFObjects have been moved to QPDFObjectHandle.cc .

**Part 3**

The third part is to change QPDFObject to use std::variant instead of std::shared_ptr. The key changes are:
- move QPDFValue data members and inner classes to QPDFObject.
- move QPDFValue virtual methods to QPDFObject, with the aim to eventually move them to BaseHandle. The methods are typically implemented in QPDFObjectHandle.cc as methods consisting essentially of a single switch statement depending on type code.
- eliminate QPDFValue and make its subclasses into stand-alone data-only classes defined in QPDFObject_private.hh, with accesses allowed by specified friend classes only.

There are two complications. Firstly, using std::variant means that objects of all value types occupy the same amount of space. This is generally not an issue as the difference in size between the common value types is small, and the amount of waste is small compared to space saved by eliminating a shared pointer (including control block) and the vtable pointer. The exception are stream objects, which are significantly larger while also being somewhat less common. This makes it worthwhile to change QPDF_Stream to a unique_ptr to the actual data.

The second complication arises from QPDF::replaceObject creating objects that share their QPDFValue with the object passed in as replacement. This is the only current case where there is a 'many to 1' relationship between QPDFObject and QPDFValue. To handle this case a new QPDF_Reference type has been added. The type is currently only used for replacement objects as using them more generally would have a significant performance impact. A future use will be to represent objects that are an indirect reference to another object.

**Part 4** 

The final part of the PR is to make Array and Dictionary iterable and use them internally by qpdf to iterate over arrays and dictionaries, without the need to copy their content except for sparse arrays. This also includes changing QPDF_Array to use vectors of QPDFObjectHandle rather than QPDFObject as part of the effort to limit the direct use of QPDFObject.


m-holger:
@jberkenbilt , could you review the first 2 commits when you get a chance. Thanks

m-holger:
@jberkenbilt , I finished my summary which hopefully makes sense and allows you to review the PR. I will do the change log summary after your feed-back.

jberkenbilt:
I'll try to get to it this weekend, but I think we should not rush on 12.

jberkenbilt:
I have carefully read your description. I will make an effort to review this over the next day or two.

include/qpdf/QPDFObjectHandle.hh
```diff
@@ -1610,6 +1606,22 @@ class QPDFObjectHandle::QPDFArrayItems
     QPDFObjectHandle oh;
 };
 
+namespace qpdf
```
jberkenbilt, line 1598: (no action required on this comment)

It took me a while to figure out what was going on here. I guess these have to be in this header instead of ObjectHandle.hh because they access private constructors. I didn't even know you could do this.

I'm okay with the use of the `qpdf` namespace. C++ didn't have namespaces when I first made qpdf in 2002, and I didn't switch to use them in 2008 for the first public release, unfortunately, but there's no reason we can't use them now.

include/qpdf/QPDFObjectHandle.hh
```diff
@@ -1629,15 +1641,9 @@ QPDFObjectHandle::isIndirect() const
 }
 
 inline bool
-QPDFObjectHandle::isInitialized() const noexcept
+QPDFObjectHandle::isInitialized() const
```
jberkenbilt, line 1633: For my edification, why remove `noexcept`?

include/qpdf/ObjectHandle.hh
```diff
@@ -0,0 +1,64 @@
...
+
+namespace qpdf
+{
+    // Basehandle is only used as a base-class for QPDFObjectHandle like classes. Currently the only
+    // methods exposed in public API are operators to convert derived objects to QPDFObjectHandle,
+    // QPDFObjGen and bool.
+    class BaseHandle
+    {
+      public:
+        explicit inline operator bool() const;
```
jberkenbilt, line 51: Given how small and obvious these are, I think it's okay for them to be inline.

include/qpdf/ObjectHandle.hh
```diff
@@ -44,6 +50,8 @@ namespace qpdf
 
         // The rest of the header file is for qpdf internal use only.
```
jberkenbilt, line 55: Is this comment in the wrong place now?

include/qpdf/ObjectHandle.hh
```diff
@@ -44,6 +50,8 @@ namespace qpdf
...
 
+        inline qpdf_object_type_e type_code() const;
```
jberkenbilt, line 62: ... or is this supposed to be protected? I would think this is public so should be above the comment.

libqpdf/qpdf/QPDFObjectHandle_private.hh
```diff
@@ -42,6 +42,84 @@ namespace qpdf
     class BaseDictionary: public BaseHandle
```
jberkenbilt, line 60: The commit comment says `BaseDirectory` instead of `BaseDictionary`

libqpdf/QPDF_Array.cc
```diff
@@ -46,6 +46,96 @@ Array::array() const
     return nullptr; // unreachable
 }
 
+Array::iterator
```
jberkenbilt, line 49: This is a lot better than aItems and dItems (vs. aitems and ditems). It creates a good, idiomatic approach that can coexist with old code. I think we should write something in the docs about switching to the new style. This doesn't have to block release of 12.0, but we should get it in at some point. Maybe I should write it to test my understanding, and you can review it.

libqpdf/QPDF.cc
```diff
@@ -2346,9 +2346,8 @@ QPDF::reserveObjects(QPDFObjectHandle foreign, ObjCopier& obj_copier, bool top)
 
     if (foreign_tc == ::ot_array) {
         QTC::TC("qpdf", "QPDF reserve array");
-        int n = foreign.getArrayNItems();
-        for (int i = 0; i < n; ++i) {
-            reserveObjects(foreign.getArrayItem(i), obj_copier, false);
+        for (auto const& item: foreign.as_array()) {
```
jberkenbilt, line 2349: It is excellent that you have split out this and the corresponding one for dictionaries into separate commits. This will be very helpful when I try to upload into my brain how to make use of the new APIs.

I wonder whether we should have a preprocessor symbol that, when defined, adds a warning around uses of aitems and ditems to help people locate them. Probably not...too complex. Better to have people just search.

jberkenbilt (approved):
This is great work. I have a few minor comments, but nothing substantive. Feel free to fix them, and I don't need to rereview. Every aspect of this is first class, including the way you broke down the commits. It was very easy to review with the great summary.

I think we should write something in the docs telling people how to make use of the new APIs. Other than release notes, this doesn't have to block 12.0, but we should get it out soon. Maybe I should write it and you should review -- it would help me get it into my head and make sure I understand it fully. The essence of this change is something I've always wanted, but I would never have gotten around to doing it, and I don't think I would have done it as well.

I'm "approving" because I know you'll go back and fix based on my comments (moving the comment if appropriate, fixing the typo in the commit comment) before merging.

include/qpdf/QPDFObjectHandle.hh
m-holger: I found that there is a noticeable and surprising performance cost to noexcept .

include/qpdf/ObjectHandle.hh
m-holger: They are part of the private API. The implementation is in a private header and not accessible externally.

libqpdf/QPDF_Array.cc
m-holger: At the moment these classes are qpdf private API. Before exposing it I want to fully implement the Array and Dictionary in case I find I have to tweak the API. At that time we will need to update the documentation as well as the example programs.

The general intention is that method comments in the header files should be written as methods are added / modified. I skipped this given the 12.0 target date but I'll need to catch up once this is out of the way.

libqpdf/QPDF.cc
m-holger: I was considering this too, but slightly more widely  than aitems / ditems.  Something for 12.1 I think.

include/qpdf/ObjectHandle.hh
jberkenbilt: What is the purpose of a public inline function whose implementation is private? I guess it just makes it accessible everywhere within the internals? I guess it's okay, but it seems a little weird. I guess I'm murky on this choice. I don't have a problem with it necessarily as a declaration of an inline function with no implementation shouldn't have any affect on the ABI.

libqpdf/QPDF_Array.cc
jberkenbilt: Okay, we can work these in gradually throughout 12.x as they will be additions rather than ABI-breaking changes. I guess that's what you meant in your summary.

PR #1373, opened 2025-02-27
Added prefix m to page ranges

marxnotdead:
The new prefix m can be used to specify a number as the maximum of pages that should be used.
Before these changes specifying a range larger than the amount of pages in the document would give an error.
When using qpdf to split large numbers of documents with varying page numbers it was not possible to use the same parameters to always get a set number of the first pages.

For example to get at most the first 5 pages one would use 1-m5.

This commit changes the regular expression to parse page ranges to accept m or r as prefixes. In addition the CLI help as well as the online documentation have been expanded. A very basic test was added aswell.

### DIFF
--- a/libqpdf/QUtil.cc
+++ b/libqpdf/QUtil.cc
@@ -1308,7 +1308,7 @@ QUtil::parse_numrange(char const* range, int max)
     // range parsing is used only during argument processing. It is not used during processing of
     // PDF files.
 
-    static std::regex group_re(R"((x)?(z|r?\d+)(?:-(z|r?\d+))?)");
+    static std::regex group_re(R"((x)?(z|[rm]?\d+)(?:-(z|[rm]?\d+))?)");
     auto parse_num = [&max](std::string const& s) -> int {
         if (s == "z") {
             return max;
@@ -1316,6 +1316,9 @@ QUtil::parse_numrange(char const* range, int max)
         int num;
         if (s.at(0) == 'r') {
             num = max + 1 - string_to_int(s.substr(1).c_str());
+        }
+        else if (s.at(0) == 'm') {
+          num = std::min(string_to_int(s.substr(1).c_str()), max);
         } else {
             num = string_to_int(s.c_str());
         }
--- a/libtests/qtest/numrange.test
+++ b/libtests/qtest/numrange.test
@@ -73,6 +73,10 @@ my @nrange_tests = (
     ["4-10,x7-9,12-8,xr5",
      "numeric range 4-10,x7-9,12-8,xr5 -> 4 5 6 10 12 10 9 8",
      0],
+    ,
+    ["4-m10",
+     "numeric range 4-m10 -> 4 5 6 7 8 9 10",
+     0],
     );
 foreach my $d (@nrange_tests)
 {
--- a/libqpdf/qpdf/auto_job_help.hh
+++ b/libqpdf/qpdf/auto_job_help.hh
@@ -301,6 +301,7 @@ previous group. A number may be one of
 
 - <n>        where <n> represents a number is the <n>th page
 - r<n>       is the <n>th page from the end
+- m<n>       is the <n>th page or the last page, if <n> is larger than the number of pages
 - z          the last page, same as r1
 
 - a,b,c      pages a, b, and c
--- a/manual/cli.rst
+++ b/manual/cli.rst
@@ -1298,6 +1298,7 @@ Page Ranges
 
    - <n>        where <n> represents a number is the <n>th page
    - r<n>       is the <n>th page from the end
+   - m<n>       is the <n>th page or the last page, if <n> is larger than the number of pages
    - z          the last page, same as r1
 
    - a,b,c      pages a, b, and c
@@ -1318,6 +1319,9 @@ section describes the syntax of a page range.
 
 - A number preceded by ``r`` counts from the end, so ``r1`` is the
   last page, ``r2`` is the second-to-last page, etc.
+
+- A number preceded by ``m`` indicates the page numbered from 1 or the last page,
+  if the number is greater than the number of pages in the document.
 
 - The letter ``z`` represents the last page and is the same as ``r1``.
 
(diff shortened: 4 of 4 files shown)


m-holger:
@marxnotdead  , thank you for your contribution. However, I am not sure that this is the right way to go for a couple a couple of reasons. 

Firstly, the syntax for specifying page ranges is already complex, and adding more options should be carefully considered as it may add flexibility at the expense of usability. 

Secondly, I think there are more cases to be considered. In your test case you use the range `4-m10`. What happens if the input file contains only 2 pages? The range effectively becomes `4-2`, which still generates errors. We could try to work around this by specifying `m4-m10`, but this would become `2-2`, and page 2 is probably not what you want if you ask for pages 4 to 10. Trying to cater for all the edge cases is going to be complex.

I think it's worth taking a step back and to look at the problem again. Fundamentally, your issue is that qpdf gives an error if you specify a page that is not present in the input file, and you argue convincingly that there are situations where this is not appropriate/helpful. To me the most obvious and least complex solution to this problem would be to provide options to handle missing pages differently, the two alternatives being (1) to produce a warning only or (2) quietly ignore requests for missing pages.

marxnotdead:
Thank you very much for your swift answer. 
I like your approach as I noticed on implementing the prefix how complex the situation with page ranges might become and how the new prefix would create additional corner cases and might spawn new feature requests.
Would it be the right way to implement the new options in a new pull request and close this one?
I would suggest an option like --ignore-missing-pages where a warning will be printed. Since the global option --no-warn allows to suppress that.

m-holger:
> Would it be the right way to implement the new options in a new pull request and close this one?

It is best to keep this PR and force push your amended implementation. This keeps all discussion of this feature in one place.

> I would suggest an option like --ignore-missing-pages where a warning will be printed. Since the global option --no-warn allows to suppress that.

When using page ranges to select at most n pages warnings about missing pages would be entirely expected. It should therefore be possible to suppress  such warnings without disabling other perhaps more surprising warnings. I would therefore suggest something like `--on-missing-page=[abort | warn | ignore]`, with `abort` as default option for backward compatibiliy.

@jberkenbilt , what are your thoughts?

jberkenbilt:
This is a good discussion, but I don't think we should allow page ranges that cover missing pages. I think covering all the edge cases is complex; e.g., 4-2 would mean pages 4 down to 2 if there were 4 pages. I think this is a bit of an edge case, and my recommendation would be to preprocess with qpdf --show-npages first and then use the shell or an external tool to construct an unambiguous page range.

m-holger:
@marxnotdead  , sorry for not coming back earlier. We have been concentrating on getting qpdf 12.0 out.

I had a quick look at the code. Unfortunately the approach I suggested is not as straightforward as I hoped. I also have to agree that it is easy to achieve what you want with an external tool, for example pikepdf which would allow you to script this easily in Python.

In view of this, do you still want to go ahead with this PR. If you do I'll give it a bit more thought.

pantomaxdotco:
**PR Summary:**

Summary: This PR introduces a new page range prefix "m" for numeric ranges.
- Updated regex and parsing logic in QUtil.cc to recognize "m<n>".
- Added corresponding help text in the command-line documentation.
- Updated tests and manual examples to verify and document the new behavior.

libqpdf/QUtil.cc
```diff
@@ -1308,14 +1308,17 @@ QUtil::parse_numrange(char const* range, int max)
...
             return max;
         }
         int num;
         if (s.at(0) == 'r') {
             num = max + 1 - string_to_int(s.substr(1).c_str());
+        }
+        else if (s.at(0) == 'm') {
+          num = std::min(string_to_int(s.substr(1).c_str()), max);
         } else {
             num = string_to_int(s.c_str());
```
pantomaxdotco, line 1323: Before using s.at(0) in the lambda, ensure s is non-empty to prevent potential out-of-range exceptions.

```cpp
auto parse_num = [&max](std::string const& s) -> int {
    if (s.empty()) {
        throw std::runtime_error("empty string in parse_num");
    }
    if (s == "z") {
        return max;
    }
    int num;
    if (s.at(0) == 'r') {
        num = max + 1 - string_to_int(s.substr(1).c_str());
    }
    else if (s.at(0) == 'm') {
        num = std::min(string_to_int(s.substr(1).c_str()), max);
    } else {
        num = string_to_int(s.c_str());
    }
    // max == 0 means we don't know the max and are just testing for valid syntax.
    if ((max > 0) && ((num < 1) || (num > max))) {
        throw std::runtime_error("number " + std::to_string(num) + " out of range");
    }
    return num;
};
```

libtests/qtest/numrange.test
```diff
@@ -73,6 +73,10 @@ my @nrange_tests = (
     ["4-10,x7-9,12-8,xr5",
      "numeric range 4-10,x7-9,12-8,xr5 -> 4 5 6 10 12 10 9 8",
      0],
+    ,
+    ["4-m10",
+     "numeric range 4-m10 -> 4 5 6 7 8 9 10",
+     0],
     );
```
pantomaxdotco, line 80: Consider adding additional tests for edge cases (e.g., "m0", non-numeric input following "m") to strengthen coverage of the new prefix.

pantomaxdotco:
Reviewed up to commit:42a8f1f936a69795be78296f5a4211d01d1876e2


<details>
  <summary>Additional Suggestion</summary>
  <details open> <summary>
libqpdf/qpdf/auto_job_help.hh, line:302-309
</summary>
Confirm that the newly added description for the "m<n>" prefix clearly conveys its behavior in line with its implementation.
</details>
<details open> <summary>
manual/cli.rst, line:1299-1307
</summary>
Verify the documentation example for "m<n>" accurately reflects its intended behavior, ensuring consistency with both code and help text.
</details>
</details>



<sub>

[Reviewed by Panto AI](https://www.getpanto.ai?utm_source=pantobotcomments&utm_medium=github)

</sub>

pantomaxdotco:
Panto AI has reviewed this pull request and provided key insights to improve your code.
Panto AI delivers highly relevant, noise-free code reviews for developers' repositories.

Need more in-depth reviews or have questions? Reach out to us at [getpanto.ai](https://www.getpanto.ai) or via email at hello@getpanto.ai 🚀

PR #1375, opened 2025-03-01
Remove more deprecated API

jberkenbilt:
@m-holger Any issues? I'm deciding to leave QPDFObject.hh around indefinitely with an error. I don't think it hurts anything, and it's entirely possible for someone to be working on a system that has an ancient qpdf installed.

### DIFF
--- a/manual/release-notes.rst
+++ b/manual/release-notes.rst
@@ -15,6 +15,14 @@ more detail.
 12.0.0: not yet released
   - API: breaking changes
 
+    - The header file ``qpdf/QPDFObject.hh`` now generates an error if
+      included. This is to prevent that includes it from accidentally
+      working because an old version is installed somewhere on the
+      system.
+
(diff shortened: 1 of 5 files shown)


manual/release-notes.rst
```diff
@@ -15,6 +15,14 @@ more detail.
 12.0.0: not yet released
   - API: breaking changes
 
+    - The header file ``qpdf/QPDFObject.hh`` now generates an error if
+      included. This is to prevent that includes it from accidentally
+      working because an old version is installed somewhere on the
+      system.
```
mara004, line 21: > This is to prevent that includes it from accidentally working because an old version is installed somewhere on the system.

I think something's wrong with this phrase?

jberkenbilt: > I think something's wrong with this phrase?

Thank you. Always read carefully to make sure you didn't any words out, as they say. (Should be "to prevent code that includes it....". I will fix. Thanks for keeping an eye out and taking the time to report. :-)

PR #1376, opened 2025-03-02
Clean up comments in QPDFObject.hh

jberkenbilt:
I've already iterated on this twice, so @m-holger, I'd appreciate if you read these comments and see if you think they're clear. If so, feel free to merge this. This only changes a comment.

### DIFF
--- a/include/qpdf/QPDFObject.hh
+++ b/include/qpdf/QPDFObject.hh
@@ -23,14 +23,14 @@
 #ifndef QPDFOBJECT_OLD_HH
 #define QPDFOBJECT_OLD_HH
 
-// This file exists so that if anyone still includes <qpdf/QPDFObject.hh>
-// and has been ignoring the warning, they will get an error. This reduces
-// the chances of someone accidentally including the file from an
-// installed, old version. The QPDFObject API, which is not part of the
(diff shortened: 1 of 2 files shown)


include/qpdf/QPDFObject.hh
```diff
@@ -23,14 +23,15 @@
...
 
-// This file exists so that if anyone still includes <qpdf/QPDFObject.hh>
-// and has been ignoring the warning, they will get an error. This reduces
-// the chances of someone accidentally including the file from an
-// installed, old version. The QPDFObject API, which is not part of the
-// public API, is defined in QPDFObject_private.hh. Prior to qpdf 11, this
-// file was exposed for some constants. This error was introduced in qpdf
-// 12.
+// Current code should not include <qpdf/QPDFObject.hh>. This file exists
+// to ensure that code that includes doesn't accidentally work because of
```
m-holger, line 27: that code that includes doesn't ->  that code that includes   it   doesn't ?

jberkenbilt: Yes, wow, I had to read it about five times before I perceived the missing word. Thanks.

jberkenbilt:
Made some more tweaks.

m-holger:
Reads ok to me

PR #1377, opened 2025-03-02
Tidy public header files

m-holger:
- Remove unnecessary QPDF_DLLs
- make deleted constructors public
- move some comments to support tooltips
- modernise some constructors and destructors
- change some member shared pointers to unique pointers

### DIFF
--- a/include/qpdf/Buffer.hh
+++ b/include/qpdf/Buffer.hh
@@ -46,6 +46,9 @@ class Buffer
     QPDF_DLL
     Buffer(std::string& content);
 
+    Buffer(Buffer const&) = delete;
+    Buffer& operator=(Buffer const&) = delete;
+
     QPDF_DLL
     Buffer(Buffer&&) noexcept;
     QPDF_DLL
--- a/include/qpdf/QPDFCryptoProvider.hh
+++ b/include/qpdf/QPDFCryptoProvider.hh
@@ -31,7 +31,6 @@
 // This class is part of qpdf's pluggable crypto provider support. Most users won't need to know or
 // care about this class, but you can use it if you want to supply your own crypto implementation.
 // See also comments in QPDFCryptoImpl.hh.
-
 class QPDFCryptoProvider
 {
   public:
@@ -92,7 +91,6 @@ class QPDFCryptoProvider
 
       public:
         Members() = default;
-        QPDF_DLL
         ~Members() = default;
 
       private:
--- a/libqpdf/Pl_Buffer.cc
+++ b/libqpdf/Pl_Buffer.cc
@@ -5,16 +5,24 @@
 #include <cstring>
 #include <stdexcept>
 
+class Pl_Buffer::Members
+{
+  public:
+    Members() = default;
+    Members(Members const&) = delete;
+
+    bool ready{true};
+    std::string data;
+};
+
 Pl_Buffer::Pl_Buffer(char const* identifier, Pipeline* next) :
     Pipeline(identifier, next),
-    m(new Members())
+    m(std::make_unique<Members>())
 {
 }
 
-Pl_Buffer::~Pl_Buffer() // NOLINT (modernize-use-equals-default)
-{
-    // Must be explicit and not inline -- see QPDF_DLL_CLASS in README-maintainer
-}
+// Must be explicit and not inline -- see QPDF_DLL_CLASS in README-maintainer
+Pl_Buffer::~Pl_Buffer() = default;
 
 void
 Pl_Buffer::write(unsigned char const* buf, size_t len)
--- a/libqpdf/Pl_Concatenate.cc
+++ b/libqpdf/Pl_Concatenate.cc
@@ -10,10 +10,8 @@ Pl_Concatenate::Pl_Concatenate(char const* identifier, Pipeline* next) :
     }
 }
 
-Pl_Concatenate::~Pl_Concatenate() // NOLINT (modernize-use-equals-default)
-{
-    // Must be explicit and not inline -- see QPDF_DLL_CLASS in README-maintainer
-}
+// Must be explicit and not inline -- see QPDF_DLL_CLASS in README-maintainer
+Pl_Concatenate::~Pl_Concatenate() = default;
 
 void
 Pl_Concatenate::write(unsigned char const* data, size_t len)
--- a/README-maintainer.md
+++ b/README-maintainer.md
@@ -886,7 +886,7 @@ still things that could potentially break ABI, such as
 * Changes to the types of public or protected data members without
   changing the size
 
-* Changes to the meanings of parameters with changing the signature
+* Changes to the meanings of parameters without changing the signature
 
 Not breaking ABI/API still requires care.
 
--- a/include/qpdf/ClosedFileInputSource.hh
+++ b/include/qpdf/ClosedFileInputSource.hh
@@ -20,22 +20,25 @@
 #ifndef QPDF_CLOSEDFILEINPUTSOURCE_HH
 #define QPDF_CLOSEDFILEINPUTSOURCE_HH
 
-// This is an input source that reads from files, like FileInputSource, except that it opens and
-// closes the file surrounding every operation. This decreases efficiency, but it allows many more
-// of these to exist at once than the maximum number of open file descriptors. This is used for
-// merging large numbers of files.
-
 #include <qpdf/InputSource.hh>
 
 #include <memory>
 
 class FileInputSource;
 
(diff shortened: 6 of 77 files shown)


include/qpdf/Buffer.hh
```diff
@@ -46,6 +46,9 @@ class Buffer
     QPDF_DLL
     Buffer(std::string& content);
 
+    Buffer(Buffer const&) = delete;
+    Buffer& operator=(Buffer const&) = delete;
+
```
m-holger, line 51: I've added these to make it explicit that they are not supported instead of heaving compile errors about implicitly deleted constructors.

include/qpdf/QPDFCryptoProvider.hh
```diff
@@ -49,7 +48,7 @@ class QPDFCryptoProvider
     // Register the given type (T) as a crypto implementation. T must be derived from QPDFCryptoImpl
     // and must have a constructor that takes no arguments.
     template <typename T>
-    QPDF_DLL static void registerImpl(std::string const& name);
+    static void registerImpl(std::string const& name);
```
m-holger, line 52: This one doesn't make sense to me. Unless I am missing something, this should be either private or the template  should be exposed.

m-holger:
Calling it a day - got  pretty cross-eyed in the end

include/qpdf/Buffer.hh
jberkenbilt: Agreed.

include/qpdf/QPDFCryptoProvider.hh
jberkenbilt: You're right. This doesn't work. And here I've been thinking qpdf supported pluggable crypto. I will investigate.

libqpdf/Pl_Concatenate.cc
```diff
@@ -10,10 +10,8 @@ Pl_Concatenate::Pl_Concatenate(char const* identifier, Pipeline* next) :
     }
 }
 
-Pl_Concatenate::~Pl_Concatenate() // NOLINT (modernize-use-equals-default)
-{
-    // Must be explicit and not inline -- see QPDF_DLL_CLASS in README-maintainer
-}
+// Must be explicit and not inline -- see QPDF_DLL_CLASS in README-maintainer
```
jberkenbilt, line 13: Here too

jberkenbilt (changes requested):
I want to check the "Must be explicit and not inline" thing -- we have to investigate and either fix comments or fix code. Perhaps this has been overcome by events, or the `= delete` satisfies it. I'm about to step away and don't have time to look at it today. As an added check, I'd like to run the abi checker to compare the ABI before and after this commit to make sure we didn't break any ABI unintentionally since I think this particular commit is supposed to be ABI-preserving (removing useless QPDF_DLL calls). I will look at it either later this evening or else will squeeze in some time during the week. Since a forward declaration of a templated function is useless, I need to figure out what to do with QPDFCryptoProvider::registerImpl. Worst case, we temporary disable it and add it in 12.1.

I'd like to sign off on this before we merge it.

jberkenbilt:
> Calling it a day - got pretty cross-eyed in the end

I bet. This must have been really tedious, but it's good housekeeping.

libqpdf/Pl_Buffer.cc
```diff
@@ -5,16 +5,24 @@
...
-    m(new Members())
+    m(std::make_unique<Members>())
 {
 }
 
-Pl_Buffer::~Pl_Buffer() // NOLINT (modernize-use-equals-default)
-{
-    // Must be explicit and not inline -- see QPDF_DLL_CLASS in README-maintainer
-}
+// Must be explicit and not inline -- see QPDF_DLL_CLASS in README-maintainer
```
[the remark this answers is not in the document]
m-holger: I have no reason to believe that it no longer applies. Specifying it as ... - default in the .cc file cause an explicit constructor to be exported in the shared library.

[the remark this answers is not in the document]
m-holger: Perhaps the wording can be made clearer, though it works for me.

>Specifying it as ... - default in the .cc file cause an explicit constructor to be exported in the shared library.

I think I checked that before, but checking that the ABI hasn't changed should pick up if this causes unintended changes.

jberkenbilt:
When I do the ABI comparison, I get this:
```
** running ABI comparison **
INTERFACES REMOVED:
  FileInputSource::FileInputSource()
  PDFVersion::PDFVersion()
  Pl_Flate::Members::~Members()
  QPDF::Members::~Members()
  QPDFAcroFormDocumentHelper::Members::~Members()
  QPDFAcroFormDocumentHelper::~QPDFAcroFormDocumentHelper()
  QPDFAnnotationObjectHelper::~QPDFAnnotationObjectHelper()
  QPDFEFStreamObjectHelper::~QPDFEFStreamObjectHelper()
  QPDFEmbeddedFileDocumentHelper::~QPDFEmbeddedFileDocumentHelper()
  QPDFFileSpecObjectHelper::~QPDFFileSpecObjectHelper()
  QPDFFormFieldObjectHelper::~QPDFFormFieldObjectHelper()
  QPDFLogger::Members::~Members()
  QPDFObjGen::unparse[abi:cxx11](char) const
  QPDFObjectHandle::QPDFArrayItems::iterator::~iterator()
  QPDFObjectHandle::QPDFDictItems::iterator::~iterator()
  QPDFOutlineDocumentHelper::Members::~Members()
  QPDFOutlineDocumentHelper::~QPDFOutlineDocumentHelper()
  QPDFOutlineObjectHelper::Members::~Members()
  QPDFOutlineObjectHelper::QPDFOutlineObjectHelper(QPDFObjectHandle, QPDFOutlineDocumentHelper&, int)
  QPDFOutlineObjectHelper::~QPDFOutlineObjectHelper()
  QPDFPageDocumentHelper::~QPDFPageDocumentHelper()
  QPDFPageLabelDocumentHelper::~QPDFPageLabelDocumentHelper()
  QPDFPageObjectHelper::~QPDFPageObjectHelper()
INTERFACES ADDED
  QPDFMatrix::operator!=(QPDFMatrix const&) const
  QPDFObjectHandle::getGeneration() const
  QPDFObjectHandle::getObjectID() const
  QPDFObjectHandle::isIndirect() const
check_abi: **ERROR**: major version must be bumped
```

jberkenbilt:
> When I do the ABI comparison, I get this:

An analysis of this output shows that none of the things that are required to explicit are affected, so if the above list looks right to you (if these are okay to remove), then I have no problems with this.

libqpdf/Pl_Buffer.cc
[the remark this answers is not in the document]
jberkenbilt: > Perhaps the wording can be made clearer, though it works for me.

Nah, it works for me too. I think it's fine as is. I was in a rush, and forgot that these were in the .cc file, so it's fine.

jberkenbilt (approved):
I'm comfortable for you to merge at will. After my additional checks, I'm satisfied that this is safe.

jberkenbilt:
I also successfully linked this program:
```
#include <qpdf/PDFVersion.hh>
int main() {
    PDFVersion p;
    return 0;
}
```
with your branch, indicating that there's no issue with things where the constructor is inlined as = default(). No one should be manually calling the mangled name from the shared library for C++, and no C APIs were affected, so I think we're good to go.

m-holger:
@jberkenbilt  , I have moved the QUtil inline functions to a new private header file and un-inlined the QUtil functions. My feeling is that we should keep future utilities private unless there is a good reason to expose them.

I've also had a look at ABI comparison output. I think this is as expected and I feel comfortable with it.

m-holger:
I am a bit puzzled why when you ran sizes it flagged that QPDFObjectHelper should be removed. It does have an explicit virtual destructor.

That said, should we make the destructor non-virtual protected. I really don't see how upcasting to QPDFObjectHelper (or QPDFDocumentHelper for that matter) makes any sense.

jberkenbilt:
> I am a bit puzzled why when you ran sizes it flagged that QPDFObjectHelper should be removed. It does have an explicit virtual destructor.
> 
> That said, should we make the destructor non-virtual protected. I really don't see how upcasting to QPDFObjectHelper (or QPDFDocumentHelper for that matter) makes any sense.

I was surprised by it, but I didn't look into it. If you think there's an issue, it might be worth looking into the logic around all that. I created that a long time ago, but I could take a fresh look.

jberkenbilt:
Congratulations on merging this. I feel like this is was basically what started the whole rework, so this is a big milestone and unblocks a lot of good future work.

PR #1378, opened 2025-03-03
Register crypto

jberkenbilt:
@m-holger This fixes registering crypto implementations. Good catch on that. I don't think we need a release note for this. Obviously no one was using it! If this merges before #1377 , I can deal with the merge conflicts, or if this passes CI and you think it's acceptable, you can merge it ahead and then rebase...your call.


jberkenbilt:
@m-holger Any objections if I merge this in front of your cleanup PR? It will cause a trivial merge conflict. Either that, or you can go first, and I'll fix the conflict here. I'll be away for a few hours and will check. If you want to merge this, go ahead.

jberkenbilt:
If #1377 is ready, go ahead and merge it first, and I'll fix this one.

m-holger:
No, go ahead.

PR #1381, opened 2025-03-05
Update release notes and reverse making QPDFObjectHandle final

m-holger:
and delay to qpdf 13 to give users notice of the change.

### DIFF
--- a/include/qpdf/QPDFObjectHandle.hh
+++ b/include/qpdf/QPDFObjectHandle.hh
@@ -63,7 +63,7 @@ class Pl_QPDFTokenizer;
 class QPDFMatrix;
 class QPDFParser;
 
-class QPDFObjectHandle final: public qpdf::BaseHandle
+class QPDFObjectHandle: public qpdf::BaseHandle
 {
     friend class QPDFParser;
 
--- a/manual/release-notes.rst
+++ b/manual/release-notes.rst
@@ -1,3 +1,6 @@
+.. _ticket: https://issues.qpdf.org
+.. _shared null: https://wiki.qpdf.org/PDF-null-objects-vs-qpdf-null-objects
+
 .. _release-notes:
 
 Release Notes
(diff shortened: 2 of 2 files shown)


include/qpdf/QPDFObjectHandle.hh
```diff
@@ -63,7 +63,7 @@ class Pl_QPDFTokenizer;
 class QPDFMatrix;
 class QPDFParser;
 
-class QPDFObjectHandle final: public qpdf::BaseHandle
```
jberkenbilt, line 66: No object, but I'm just curious as to why. I don't have a strong opinion about whether it should or shouldn't be final.

manual/release-notes.rst
```diff
@@ -1,3 +1,6 @@
+.. _ticket: https://issues.qpdf.org
```
jberkenbilt, line 1: huh, didn't know you could do that.

manual/release-notes.rst
```diff
@@ -41,13 +44,42 @@ more detail.
     - ``QIntC.hh`` contained the typo ``substract`` in function names,
       which has been fixed to ``subtract``.
 
+    - The protected ``QPDFObjectHelper::oh`` data member has been replaced with
+      the new accessor method ``QPDFObjectHelper::oh()``.
+
+    - Except for abstract classes and the exceptions listed below, sub-classing of
```
jberkenbilt, line 50: Oh, that's why. Good call.

manual/release-notes.rst
```diff
@@ -41,13 +44,42 @@ more detail.
...
 
       - All ``QPDFTokenizer`` push-mode methods.
 
+  - CLI breaking Changes
+
+    - To support the future introduction of sub-commands, the use of filenames without
+      extension and path element as the first argument is no longer supported, and the
+      result may change in the future. For example, ``qpdf check out.pdf`` currently
+      copies the file ``check`` to ``out.pdf`` but may in future check ``out.pdf``.
+      Use ``qpdf ./check out.pdf`` or ``qpdf -- check out.pdf`` instead.
```
jberkenbilt, line 81: Should we warn for this?

jberkenbilt (approved):
You didn't ask for my review, but I thought I'd review it anyway. This is a good level of detail, including for the internal changes.

jberkenbilt:
> You didn't ask for my review, but I thought I'd review it anyway. This is a good level of detail, including for the internal changes.

Oh, you did ask for my review. :-)

manual/release-notes.rst
m-holger: Probably only for names that are likely to clash with sub commands.

I don't thinks there are many pdf files out there without extension and probably far fewer that are likely to clash. In any case, I don't think its urgent. I can't see much happening on this before 13.

PR #1383, opened 2025-03-06
Deprecate QPDFObjectHandle::parse overload and undeprecate isInitialized

m-holger:


m-holger:
I'm also planning to change the wiki from 

> To implement shared nulls, qpdf 12 will use default constructed or uninitialized object handles. The isInitialized method will be removed, and for other methods such as isNull, unparse, shallowCopy, etc, uninitialized object handles will be treated as null objects.

to

> To implement shared nulls, qpdf 13 will use default constructed or uninitialized object handles. For backward compatibility, isInitialized() and isNull() will continue to behave as they do now. A new method null() will return true if the object is an actual null object or is uninitialized. For all other methods such as unparse, shallowCopy, etc, uninitialized object handles will be treated as null objects.

jberkenbilt (approved):

PR #1384, opened 2025-03-07
Enhance error checking of object stream object ids and offsets

m-holger:
The original test file contains multiple entries with id 0 and offset 0. One entry has been modified such that the id is valid (6).

Object streams with invalid offsets are a source of unreproduceable oss-fuzz time-outs.


m-holger:
@jberkenbilt  , not sure whether to call this a bug fix or enhancement

jberkenbilt (approved):
I tested, and other readers reject files with out-of-order objects even when their offsets are valid, so I would agree with calling this a bug fix.

m-holger:
I didn't know that, but I guess it is not that surprising. The only legal way to find the end of an object is using the offset of the next object.

I guess that makes fixing  #1266 a good candidate for 12.0.1 . I was working on that when I was knocked of course in November.

PR #1387, opened 2025-03-08
Update library notes

m-holger:


jberkenbilt (commented):
will merge with mine and the other and push with the preparing release commit

PR #1388, opened 2025-03-08
Pre 12 fixes

jberkenbilt:
Here's a small group of changes I made while preparing the qpdf release. It was enough stuff that I wouldn't mind a second set of eyes, though don't worry if you don't follow the ABI checks. This is not super-critical as only we run it and it doesn't actually impact functionality.

I'm at the end of my energy and will work on pushing out the release tomorrow.


manual/release-notes.rst
```diff
@@ -25,21 +27,22 @@ more detail.
       include ``<qpdf/Constants.h>``, and replace ``QPDFObject::ot_``
       with ``::ot_`` in your code.
 
-    - ``QPDFObjectHandle::replaceOrRemoveKey`` has been removed since
-      it was identical to ``QPDFObjectHandle::replaceKey``.
+    - The deprecated ``QPDFObjectHandle::replaceOrRemoveKey`` function has been
```
m-holger, line 30: function -> method

manual/release-notes.rst
```diff
@@ -25,21 +27,22 @@ more detail.
...
-      has been removed.
+    - The deprecated versionless overload of ``QPDFObjectHandle::getJSON`` has
+      been removed.
 
-    - ``Buffer`` copy constructor and assignment operator have been
-      removed. ``Buffer`` copy operations are expensive as they always
-      involve copying the buffer content. Use ``buffer2 = buffer1.copy();``
-      or ``Buffer buffer2{buffer1.copy()};`` to make it explicit that
-      copying is intended.
+    - The deprecated ``Buffer`` copy constructor and assignment operators have
```
m-holger, line 41: I think operator should be singular

m-holger:
I had a quick read. I'll read it properly in the morning.

jberkenbilt:
Fixed. It's not that critical...I'll probably work on the release after I get back from the grocery store, late morning Eastern US. We switch to daylight saving time this weekend. I have a factoid in my brain that the UK switches a week earlier or later than we do.

m-holger:
I have added one more PR #1389 for 12.0.

Just merge it if you are happy to include it.

PR #1389, opened 2025-03-09
Remove InputSource::Members

m-holger:
There simply is no chance Members will be required during the life of qpdf 12.


jberkenbilt (approved):
will merge with mine and the other and push with the preparing release commit

PR #1392, opened 2025-03-10
Refine recovery from missing startxref (fixes #1335)

### LINKED ISSUE #1335: qpdf fails to repair this pdf
https://dl.acm.org/doi/pdf/10.1109/SC41406.2024.00008

PDF is behind a paywall but there's a workaround for that:

Start private browsing session, go to google, input exactly the address https://dl.acm.org/doi/pdf/10.1109/SC41406.2024.00008  and click on the first link. There is a red PDF button that gives you the PDF file.

```
$ sha256sum SC41406.2024.00008.pdf 
6690cf80b47f385a16b34e41a4d6566a86681eb1b920ba37d3a9a2fd90c82565  SC41406.2024.00008.pdf
```

Firefox, Chrome and Adobe Reader are able to recover from it, but qpdf is not able to repair it.

```
$ qpdf SC4.pdf SC4_repaired.pdf
WARNING: SC4.pdf: file is damaged
WARNING: SC4.pdf (offset 216): xref not found
WARNING: SC4.pdf: Attempting to reconstruct cross-reference table
WARNING: SC4.pdf: reported number of objects (99) is not one plus the highest object number (250)
qpdf: SC4.pdf: unable to find /Root dictionary
```
qpdf did not even write the output file SC4_repaired.pdf

m-holger:
If startxref cannot be found in the last 1024 bytes try finding it in the
whole file and check whether it is valid.


jberkenbilt (approved):
I'm okay with this. I have somewhat lost track of the recovery logic. Originally it followed an implementation note in an ancient version of the PDF spec, probably 1.3, which was current when I started qpdf, but it has evolved (mostly from your changes) to be better able to recover. I have had in my mind actually trying to parse objects as we go for al ong time.

PR #1391, opened 2025-03-10
Refactor xref table reconstruction (Fixes #1362)

m-holger:
Split reconstruction into three passes - scanning of input for objects and
trailer, insertion of objects into the xref table, and loading the trailer.

This allows insertion to take place in the usual reverse order and removes
the need for a separate insertReconstructedXrefEntry method.

It also allows trailer to be tried from most recent to oldest.

Ignore any found trailers without /Root entry.

PR #1396, opened 2025-03-17
Fix object stream error/warning messages reporting wrong object id

m-holger:
This was due to the use of last_object_description, which is not set for the object stream itself.

Also, modify the messages introduced #1391 and #1392 to report the supposed offset of the objects.

### DIFF
--- a/qpdf/qtest/qpdf/issue-143.out
+++ b/qpdf/qtest/qpdf/issue-143.out
@@ -14,9 +14,9 @@ WARNING: issue-143.pdf (object 1 0, offset 24): expected dictionary key but foun
 WARNING: issue-143.pdf (object 1 0, offset 21): stream dictionary lacks /Length key
 WARNING: issue-143.pdf (object 1 0, offset 84): attempting to recover stream length
 WARNING: issue-143.pdf (object 1 0, offset 84): recovered stream length: 606
-WARNING: issue-143.pdf object stream 1 (object 0 0, offset 0): object id is invalid
-WARNING: issue-143.pdf object stream 1 (object 0 0, offset 0): object id is invalid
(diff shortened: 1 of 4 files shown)


qpdf/qtest/qpdf/issue-143.out
```diff
@@ -14,9 +14,9 @@ WARNING: issue-143.pdf (object 1 0, offset 24): expected dictionary key but foun
 WARNING: issue-143.pdf (object 1 0, offset 21): stream dictionary lacks /Length key
 WARNING: issue-143.pdf (object 1 0, offset 84): attempting to recover stream length
 WARNING: issue-143.pdf (object 1 0, offset 84): recovered stream length: 606
-WARNING: issue-143.pdf object stream 1 (object 0 0, offset 0): object id is invalid
-WARNING: issue-143.pdf object stream 1 (object 0 0, offset 0): object id is invalid
-WARNING: issue-143.pdf object stream 1 (object 6 0, offset 0): offset is invalid (must be larger than previous offset 0)
-WARNING: issue-143.pdf object stream 1 (object 0 0, offset 0): object id is invalid
+WARNING: issue-143.pdf object stream 1 (object 0 0, offset 4): object id is invalid
+WARNING: issue-143.pdf object stream 1 (object 0 0, offset 15): object id is invalid
```
m-holger, line 18: @jberkenbilt  , what offset should we be using here? The position where the invalid entry is at the beginning of the object stream, or the offset where the invalid entry claims the object is located?

jberkenbilt: @m-holger  I'd probably go with the position of the invalid entry at the beginning of the object stream.

PR #1397, opened 2025-03-18
Refactor QPDF::resolveObjectsInStream

m-holger:
Main changes:

- object description strings are now only created when needed
- a vector is used instead of a map to process the object id / offset info


m-holger:
For info, results of `qpdf --report-memory-usage --object-streams=generate pdf-spec-1-100.pdf  a.pdf` :

|  | | |
| --- | --- | --- |
| PPR1397 |  172,521,308 | 0.695 |
| 12.0.0 | 199,769,286 | 0.805 |
| 11.10.1 | 248,173,905  | 1.000 |

PR #1400, opened 2025-03-18
Fix parsing of object streams containing objects with no white-space between them

m-holger:
To enforce the rule that objects end at the start-offset of the next
object, each object is parsed in it own object stream.

To facilitate this, a new private API input source is::OffsetBuffer has
been added which only contains the object but reports offsets relative to
the start of the object stream. This is adapted from OffsetInputSource by
changing the direction of the offset, endowing it with its own
BufferInputSource and striooing out checks duplicated in BufferInputSource.

Fixes the expected failure in the test case added in #1266.

### DIFF
--- a/manual/release-notes.rst
+++ b/manual/release-notes.rst
@@ -21,6 +21,11 @@ more detail.
       integer object. Previously the method returned false if the first
       dictionary object was not a linearization parameter dictionary.
 
+    = Fix parsing of object streams containing objects not seperated by
+      white-space. Pre-2020 editions of the PDF specification incorrectly
+      stated that white-space was required between objects. qpdf relied on this
+      when parsing object streams.
+
     - Fix two object stream error/warning messages that reported the wrong
       object id.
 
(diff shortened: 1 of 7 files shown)


jberkenbilt:
> BufferInputSource and striooing out checks duplicated in BufferInputSource.

My typo correction decoder can't figure out what word `striooing` was supposed to be, but it's a nice-looking word. :-) Anyway, I get the idea.

manual/release-notes.rst
```diff
@@ -21,6 +21,11 @@ more detail.
       integer object. Previously the method returned false if the first
       dictionary object was not a linearization parameter dictionary.
 
+    = Fix parsing of object streams containing objects not seperated by
```
jberkenbilt, line 24: `-`, not `=`

m-holger: Already fixed in a later commit

m-holger:
> My typo correction decoder can't figure out what word striooing was supposed to be, but it's a nice-looking word. :-) Anyway, I get the idea.

stripping - adjacent keys

PR #1404, opened 2025-03-22
Refactor writing of object streams

m-holger:


m-holger:
@jberkenbilt  , can you have a look at this. It just feels too easy.

PR #1409, opened 2025-03-26
Add new CLI Option --remove-structure

m-holger:
Removing the structure tree can have a noticeable impact on the size of the output file, eg
```
./qpdf PDF32000_2008.pdf --pages . 81-83 -- table28.pdf
```
table28.pdf size: 14,690,548 bytes

```
./qpdf PDF32000_2008.pdf --remove-structure --pages . 81-83 -- table28.pdf
```
table28.pdf size: 537,395 bytes


jberkenbilt (approved):
I'm fine with this change. For people who are only interested in print/rendering, it can be a big savings as you pointed out.

PR #1418, opened 2025-04-05
Deprecate QPDF::optimize

m-holger:
optimize populates private tables that are not accessible through the public API. There seems no point exposing the method publicly.


jberkenbilt (approved):
agreed

PR #1420, opened 2025-04-05
Detect text using a heuristic for clear-text metadata

jberkenbilt:
I'm not quite done yet because I still need to add tests with actual PDF files with various combinations of encrypted/clear-text metadata, but this is ready for your initial comments. There is a performance cost to this, but hopefully not very large because it will only impact metadata streams. Unfortunately, I had to change std::unique_ptr<Pipeline> to std::vector<std::shared_ptr<Pipeline>> for heap. I could pass two arguments if you think it is important enough. This is where I have tended toward simplicity over performance, but I don't want to undo your hard work.

### DIFF
--- a/include/qpdf/QPDF.hh
+++ b/include/qpdf/QPDF.hh
@@ -835,6 +835,7 @@ class QPDF
         qpdf_offset_t offset,
         size_t length,
         QPDFObjectHandle dict,
+        bool is_root_metadata,
         Pipeline* pipeline,
         bool suppress_warnings,
         bool will_retry);
@@ -848,6 +849,7 @@ class QPDF
         qpdf_offset_t offset,
         size_t length,
         QPDFObjectHandle dict,
+        bool is_root_metadata,
         Pipeline* pipeline,
         bool suppress_warnings,
         bool will_retry);
@@ -920,6 +922,7 @@ class QPDF
         Pipeline*& pipeline,
         QPDFObjGen og,
         QPDFObjectHandle& stream_dict,
+        bool is_root_metadata,
         std::unique_ptr<Pipeline>& heap);
 
     // Methods to support object copying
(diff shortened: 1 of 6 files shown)


jberkenbilt:
The logic of apply the filter is wrong as currently implemented, but the text detection piece is probably okay.

jberkenbilt:
I don't think this is going to work. I may back it out and try a different approach closer to the original submission but with a fix for foreign streams.

jberkenbilt:
I will be abandoning the text detector approach and replacing it with something else.

libqpdf/qpdf/Pl_TextIfElse.hh
```diff
@@ -0,0 +1,30 @@
...
+
+// This pipeline has no `next` associated with the base class. Instead, it picks which of two
+// candidate `next`s to call based on its detection of text with a bias toward text consisting
+// of characters <= 0x7FF. I had considered writing a general IfElse pipeline that takes a closure,
+// but I decided that's premature generalization.
+class Pl_TextIfElse: public Pipeline
+{
+  public:
+    Pl_TextIfElse(char const* identifier, Pipeline* next_if_text, Pipeline* next_if_binary);
+    ~Pl_TextIfElse() override;
```
m-holger, line 14: I would be less general and write this in the knowledge that next_is_binary is the decrypt pipeline, and change the parameter to a unique_ptr<Pipeline>.  This allows us to move the decrypt pipeline into Pl_TextIfElse and setting the decrypt pipeline to  Pl_TextIfElse

m-holger:
I'll continue after dinner

jberkenbilt:
@m-holger I'll be reverting the Pl_TextIfElse pipeline but I'm making changes to the inner Stream class which is new to me, so your review will be appreciated on that. I'll tag you when ready. I know it's late for you, so no worries if it's not today.

jberkenbilt:
will create a new PR with correct fix when ready for review...hopefully not more than a few minutes. working on tests.

jberkenbilt:
See #1421 for a (hopefully) correct implementation

PR #1422, opened 2025-04-06
Jpeg quality

jberkenbilt:
@cdosborn This is based on your fix but different. If interested, look it over to see how I did it. This doesn't break any API compatibility and uses the existing facility for performing additional configuration. Thanks for the code and nudge on this old issue.


m-holger:
2nd commit message:

>> Things << to github user cdosborn for the basic enhancement.

Also, would @cdosborn work?

include/qpdf/Pl_DCT.hh
```diff
@@ -62,6 +63,10 @@ class QPDF_DLL_CLASS Pl_DCT: public Pipeline
         virtual void apply(jpeg_compress_struct*) = 0;
     };
 
+    QPDF_DLL
+    static std::shared_ptr<CompressConfig>
+        make_compress_config(std::function<void(jpeg_compress_struct*)>);
```
m-holger, line 68: I think a unique_ptr would work here. If necessary the caller can let a shared_ptr take ownership.

manual/cli.rst
```diff
@@ -1028,6 +1028,22 @@ Related Options
...
+      When rewriting images with --optimize-images, set a
+      quality level from 0 (lowest) to 100 (highest) for writing new
+      images. Higher quality results in larger images, and lower quality
+      results in smaller images. This option is only useful with
+      --optimize-images.
+
+   When rewriting images with :qpdf:ref:`--optimize-images`, set a
+   quality level from 0 (lowest) to 100 (highest) for writing new
+   images. Higher quality results in larger images, and lower quality
+   results in smaller images. This option is only useful with
```
m-holger, line 1044: ditto

manual/cli.rst
```diff
@@ -4005,6 +4021,11 @@ generate:
   lossy, so images may have artifacts. These are not usually
   noticeable to the casual observer.
 
+- ``--jpeg-quality=...``: use a lower value for JPEG quality level.
+  The default of most JPEG libraries is 75. Smaller numbers result in
```
m-holger, line 4025: Should we mention that this requires --optimize-images. As it stands, on a quick read it would be reasonable to assume that one could compress jpeg images without replacing non-jpeg images.

manual/cli.rst
```diff
@@ -1028,6 +1028,22 @@ Related Options
...
 
+.. qpdf:option:: --jpeg-quality=level
+
+   .. help: set jpeg quality level for jpeg
+
+      When rewriting images with --optimize-images, set a
+      quality level from 0 (lowest) to 100 (highest) for writing new
+      images. Higher quality results in larger images, and lower quality
+      results in smaller images. This option is only useful with
+      --optimize-images.
```
m-holger, line 1039: useful or effective?

manual/cli.rst
```diff
@@ -1028,6 +1028,22 @@ Related Options
...
+
+   When rewriting images with :qpdf:ref:`--optimize-images`, set a
+   quality level from 0 (lowest) to 100 (highest) for writing new
+   images. Higher quality results in larger images, and lower quality
+   results in smaller images. This option is only useful with
+   :qpdf:ref:`--optimize-images`. See also :ref:`small-files`.
```
m-holger, line 1045: ditto

manual/release-notes.rst
```diff
@@ -38,11 +38,20 @@ more detail.
       unencrypted. qpdf has always incorrectly handled all
       ``/Metadata`` streams as special with cleartext metadata.
 
+  - Library Enhancements
+
+    - Add function ``Pl_DCT::make_compress_config`` to return a
+      ``Pl_DCT::CompressConfig`` shared pointer from a
+      ``std::function`` for a more modern configuration option.
```
m-holger, line 45: unique pointer?

m-holger:
Looking at QPDFJob, as far as I can tell the option only applies to images converted to JPEG, not existing JPEGs. I don't think this is clear enough in the option help.

jberkenbilt:
> Also, would `@cdosborn` work?

Fixed typo. But work for what? It's a comment in the git history. Maybe it would work to make it linkable when viewed on github, but I don't think that's useful, and people will see the comment (if anyone ever looks) in other contexts.

manual/cli.rst
jberkenbilt: Actually, making it recompress existing jpeg images is a very small fix, and it seems useful since it will still only keep the new result if it shrinks the file size.

m-holger:
Not saying you should drop the "github user". But I just checked `github user @cdosborn` brings up the link and I don't think it is going to confuse anybody who doesn't read it on github.

manual/cli.rst
m-holger: I looked at some of the history in the meantime. I think this would make a few users happy.

jberkenbilt:
> Looking at QPDFJob, as far as I can tell the option only applies to images converted to JPEG, not existing JPEGs. I don't think this is clear enough in the option help.

I fixed this. I think it should apply to existing JPEG images.

jberkenbilt:
@m-holger Great comments. You can take a fresh look or just compare the latest force push at your discretion.

m-holger:
Looks good to me.

Longer term I think it would make sense to allow the option at input file and pages level. I'll update https://github.com/qpdf/qpdf/wiki/Pages-epic-%3A-next-steps accordingly.

PR #1425, opened 2025-04-09
Fix macOS build failure (fixes #1424)

### LINKED ISSUE #1424: Minor issue in building qpdf on macOS
I have been building and deploying qpdf on macOS for some time. While this is not officially supported, the building instructions for UNIX/Linux usually worked fine on macOS. Today I'm trying to build the latest release 12.1.0 and hit this issue on macOS:

```
qpdf-12.1.0/libqpdf/QPDFJob.cc:126:54: error: no matching function for call to 'jpeg_set_quality'
  126 |             [quality](jpeg_compress_struct* cinfo) { jpeg_set_quality(cinfo, quality, false); });
      |                                                      ^~~~~~~~~~~~~~~~
/opt/jpeg/include/jpeglib.h:992:14: note: candidate function not viable: no known conversion from 'bool' to 'boolean' for 3rd argument
  992 | EXTERN(void) jpeg_set_quality JPP((j_compress_ptr cinfo, int quality,
      |              ^
  993 |                                    boolean force_baseline));
      |                                    ~~~~~~~~~~~~~~~~~~~~~~
```
Did some dig and found out macOS system header defines `boolean` to be `unsigned char`, so the error above is because 3rd argument to `jpeg_set_quality` (`bool`) cannot be implicitly converted to `boolean` (`unsigned char`). An explicit cast resolves this issue:

`jpeg_set_quality(cinfo, quality, static_cast<boolean>(false))`

I'm wondering whether this issue could be fixed or not, even if macOS is not officially supported?

m-holger:
@jberkenbilt ,  do you want to do a bug fix release for this?


jberkenbilt:
Where is `boolean` defined? Is it defined by all jpeg libraries? I'm not comfortable with this fix without more research or investigation.

jberkenbilt:
But, yes, I would do a bugfix release. But homebrew already has 12.1.0, and it passes in CI as you pointed out, so this clearly doesn't affect everyone. If this ends up being the right fix, it needs a comment explaining where boolean comes from and why it's okay, or we need some more reliable test.

m-holger:
turbo-jpeg   header files contains

```C

/*
 * On a few systems, type boolean and/or its values FALSE, TRUE may appear
 * in standard header files.  Or you may have conflicts with application-
 * specific header files that you want to include together with these files.
 * Defining HAVE_BOOLEAN before including jpeglib.h should make it work.
 */

#ifndef HAVE_BOOLEAN
typedef int boolean;
#endif
#ifndef FALSE                   /* in case these macros already exist */
#define FALSE   0               /* values of boolean */
#endif
#ifndef TRUE
#define TRUE    1
#endif
```

jpeg have:

```C
#ifndef HAVE_BOOLEAN
#if defined FALSE || defined TRUE || defined QGLOBAL_H
/* Qt3 defines FALSE and TRUE as "const" variables in qglobal.h */
typedef int boolean;
#ifndef FALSE			/* in case these macros already exist */
#define FALSE	0		/* values of boolean */
#endif
#ifndef TRUE
#define TRUE	1
#endif
#else
typedef enum { FALSE = 0, TRUE = 1 } boolean;
#endif
#endif
```

PR #1428, opened 2025-04-11
Don't apply filters to empty stream objects

### LINKED ISSUE #1426: Don't apply filters to zero-length stream objects
I've noticed my Brother printer is rejecting PDF files if you have an empty stream object with a filter applied to it. The PDF 1.7 specification is not clear about whether this is allowed or not, but this is technically permissible as far as zlib is concerned. Since applying the filter to an empty stream object doesn't accomplish anything other than increasing the output file size, can we stop adding filters to zero-length stream objects regardless of whether compression is enabled or not? This would improve compatibility with devices that ship with buggy PDF parsers that the user have no control over. I'm happy to send a pull request to fix this if there are no objections to this proposal.

Just to be clear, this is completely unrelated to CUPS and any filters. I'm submitting the PDF file directly over IPP to a printer that reports PDF 1.7 support (`pdf-versions-supported (keyword) = iso-32000-1_2008`). Non-empty stream objects can have filters applied without any problems.

Here are some test files that I used to reproduce the issue. The original file is `about:blank` saved as PDF in Chromium, which the printer will accept. Running it through qpdf without disabling compression produces a file that the printer rejects, with stream compression disabled it's printable once again.

[original.pdf](https://github.com/user-attachments/files/19683606/original.pdf)

[nocompress.pdf](https://github.com/user-attachments/files/19683607/nocompress.pdf)

[compressed.pdf](https://github.com/user-attachments/files/19683608/compressed.pdf)

`qpdf original.pdf compressed.pdf`, rejected by the printer
```
5 0 obj
<< /Length 0 /Filter /FlateDecode >>
stream
endstream
endobj
```

`qpdf --compress-streams=n original.pdf nocompress.pdf`, accepted by the printer
```
5 0 obj
<< /Length 0 >>
stream
endstream
endobj
```

TinyServal:
Decoding an empty stream is never going to yield a meaningful output, drop the filter for empty stream objects to improve compatibility with some Brother printers. Fixes #1426.

### DIFF
--- a/libqpdf/QPDFWriter.cc
+++ b/libqpdf/QPDFWriter.cc
@@ -1293,6 +1293,12 @@ QPDFWriter::willFilterStream(
         QTC::TC("qpdf", "QPDFWriter compressing uncompressed stream");
     }
 
+    // Disable compression for empty streams to improve compatibility
+    if (stream_dict.getKey("/Length").isInteger() && stream_dict.getKey("/Length").getIntValue() == 0) {
+        filter = true;
+        compress_stream = false;
+    }
+
     bool filtered = false;
     for (bool first_attempt: {true, false}) {
         PipelinePopper pp_stream_data(this);
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,7 @@ Doxyfile
 
 # ./Testing is created if you run ctest from the wrong place.
 /Testing
+
+# macOS junk files
+.DS_Store
+._*
--- a/qpdf/qtest/object-stream.test
+++ b/qpdf/qtest/object-stream.test
@@ -16,7 +16,7 @@ cleanup();
 
 my $td = new TestDriver('object-stream');
 
-my $n_tests = 10 + (36 * 4) + (12 * 2);
+my $n_tests = 10 + (36 * 4) + (12 * 2) + 4;
 my $n_compare_pdfs = 36;
 
 for (my $n = 16; $n <= 19; ++$n)
@@ -125,6 +125,24 @@ $td->runtest("adjacent compressed objects",
              {$td->FILE => "no-space-compressed-object.out",
               $td->EXIT_STATUS => 0},
              $td->NORMALIZE_NEWLINES);
+
+# Never compress empty streams
+$td->runtest("never compress empty streams",
+             {$td->COMMAND => "qpdf --compress-streams=y --static-id" .
+                  " empty-stream-uncompressed.pdf a.pdf"},
+             {$td->STRING => "", $td->EXIT_STATUS => 0});
+$td->runtest("check file",
+             {$td->FILE => "a.pdf"},
+             {$td->FILE => "empty-stream-uncompressed.pdf"});
+
+# Always remove filters from compressed empty streams
+$td->runtest("always remove filters from empty streams",
+             {$td->COMMAND => "qpdf --compress-streams=y --static-id" .
+                  " empty-stream-compressed.pdf a.pdf"},
+             {$td->STRING => "", $td->EXIT_STATUS => 0});
+$td->runtest("check file",
+             {$td->FILE => "a.pdf"},
+             {$td->FILE => "empty-stream-uncompressed.pdf"});
 
 cleanup();
 $td->report(calc_ntests($n_tests, $n_compare_pdfs));
--- a/qpdf/qtest/qpdf/empty-stream-compressed.pdf
+++ b/qpdf/qtest/qpdf/empty-stream-compressed.pdf
@@ -1,0 +1,31 @@
+%PDF-1.4
+%¿÷¢þ
+1 0 obj
+<< /Pages 3 0 R /Type /Catalog /ViewerPreferences << /DisplayDocTitle true /Type /ViewerPreferences >> >>
+endobj
+2 0 obj
+<< /CreationDate (D:20250409064524+00'00') /Creator (Mozilla/5.0 \(X11; Linux x86_64\) AppleWebKit/537.36 \(KHTML, like Gecko\) Chrome/135.0.0.0 Safari/537.36) /ModDate (D:20250409064524+00'00') /Producer (Skia/PDF m135) /Title (about:blank) >>
+endobj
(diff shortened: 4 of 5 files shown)


jberkenbilt (commented):
I don't think this is how I would have gone about it, but I haven't been in the code for a while. I think I would have probably mutated a copy of the filters in one place if the stream was empty.  I don't have time to look at it right now though. Anyway, I don't see any reason to keep any of the filters for an empty stream. While qpdf has its own lzw implementation, I don't know if it's tested for empty streams, and the same would be true of dct, etc.

Ignoring macos junk files is good. I do all my dev on Linux, so I guess I haven't run into that. :-)

TinyServal:
I agree, it's probably cleaner to get rid of all filters in one place if the stream is empty, instead of manually patching out all the places where `/FlateDecode` is added to the object. Will go with that approach instead then.

libqpdf/QPDFWriter.cc
```diff
@@ -1274,8 +1274,14 @@ QPDFWriter::willFilterStream(
         QPDFObjectHandle filter_obj = stream_dict.getKey("/Filter");
         if (!m->recompress_flate && !stream.isDataModified() && filter_obj.isName() &&
             (filter_obj.getName() == "/FlateDecode" || filter_obj.getName() == "/Fl")) {
-            QTC::TC("qpdf", "QPDFWriter not recompressing /FlateDecode");
-            filter = false;
+            bool empty_stream = stream_dict.getKey("/Length").getIntValue() == 0;
+            if (empty_stream) {
+                QTC::TC("qpdf", "QPDFWriter is removing /FlateDecode from an empty stream");
+                filter = true;
```
m-holger, line 1280: If you add a new QTC::TC entry, you also have to update the qpdf/qpdf.testcov file and ensure that there is a test case to exercise the line during CI. See 
[coverage](https://qpdf.readthedocs.io/en/stable/contributing.html#coverage) in the manual.

libqpdf/QPDFWriter.cc
```diff
@@ -1750,7 +1759,7 @@ QPDFWriter::writeObjectStream(QPDFObjectHandle object)
     adjustAESStreamLength(length);
     writeString(" /Length " + std::to_string(length));
     writeStringQDF("\n ");
-    if (compressed) {
+    if (compressed && length > 0) {
```
m-holger, line 1762: Again, just set compressed to false if the object stream is empty (i.e. if m->object_stream_to_objects[old_id] is empty).

libqpdf/QPDFWriter.cc
```diff
@@ -1274,8 +1274,14 @@ QPDFWriter::willFilterStream(
         QPDFObjectHandle filter_obj = stream_dict.getKey("/Filter");
         if (!m->recompress_flate && !stream.isDataModified() && filter_obj.isName() &&
             (filter_obj.getName() == "/FlateDecode" || filter_obj.getName() == "/Fl")) {
-            QTC::TC("qpdf", "QPDFWriter not recompressing /FlateDecode");
-            filter = false;
+            bool empty_stream = stream_dict.getKey("/Length").getIntValue() == 0;
```
m-holger, line 1277: This if statement deals with the the specific case of a single flate filter. You want to set filter for empty streams outside the if statements.
 
Also, you should set compress_stream to false to stop the empty output stream from getting recompressed rather than  trying to suppress the writing of `/Filter /FlateDecode` for empty streams.

TinyServal:
Made some changes and now it should be better. Now I'm just setting `compress_stream` to false for empty object streams so they won't get compressed, which seems to work well with PDF files that already has compressed zero-length stream objects.

libqpdf/QPDFWriter.cc
TinyServal: Disabling compression in `QPDFWriter::willFilterStream()` is good enough in this case, so I just dropped this change.

libqpdf/QPDFWriter.cc
```diff
@@ -1293,6 +1293,12 @@ QPDFWriter::willFilterStream(
         QTC::TC("qpdf", "QPDFWriter compressing uncompressed stream");
     }
 
+    bool empty_stream = stream_dict.getKey("/Length").getIntValue() == 0;
+    if (empty_stream) {
```
m-holger, line 1297: It is preferable not to create the temporary variable `empty_stream` but to put the condition directly into the if statement `if (stream_dict.getKey("/Length").getIntValue() == 0) {...`. It makes no difference to the compiler, but to the human reader it raises the question whether something else in the method depends on it.

TinyServal: Makes sense, fixed.

TinyServal:
I'm having trouble running the test suite locally, `ctest` says there are no tests even when `ENABLE_QTC` is set.

TinyServal:
Finally figured out how to run tests locally and they are all passing now.

TinyServal:
Forgot to specify the directory when running the tests so a lot of them are still failing. Have to look into that still.

TinyServal:
Should be all fixed now, wasted quite a bit of time before I realized I was running stale code from the wrong build directory. 🤦

m-holger (commented):
@TinyServal , looks good to me. Ignore the CI failures.

m-holger:
@jberkenbilt , should we adjust Stream::filterable such that all empty streams are `filterable`, `!specialized_compression`and `!lossy_compression` as well as intermediately calling finish on the output pipeline in   Stream::pipeStreamData ?

jberkenbilt:
> @jberkenbilt , should we adjust Stream::filterable such that all empty streams are `filterable`, `!specialized_compression`and `!lossy_compression` as well as intermediately calling finish on the output pipeline in Stream::pipeStreamData ?

Sounds reasonable. Let's try to get this over the finish line for 12.3.

m-holger:
The adjustment to treat empty streams as filterable is implemented in #1458 .

@TinyServal  , your commits are included in and will merge with #1458.

m-holger:
Included in #1458

PR #1430, opened 2025-04-11
Refactor QPDFFormFieldObjectHelper ValueSetter::writeAppearance

m-holger:
Adjusting for under/overflow by repeated incrementing/decrementing can cause excessive runtime with invalid BBox coordinates.

Fixes oss-fuzz issue 409905355.


m-holger:
@jberkenbilt , I'll merge when CI passes, but can you check my fix when you get time.

m-holger:
As far as I can tell oss-fuzz didn't build any projects today. and therefore is still fuzzing qpdf excluding this PR.

jberkenbilt:
I can't think of any reason why this was a `while` instead of an `if`. Perhaps it's a remnant of earlier logic. It seems like all it's doing is handling lists that are too long to handle all the items, and it seems like all you have to do is move the start and/or end once, so I think your fix looks fine.

PR #1431, opened 2025-04-15
Introduce MbedTLS crypto provider

### LINKED ISSUE #1427: Add MbedTLS crypto provider
[MbedTLS](https://www.trustedfirmware.org/projects/mbed-tls/) is a very mature crypto library suitable for use cases with strict resource constraints, such as embedded linux and microcontrollers. The resulting binary size will be significantly smaller compared to OpenSSL or GnuTLS because you can cherry-pick the exact algorithms you need and leave everything else out. This would be pretty useful for containers or niche use cases like using libqpdf as a javascript library (compiled to wasm). Additionally, having MbedTLS support will help with deprecating the native crypto provider too.

Context: I'm currently trying to use libqpdf as a JavaScript library, both OpenSSL and GnuTLS are difficult to compile to wasm without patches, and the resulting binary sizes are unacceptable for the web. [SubtleCrypto](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto) does not support RC4 and MD5, and I'd like to use a known good crypto provider instead of the native one in qpdf.

This isn't a feature request, I am willing to implement this myself, just want to discuss this before submitting a major change.

TinyServal:
This has now been tested locally with MbedTLS 2.28.8 and 3.6.3. While the final 2.28 LTS version (2.28.10) was released 3 weeks ago and this series is officially EoL, Ubuntu 24.04 and Debian 12 are still shipping 2.28.x, so it's important to maintain compatibility with these two mainstream LTS distros.

This provider uses the native RC4 implementation since it's no longer supported by MbedTLS v3.

Fixes #1427.


TinyServal:
Hm, this doesn't work on mbedtls 2.28, I'll see if I can do something about it.

jberkenbilt:
> Hm, this doesn't work on mbedtls 2.28, I'll see if I can do something about it.

Most likely, we would have a test in CI that would just pull down a more recent mbedtls. I don't imagine this will be enabled by default in any of the distributions, so I think it's okay for us to have a more recent lower bound.

TinyServal:
If you're okay with not supporting 2.28 then I can try to build a more recent mbedtls release in the CI scripts as a workaround until ubuntu switches to the current 3.x series.

TinyServal:
Poked at mbedtls 2.28 and I couldn't get AES to work reliably, I'll push what I have right now and continue troubleshooting.

TinyServal:
This is ready for review now, it's been fully tested with both mbedtls 2.28 and 3.6.

jberkenbilt:
Thanks for all your hard work. I would like to review this even if @m-holger also reviews it, and I'm kind of slow because I have limited qpdf time. I promise I will give this the attention it deserves. The earliest I will look at it will be this weekend.

TinyServal:
Sounds good, I'll push the documentation updates shortly and you can approve the CI workflows again.

libqpdf/QPDFCrypto_mbedtls.cc
```diff
@@ -0,0 +1,187 @@
...
+{
+    size_t output_len = 0;
+    check_mbedtls(mbedtls_platform_entropy_poll(nullptr, data, len, &output_len));
+}
+
+void
+QPDFCrypto_mbedtls::MD5_init()
+{
+    mbedtls_md_init(&md_ctx);
+    check_mbedtls(mbedtls_md_setup(&md_ctx, mbedtls_md_info_from_type(MBEDTLS_MD_MD5), 0));
```
TinyServal, line 53: Just found out there's a memory leak here, need to get to the bottom of it at some point.

libqpdf/QPDFCrypto_mbedtls.cc
```diff
@@ -0,0 +1,187 @@
...
+{
+    memset(md_out, 0, sizeof(md_out));
+    mbedtls_md_init(&md_ctx);
+    mbedtls_cipher_init(&cipher_ctx);
+}
+
+QPDFCrypto_mbedtls::~QPDFCrypto_mbedtls()
+{
+    mbedtls_md_free(&md_ctx);
+    mbedtls_cipher_free(&cipher_ctx);
```
TinyServal, line 39: Freeing these two contexts unconditionally is a safe operation, as far as I can tell.

jberkenbilt:
I added the `next` label in hopes of finding time before the next release review this and get it in/iterate on it. I want to get a release out with accumulated fixes and don't have time this weekend (or probably next) to give this the attention it deserves. It will get the attention though!

TinyServal:
Yeah, I also need some time to track down that memory leak before this can be merged anyway.

PR #1437, opened 2025-04-18
Detect shared /Kids arrays in pages tree

m-holger:
In QPDF::getAllPagesInternal include the /Kids array in the visited set for loop detection.

Fixes oss-fuzz issue 411312393


m-holger:
@jberkenbilt  , I am going to merge despite the CI failure. It seems pretty clear that this is a mingw problem which hopefully will sort itself out.

Issue #1056, opened 2023-10-24
`generateAppearancesIfNeeded` does not allow enabled values other than "/Yes" for checkboxes


dmjohnsson23:
The `QPDFAcroFormDocumentHelper::generateAppearancesIfNeeded` (or, more accurately, `QPDFFormFieldObjectHelper::setCheckBoxValue`) method does not allow enabled values other than "/Yes". However, this does not work for all PDFs. For example, [here](https://www.va.gov/vaforms/va/pdf/VA9.pdf) is a PDF that uses "/1" as the enabled value. Checkboxes should be allowed to be set to any value in the "/AP" dictionary, similar to how radio buttons work.

I was able to track the issues to these two lines, but I'm not a C programmer, so I thought it best to let someone else make the actual changes. (I'm using QPDF by way of PikePdf, rather than directly)
* https://github.com/qpdf/qpdf/blob/81823f4032caefd1050bccb207d315839c1c48db/libqpdf/QPDFFormFieldObjectHelper.cc#L307
* https://github.com/qpdf/qpdf/blob/81823f4032caefd1050bccb207d315839c1c48db/libqpdf/QPDFFormFieldObjectHelper.cc#L419

I've never encountered a checkbox that used anything other than "/Off" for the disabled value, but the enabled value appears to be a bit a a wild west.

jberkenbilt:
Thanks -- I should be able to work this in.

jberkenbilt:
@dmjohnsson23 Thanks for your analysis. This made my work very easy. I verified in the spec as well -- the old functionality was incorrect. Thanks for taking the time to report this. It will be fixed for 11.7.0.

amargerandhw:
Hello @jberkenbilt, thanks for your awesome project !
I think the fix done here https://github.com/qpdf/qpdf/pull/1089 is not complete as the method `QPDFFormFieldObjectHelper::setCheckBoxValue` still set `/Yes` when you pass true. I think it should be the corresponding true value of the checkbox.
If I'm not mistaken that could be done by checking the keys returned by `AP.getKey("/N").getKeys()` ?

Thank you in advance

dmjohnsson23:
This may or may not be useful to you, but here is how I ended up working around this bug to set the checkbox appearance:

https://github.com/dmjohnsson23/pdform/blob/432dc1474d552e1eea1f07dad8c729e22e0ea610/pdform/model/form_field.py#L240

[...]

PR #1439, opened 2025-04-19
Implement QPDFFormFieldObjectHelper::isChecked

jbarlow83:
API was defined, but .cc had no implementation. PDF 2.0 manual is surprisingly unclear as to on/off values, giving /Yes in an example and /Off in descriptive text as "not on".

### DIFF
--- a/libqpdf/QPDFFormFieldObjectHelper.cc
+++ b/libqpdf/QPDFFormFieldObjectHelper.cc
@@ -248,6 +248,12 @@ QPDFFormFieldObjectHelper::isCheckbox()
 }
 
 bool
+QPDFFormFieldObjectHelper::isChecked()
+{
+    return isCheckbox() && getValue().isName() && (getValue().getName() != "/Off");
+}
+
+bool
 QPDFFormFieldObjectHelper::isRadioButton()
 {
     return ((getFieldType() == "/Btn") && ((getFlags() & ff_btn_radio) == ff_btn_radio));


jbarlow83 (commented):
An alternative would be to drop ::isChecked if it's ill-defined.

m-holger:
@jbarlow83 , my reading of the spec is that /Off is not checked, anything else is checked.

jberkenbilt:
> @jbarlow83 , my reading of the spec is that /Off is not checked, anything else is checked.

I'm pretty sure that's right. See #1056 .

jberkenbilt:
It's interesting that isChecked sat there in the API for all these years without anyone ever noticing that I had never implemented it. That's telemetry at its best, I guess.

jberkenbilt:
I will accept this once I figure out what's messed up about mingw builds. I doubt it's related to this, but I'm going to push another PR soon and will see.

m-holger:
> I doubt it's related to this,

It isn't. Bumping to 12.1.1 fails, as does rebuilding my fork of qpdf which passed CI originally.

jberkenbilt:
If I had to guess, I'd say it's line endings, but we'll see. Ordinarily I'd just wait it out, but I was hoping to get 12.1.1 out.

jberkenbilt:
My current PR is not failing on Windows. I've rerun the failed jobs. Perhaps github broke the build image and has subsequently repaired it. We'll see.

PR #1443, opened 2025-04-23
zlib-flate: add option skip=N to ignore leading bytes

iam-TJ:
Allow ignoring leading bytes in the stream before beginning the
    [un]compress operation.

    This tool has many valuable uses beyond PDF processing. There is no
    similar tool for raw zlib streams widely available; The entire Debian
    archive lacks any others for example.

    Use-case: extracting Android's native 'backup' tool archives that
    contain a 24-byte header followed by a raw zlib stream. E.g:

     $ adb shell bu backup -shared > shared.backup
     $ cat shared.backup | zlib-flate -skip=24 -uncompress | tar --list -f -

One can of course use `dd bs=24 skip=1 ...`  in the pipeline but having this ability native in `zlib-flate` feels more polished and makes integrating it into more complex analysis and/or data-recovery tool-kits much cleaner.

### DIFF
--- a/zlib-flate/zlib-flate.cc
+++ b/zlib-flate/zlib-flate.cc
@@ -13,7 +13,8 @@ static char const* whoami = nullptr;
 void
 usage()
 {
-    std::cerr << "Usage: " << whoami << " { -uncompress | -compress[=n] }" << std::endl
+    std::cerr << "Usage: " << whoami << " [ -skip=N ] { -uncompress | -compress[=n] }" << std::endl
+              << "  skip=N will discard first N bytes of input before doing [un]compress" << std::endl
               << "If n is specified with -compress, it is a zlib compression level from"
               << std::endl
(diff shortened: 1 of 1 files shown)


jberkenbilt:
Interestingly, I use zlib-flate for this purpose. But rather than building skip into zlib-flate, I'd suggest just using dd.
```
dd if=shared.backup ibs=24 skip=1 | zlib-flate -uncompress | ...
```
should probably work fine, and I'd rather not add the extra feature to zlib-flate. But you're right -- I've used zlib-flate to decide raw streams in all sorts of applications. While I also haven't found other CLI tools for this (not that I've looked very hard), it's also pretty easy to do programmatically in various scripting languages that have raw zlib support. But I think the `dd` command above will work well enough.

jberkenbilt:
Thanks again for the suggestion. I'm closing this as `wontfix` because I think just using dd is adequate and we shouldn't add extra complexity to zlib-flate.

PR #1447, opened 2025-04-27
Fix QPDFFormFieldObjectHelper::getChoices (fixes #1433)

### LINKED ISSUE #1433: `QPDFFormFieldObjectHelper::getChoices()` returns empty list
QPDF seems to only return results that are strings, which is the cause of the issue:

https://github.com/qpdf/qpdf/blob/ac94437ae9d295efe991f148b7609c5f67580203/libqpdf/QPDFFormFieldObjectHelper.cc#L276

However, this is not correct. According to the PDF spec, the `/Opt` array may contain *either* strings, or `[export_value, display_value]` pairs. See Table 231 in 12.7.4.4 of the [PDF spec](https://opensource.adobe.com/dc-acrobat-sdk-docs/pdfstandards/PDF32000_2008.pdf).

An example PDF: [dd0293.pdf](https://github.com/user-attachments/files/19780689/dd0293.pdf)

This may be difficult to fix without changing the signature of the method, which is perhaps why it is implemented the way that it is, but the current implementation is not correct. Perhaps in this case, the existing method could return the export value, and a new method, `getChoicePairs()` could be created with a more correct implementation?

m-holger:
Return the display value if the choices entry is an array of strings rather than a single string.

Test file is need-appearances.pdf modified to contain one array entry.

### DIFF
--- a/include/qpdf/QPDFFormFieldObjectHelper.hh
+++ b/include/qpdf/QPDFFormFieldObjectHelper.hh
@@ -147,7 +147,7 @@ class QPDFFormFieldObjectHelper: public QPDFObjectHelper
     // Returns true if fields if of type /Ch
     QPDF_DLL
     bool isChoice();
-    // Returns choices as UTF-8 strings
+    // Returns choices display values as UTF-8 strings
     QPDF_DLL
     std::vector<std::string> getChoices();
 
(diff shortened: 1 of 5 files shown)


m-holger:
Open issues:

- what is the permanent solution for returning export values?

- how do we handle invalid entries? Simply ignoring them results in the returned array not alligning with the choices.

jberkenbilt (approved):
Approving as an improvement with no apparent downside over the existing behavior, even if this may not be the ultimate solution.

Issue #1068, opened 2023-11-20
qpdf-11.6.3-bin-linux-x86_64.zip doesn't work with AWS Lambda as layer


deathemperor:
I follow [this instruction](https://github.com/qpdf/qpdf/issues/352#issuecomment-1343316831) to create lambda layer with 11.6.3

got this runtime error:

```
qpdf: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.26' not found (required by /opt/lib/libqpdf.so.29)
```

jberkenbilt:
I guess my CI environment for building the lambda layer uses libraries that are too new now. I'll take a look at this before the next release.

deathemperor:
@jberkenbilt thanks. until that comes do you have any suggestion on how to build it manually? 

Running the last OK version I got this (that's why I upgraded to 11.6.3 trying to fix)

```
WARNING: /tmp/96393646-9ec5-4db6-b0ee-b2025780d483.pdf: file is damaged
WARNING: /tmp/96393646-9ec5-4db6-b0ee-b2025780d483.pdf: can't find startxref
WARNING: /tmp/96393646-9ec5-4db6-b0ee-b2025780d483.pdf: Attempting to reconstruct cross-reference table
qpdf: /tmp/96393646-9ec5-4db6-b0ee-b2025780d483.pdf: unable to find trailer dictionary while recovering damaged file
```

jberkenbilt:
Sorry for the delay -- this finished a long time ago but I got distracted by my day job. :-)
[qpdf-11.6.3-bin-linux-x86_64-ci.zip](https://github.com/qpdf/qpdf/files/13419075/qpdf-11.6.3-bin-linux-x86_64-ci.zip)
If this linux binary works for you, then you can use it as a place-holder until I release 11.6.4 or 11.7.0 or whatever is next. I may still want to release this as 11.6.4, but it might be a few days before I can do it, particularly with the holiday (in the US).

deathemperor:
no problem @jberkenbilt . good thing is I've tested and it works! thanks for the quick help.

PR #1456, opened 2025-05-08
Bump version to 12.3.0 and C++ to C++-20

m-holger:


m-holger:
@jberkenbilt  , the Linux used the build the AppImage (chosen in response to #1068)  does not support C++-20 .

jberkenbilt:
Okay, I will deal with that. I have some ideas. The AppImage needs to be built with an older distribution for compatibility reasons, but there are options that I can explore. We should hold on merging C++-20 until I deal with it -- perhaps I will have time this weekend, but if not, it won't be too long.

m-holger:
@jberkenbilt , merge when you are ready

jberkenbilt:
I've spent an hour or so researching and experimenting and have decided to make a pragmatic decision: we will change the AppImage build to be based off Ubuntu 22.04. This means that the AppImage will be less useful on older systems, as will the binary distribution that derives from it. However, the binary distribution primary exists to be used in AWS Lambda, which now uses AL2023, which has a new enough glibc/libstdc++ to support something built with 22.04. For people who want to run on older systems, they will just have to use 12.2.0 or below.

If you feel like attempting the AppImage build update, feel free, but I would also be fine with you just commenting it out in CI and merging this, and I will make sure to reactivate it before we release 12.3. There are usually some subtle things about updating the AppImage version. I don't remember off hand what they are, but I do this kind of thing all the time and am sure it would be light work for me to do it. I most likely will not have time this weekend.

If you decide to comment out the failing build in CI, please put something in the release notes at the top of the list reminding us to fix it before 12.3.0 is released.

jberkenbilt:
actually, I'm going to set a 20 minute timer and see if I can knock it out

m-holger:
I am happy to wait for now. If not bumping C++ starts causing extra work I'll merge with the suggested CI fix.

jberkenbilt:
We'll see if #1459 passes CI. It works for me locally. I have no idea if anyone actually uses the AppImage. One tiny data point is that the version on AppImagHub is 7 years old, and the AppImage got 9 downloads this week from SourceForge vs. 580ish for Windows and 90ish for source.

jberkenbilt:
Closing in favor of #1459

Issue #1464, opened 2025-05-13
Pipeline_private.hh: error: unknown type name 'qpdf_offset_t'


zclifford:
I ran into a small compilation error in 12.0 that I'll send a pull request for in a bit. It looks like `Pipeline_private.hh` is simply missing an `#include <qpdf/Types.h>`.

I'm not using cmake (Google means everything is Bazel) so it's possible there's something wrong with my build script, though it didn't look like it to me.

I send pull requests infrequently enough that I always forget how to do it every time.

```
third_party/qpdf/src/libqpdf/qpdf/Pipeline_private.hh:127:9: error: unknown type name 'qpdf_offset_t'
  127 |         qpdf_offset_t
      |         ^
third_party/qpdf/src/libqpdf/qpdf/Pipeline_private.hh:140:9: error: unknown type name 'qpdf_offset_t'
  140 |         qpdf_offset_t count{0};
      |         ^
third_party/qpdf/src/libqpdf/qpdf/Pipeline_private.hh:109:38: error: unknown type name 'qpdf_offset_t'
  109 |                 count += static_cast<qpdf_offset_t>(len);
      |                                      ^
third_party/qpdf/src/libqpdf/qpdf/Pipeline_private.hh:111:21: error: cannot initialize object parameter of type 'const Pipeline' with an expression of type 'qpdf::pl::Count'
  111 |                     next()->write(buf, len);
      |                     ^~~~
third_party/qpdf/src/libqpdf/qpdf/Pipeline_private.hh:119:17: error: cannot initialize object parameter of type 'const Pipeline' with an expression of type 'qpdf::pl::Count'
  119 |             if (next()) {
      |                 ^~~~
third_party/qpdf/src/libqpdf/qpdf/Pipeline_private.hh:121:21: error: cannot initialize object parameter of type 'const Pipeline' with an expression of type 'qpdf::pl::Count'
  121 |                     next()->write(reinterpret_cast<unsigned char const*>(str->data()), str->size());
      |                     ^~~~
third_party/qpdf/src/libqpdf/qpdf/Pipeline_private.hh:123:17: error: cannot initialize object parameter of type 'const Pipeline' with an expression of type 'qpdf::pl::Count'
  123 |                 next()->finish();
      |                 ^~~~
third_party/qpdf/src/libqpdf/qpdf/Pipeline_private.hh:130:38: error: unknown type name 'qpdf_offset_t'
[...]

PR #1465, opened 2025-05-13
Pipeline_private.hh should include qpdf/Types.h for qpdf_offset_t

zclifford:
Pipeline_private.hh should include qpdf/Types.h for qpdf_offset_t

https://github.com/qpdf/qpdf/issues/1464

### DIFF
--- a/libqpdf/qpdf/Pipeline_private.hh
+++ b/libqpdf/qpdf/Pipeline_private.hh
@@ -4,6 +4,7 @@
 #include <qpdf/Pipeline.hh>
 
 #include <qpdf/Pl_Flate.hh>
+#include <qpdf/Types.h>
 
 namespace qpdf::pl
 {


jberkenbilt:
We're supposed to have an internal policy of including private header files "first" in their implementations to catch these things. Maybe Pipeline doesn't do this, or maybe we lost it when splitting header files into *_private. Thanks.

m-holger:
The private header-only and is not actually use by Pipeline.cc nor does have an implementation. I am a bit puzzled by this.

zclifford:
Hmm, if it's puzzling let me get a bit more details then, one sec. I did mention I'm using bazel instead of cmake, so it's possible I'm doing something not right :D

zclifford:
OK some more details:

As I mentioned I'm using Bazel instead of cmake. Our bazel setup requires listing "textual_hdrs" (https://bazel.build/reference/be/c-cpp#cc_library) seperately. Our environment is a bit unlike most C++ environments, where all headers are allowed to be "textual".

Textual headers are headers that cannot be self contained. The classic example are files like `include/qpdf/auto_job_c_att.hh` in qpdf, they are designed to be included from other headers and usually lack an ifndef guard.

Here the compilation succeeds if I treat `Pipeline_private.hh` as a textual header, but fails otherwise. It fails because `Pipeline_private.hh` is not "standalone" (e.g. it references a type that it does not include). I neglected to consider that it might be a textual header, because it has an ifndef guard.

Indeed the more verbose compiler error shows that it was compiling this header by itself (I don't know why we do things this way). I have ommitted a screen full of clang but the failing command boiled down to

```
clang  -c third_party/qpdf/src/libqpdf/qpdf/Pipeline_private.hh -o blaze-out/k8-fastbuild/bin/third_party/qpdf/_objs/qpdf/Pipeline_private.h.processed
```

It indeed looks like it's trying to compile it by itself. I'm happy to treat this as my problem and treat it as a textual header if you're not interested in it declaring all it's dependencies. And will be more careful to check for this in the future.

Sorry for the noise!

m-holger:
I agree Types.h should have  been included.

However, Pipeline_private.hh is included by QPDFWriter_private.hh, which includes QPDFWriter.hh as the first include which in turn include DLL.h and Types.h as the first two includes. It is also included in QPDF_linerazition.cc  which includes QPDF_private.hh as the first include, which includes QPDF.hh first,   which  include DLL.h and Types.h as the first two includes. Either way, Types.h should have been found.

m-holger:
> if you're not interested in it declaring all it's dependencies

Dependencies should be declared. If you spot any others please let us know.

zclifford:
Right but the problem was that I was trying to compile it not through QPDFWriter_private.hh (or  QPDF_linearization.cc) but by itself.

If you compile the header _by itself_ rather than starting from a translation unit that includes it then it has no dependency on Types and this is why it failed. Standalone it includes recursively: qpdf/Pipeline.hh, qpdf/DLL.h, qpdf/QPDFLogger.hh, memory, string, functional, iostream, and memory.

Yeah it's a pretty weird setup to compile headers seperate from the translation units that include them but I do think that's what's going on...

PR #1468, opened 2025-05-19
Fix apparent typo for zopfli enablement

thinkpanzer:
cmake complains when building with -DZOPFLI-ON as opposed to =ON


jberkenbilt:
thnaks for notcign the tpyo.

PR #1486, opened 2025-07-02
Refactor Pl_AES_PDF and Pl_RC4

m-holger:


libqpdf/QPDFWriter.cc
```diff
@@ -844,8 +844,10 @@ QPDFWriter::setEncryptionParametersInternal(
 void
 QPDFWriter::setDataKey(int objid)
 {
-    m->cur_data_key = QPDF::compute_data_key(
-        m->encryption_key, objid, 0, m->encrypt_use_aes, m->encryption_V, m->encryption_R);
+    if (m->encrypted) {
+        m->cur_data_key = QPDF::compute_data_key(
```
m-holger: @jberkenbilt , am I missing something? Are there any situations where we have a cur_data_key when the file is not encrypted?

PR #1490, opened 2025-07-09
Modernize QPDFWriter write methods

m-holger:


m-holger:
@jberkenbilt , could you review the first commit.

Disclaimer: I disavow all knowledge of pkg-config.

m-holger:
@jberkenbilt  , I am merging this as is. We can amend once you had a chance to review it.

jberkenbilt:
I defer to you on this and don't intend to weigh in. I don't know about C++ concepts or fully follow the code.

jberkenbilt:
@m-holger I think this is incorrect -- the cmake files installed my qpdf should take care of this. The fact that this fails indicates a problem with packaging itself, not a problem with the test. The point of the package test is to make sure that qpdf, as packaged, works with a minimal cmake file. This shows it does not,.

PR #1491, opened 2025-07-11
Deprecate QPDF::compute_encryption_key

m-holger:


m-holger:
> Is your intention to just make these private? 

I am trying to remove as much implementation detail from the public header files as possible. This is partly because of #1567 , but also because makes it difficult to achieve greater encapsulation and locality without severe contortions because of the required  order of  include files.

PR #1510, opened 2025-08-05
Update release notes: document changes to stream EOL handling, indire…

m-holger:
…ct object limits, and xref `/Size` behavior

### DIFF
--- a/manual/release-notes.rst
+++ b/manual/release-notes.rst
@@ -62,6 +62,9 @@ more detail.
    - Option :qpdf:ref:`--check` now includes additional basic checks of the
      AcroForm, Dests, Outlines, and PageLabels structures.
 
+    - New `--remove-javascript` option to disable JavaScript/ECMAScript actions
+      in the output PDF.
+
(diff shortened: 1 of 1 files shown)


m-holger:
@jberkenbilt  , just spotted the `--newline-before-endstream` option. For some reason CLion failed to mention `QPDFWriter::setNewlineBeforeEndstream`


image

I still think we should make this the default behaviour given that it is recommended in the spec.

jberkenbilt:
> I still think we should make this the default behaviour given that it is recommended in the spec.

I've been historically conservative about changing defaults and gratuitously breaking people's tests or breaking indempotency, but if there are things like this that we should do, I'm willing to relax on that. Probably improving the quality of PDFs generated by qpdf outweighs my other concerns. I have no objection to making that the default, but it probably needs to be called out in the release notes. I have tried in the past to call out changes that alter qpdf's output for the same previous input.

jberkenbilt (approved):
These are all fine.

PR #1535, opened 2025-08-31
Refactor `QPDFAcroFormDocumentHelper`: simplify `traverseField`, remo…

m-holger:
…ve `visited` parameter, update conditional checks, and streamline logic for field and annotation handling.

Tighten validation of the PDF spec requirements that form fields form a tree and that widget annotations have a single parent by utilizing `field_to_annotations` and `annotation_to_field` in the loop detection check.

### DIFF
--- a/include/qpdf/QPDFAcroFormDocumentHelper.hh
+++ b/include/qpdf/QPDFAcroFormDocumentHelper.hh
@@ -226,8 +226,7 @@ class QPDFAcroFormDocumentHelper: public QPDFDocumentHelper
 
   private:
     void analyze();
-    void traverseField(
-        QPDFObjectHandle field, QPDFObjectHandle parent, int depth, QPDFObjGen::set& visited);
+    void traverseField(QPDFObjectHandle const& field, QPDFObjectHandle const& parent, int depth);
     QPDFObjectHandle getOrCreateAcroForm();
     void adjustInheritedFields(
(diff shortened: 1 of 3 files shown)


m-holger:
@jberkenbilt , it looks to me that the loop detection in `traverseField` is in reality a detection of the violation of the various uniqueness requirements of the PDF spec, which may or may not indicate a loop.

Given the uniqueness requirements, loops are impossible in a PDF file provided  the requirement is properly enforced and therefore enforcement eliminates the need for loop detection. This leaves the question of what to do if a violation is detected. Simply warn and ignore does not feel like the best option.

jberkenbilt:
> Given the uniqueness requirements, loops are impossible in a PDF file provided the requirement is properly enforced and therefore enforcement eliminates the need for loop detection. This leaves the question of what to do if a violation is detected. Simply warn and ignore does not feel like the best option.

What do you think we should do instead? Do we have files that suffer from these problems? I think loop detection is needed in the code if the code does a traversal, and the presence of QTC for that indicates that it can happen. (This is part of what I like about QTC -- seeing a coverage call tells me immediately that it's covered without having to go elsewhere and look at a coverage report.) Anyway, I see you've removed that, and obviously it's not crashing, so if you've reworked the logic to avoid loops, that's fine.

I'm not sure exactly what you want my input on.

PR #1546, opened 2025-09-15
Avoid unnecessary re-compression in QPDFJob::doCheck

jbarlow83:
While investigating https://github.com/ocrmypdf/OCRmyPDF/issues/1570, I discovered that QPDFJob::doCheck does some unnecessary work by re-compressing some streams to Flate. This is mostly noticeable on larger PDFs that contain a lot of JPEGs where DCT decode -> Flate encode uses a lot CPU time and memory. 

Recompressing in this way should not affect the quality of checks, since we only need to confirm we can decode all present streams.

### DIFF
--- a/libqpdf/QPDFJob.cc
+++ b/libqpdf/QPDFJob.cc
@@ -793,6 +793,11 @@ QPDFJob::doCheck(QPDF& pdf)
         Pl_Discard discard;
         w.setOutputPipeline(&discard);
         w.setDecodeLevel(qpdf_dl_all);
+        // Disable compression of streams, since we only need to confirm we can decode them. This
+        // avoids JPEG DCT -> Flate recompression of all images.
+        w.setCompressStreams(false);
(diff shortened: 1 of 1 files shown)


m-holger:
@jbarlow83  , thanks. #1547 should take care of the CI failure. If you make the requested amendment I'll merge later.

m-holger:
> https://github.com/qpdf/qpdf/pull/1547 should take care of the CI failure

Probably after rebasing.

OliverWich:
Nice, thank you!

Is there a way I can try this out?

(And if not, do you know when this will land in a main release? :D)

m-holger:
@OliverWich 

> Is there a way I can try this out?

The pikepdf documentation contains https://pikepdf.readthedocs.io/en/latest/source_build.html on how to build pikepdf against a qpdf source tree.

>  do you know when this will land in a main release?

I am not sure when the next qpdf release is going to be. Hopefully within the next couple of month.

Issue #1543, opened 2025-09-11
c++20 features in public headers make using in projects written in previous c++ standards impossible


bhennion:
On the dev version (at least since 485399c71403f49f4becd625b0df6e17ce59f439 but there may be other commits touching other header files), it is no longer possible to use QPDF from a project compiled with a previous c++ standard.
It will break every project that is not compiled/written in c++20 and depends on QPDF once you release v12.3.0.

This is due to c++20 features being used in public headers.
In the case of 485399c71403f49f4becd625b0df6e17ce59f439: the concept std::integral is used in QPDFWriter.hh. Note that it is used in the `private` part of a class, so it does not add anything to the interface. This private part should not lie in the public headers.

Again, there may be other public headers like that, I haven't checked.

m-holger:
I think `<concepts>` in `QPDFWriter.hh` is the only header new in C++-20. I'll remove this.

bhennion:
It's also probably quite easy to have a CI test that checks it: a `main.cpp` including all the pubilc headers and which is compiled against c++11 (or whichever earliest standard you want to support).

m-holger:
> It's also probably quite easy to have a CI test that checks it: a main.cpp including all the pubilc headers

The thought did occur to me too.

> against c++11 (or whichever earliest standard you want to support)

C++-17 is required to read the public headers. From 12.3, C++20 will be required to build qpdf.

m-holger:
#1611 adds a CI test for >C++17 includes in public header files. Closing as completed.

PR #1567, opened 2025-10-27
Maintain C++-17 compatibility for public header files

m-holger:
Continued from #1490 .

@jberkenbilt , I agree that bumping the requirement for pkg-test to C++-20 was not a good idea, especially in view of the points made by  @bhennion in #1543. I think it makes sense to maintain C++-17 compatibility for public header files for as long as possible, and at the moment I don't have anything on my radar that would significantly benefit from C++-20 in public headers.

I have updated the release notes accordingly.

If we are committing to C++-17 compatibility for public header files for the time being then we should test for it in CI. I have updated `qpdf-version.cc` to pull in all public header files and required C++-17 compliance in the CMake file. I don't know how pkg-test is used by Debian and whether this has repercussions there - if it does, or if you feel that the change interferes with the purpose of testing that qpdflib works with a minimal CMake file we can leave pkg-test unchanged and implement this as a separate CI test.

### DIFF
--- a/pkg-test/CMakeLists.txt
+++ b/pkg-test/CMakeLists.txt
@@ -1,6 +1,5 @@
 cmake_minimum_required(VERSION 3.10)
 project(qpdf-version LANGUAGES CXX)
-set(CMAKE_CXX_STANDARD 20)
 find_package(qpdf)
 add_executable(qpdf-version qpdf-version.cc)
 target_link_libraries(qpdf-version qpdf::libqpdf)
--- a/manual/release-notes.rst
+++ b/manual/release-notes.rst
@@ -16,7 +16,10 @@ more detail.
 12.3.0: not yet released
   - Build changes
 
-    - A C++-20 compiler is now required.
+    - A C++-20 compiler is now required to build or test ``qpdf``. All public header
+      files remain C++-17 compatible. Until further notice, any future header files
+      that require C++-20 will be flagged in the release notes and their use will be
+      optional.
 
     - The AppImage and Linux standalone binary distributions are now
       built with Ubuntu 22.04 to support C++-20, which means they will
--- a/pkg-test/qpdf-version.cc
+++ b/pkg-test/qpdf-version.cc
@@ -1,9 +1,72 @@
+#include <qpdf/Buffer.hh>
+#include <qpdf/BufferInputSource.hh>
+#include <qpdf/ClosedFileInputSource.hh>
+#include <qpdf/Constants.h>
+#include <qpdf/DLL.h>
+#include <qpdf/FileInputSource.hh>
+#include <qpdf/InputSource.hh>
+#include <qpdf/JSON.hh>
+#include <qpdf/ObjectHandle.hh>
+#include <qpdf/PDFVersion.hh>
+#include <qpdf/Pipeline.hh>
+#include <qpdf/Pl_Buffer.hh>
+#include <qpdf/Pl_Concatenate.hh>
+#include <qpdf/Pl_Count.hh>
+#include <qpdf/Pl_DCT.hh>
+#include <qpdf/Pl_Discard.hh>
+#include <qpdf/Pl_Flate.hh>
+#include <qpdf/Pl_Function.hh>
+#include <qpdf/Pl_OStream.hh>
+#include <qpdf/Pl_QPDFTokenizer.hh>
+#include <qpdf/Pl_RunLength.hh>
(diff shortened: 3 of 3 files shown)


pkg-test/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.10)
 project(qpdf-version LANGUAGES CXX)
-set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD 17)
```
jberkenbilt: I think it would be better to omit this entirely. I think it's possible for us to make the cmake package files we install indicate what version of C++ we require, and that's the right place for it. The goal of this test is that a minimal cmake file should continue to work when you reinstall the package. While a project using qpdf probably wants to do this, the package test probably shouldn't.

jberkenbilt:
> If we are committing to C++-17 compatibility for public header files for the time being then we should test for it in CI. I have updated `qpdf-version.cc` to pull in all public header files and required C++-17 compliance in the CMake file. I don't know how pkg-test is used by Debian and whether this has repercussions there - if it does, or if you feel that the change interferes with the purpose of testing that qpdflib works with a minimal CMake file we can leave pkg-test unchanged and implement this as a separate CI test. I don't think the pkg-test check is really strong enough or the right place.

We have a quick test in CI for C++ next. We could also have a quick test for C++17 perhaps just by having a file that includes the main public headers, perhaps QPDF.hh and QPDFWriter.hh, which is probably enough for most people. Since it won't have to compile the whole library, it should be very fast. Actually, the `build-linux` script for CI has an extra test for signed char. I would be okay with adding something to build-linux that uses the just-built library to build and link a program with C++-17 that includes QPDF.hh and QPDFWriter.hh. That would show that, even with the library built with C++-20, a C++-17 compiler can include headers and link and get a usable application. Probably we need QUtil.hh as well as that's a very likely place where we might accidentally add C++-20.

There some time and filesystem related things I wanted to do with C++-20, but I don't think those require any changes to public headers....but if I weren't careful, I'd break QUtil.hh.

pkg-test/CMakeLists.txt
m-holger: I have removed the entry.

m-holger:
I have removed the entry in the CMake file, but left the includes for the public header files in, While he  g++ defaults to C++-17 that should pick up any C++-20 use that may slip in.

>  We could also have a quick test for C++17 perhaps just by having a file that includes the main public headers

I think the safest option would be to check all public headers, and ideally check each header individually not only for C++-17 compliance but also for completeness as a standalone include. Perhaps iterate over all the public header files, generate a simple program consisting of one include and a simple and try to compile. 

Mentioning header files, as a result of moving implementation detail out of the public headers, we can tidy the includes in those headers by dropping some of them (either entirely pr replacing them with forward declarations). Is that something we want to do now or do we want to wait for a major release?

jberkenbilt:
> Mentioning header files, as a result of moving implementation detail out of the public headers, we can tidy the includes in those headers by dropping some of them (either entirely pr replacing them with forward declarations). Is that something we want to do now or do we want to wait for a major release?

CLion makes this pretty easy. I'd do it now so we can catch problems before a major release. It should be totally safe, but there may be some weird cases where we are getting some headers indirectly that we should be including directly where compiler implementation differences bite us. This used to be a huge problem -- I used to get bitten by it every compiler version update. It's been quite a long time -- I think the C++ standard has tightened this up a lot, and gcc and clang are both a lot more careful to avoid side effects. But it could still happen.

My rule has always been that each impl file includes its header first as a test for standalone, but that's hard to enforce. It might be worth actually having a little script that iterates through all the files in qpdf/include, with possibly some hard-coded exceptions if needed, and includes the file by itself as you say. I'd rather see this automated to ensure we never forget anything. Should be a light lift.

m-holger:
>  I'd rather see this automated to ensure we never forget anything. 

Me too

> Should be a light lift

Not for me. Both Perl and bash scripts are something I only handle under duress and with a lot of head-scratching.

> My rule has always been that each impl file includes its header first as a test for standalone

This  requires 1 - 1 or 1 - n header to implementation file relationship, which is difficult to achieve while attempting a separation between public interface and private implementation detail and while giving the compiler a decent chance at optimisation. I think checks are the way forward.

jberkenbilt:
I can agree to take this one on before we release.

m-holger:
That would be great

PR #1587, opened 2025-11-14
Utilize the GNUTLS_FIPS140_LAX around MD5 initialization.

### LINKED ISSUE #1566: QPD::compute_data_key fails in FIPS mode on Ubuntu 24.04 (and others)
I have been looking into a problem reported by one of Ubuntu users.

After enabling FIPS mode they are unable to print using cups.

Minimal reproducer:
```
pro attach # to be able to enable FIPS mode
pro enable fips-updates
reboot # to boot the fips kernel
# with FIPS mode enabled
/usr/lib/cups/filter/pdftopdf 555 $USER title 1 "" /usr/share/cups/data/confidential.pdf
```

The output ends with:
```
ERROR: cfFilterPDFToPDF: Exception: gnutls: MD5 error: An algorithm that is not enabled was negotiated.
ERROR: pdftopdf filter function failed.
```

I have traced it down to:
```
#0  gnutls_hash_init (dig=0x5555555ad818, algorithm=GNUTLS_DIG_MD5) at ../../lib/crypto-api.c:839
#1  0x00007ffff75b0e83 in QPDFCrypto_gnutls::MD5_init (this=0x5555555ad810) at /usr/src/qpdf-11.9.0-1.1build1/libqpdf/QPDFCrypto_gnutls.cc:46
#2  0x00007ffff756f309 in MD5::MD5 (this=0x7fffffffb8e0) at /usr/src/qpdf-11.9.0-1.1build1/libqpdf/MD5.cc:11
#3  QPDF::compute_data_key (encryption_key=..., objid=<optimized out>, generation=0, use_aes=<optimized out>, encryption_V=<optimized out>, 
    encryption_R=0) at /usr/src/qpdf-11.9.0-1.1build1/libqpdf/QPDF_encryption.cc:352
#4  0x00007ffff7553509 in QPDFWriter::setDataKey (this=0x7fffffffbdb0, objid=<optimized out>)
    at /usr/src/qpdf-11.9.0-1.1build1/libqpdf/QPDFWriter.cc:845
#5  0x00007ffff755e683 in QPDFWriter::writeObject (this=this@entry=0x7fffffffbdb0, object=..., object_stream_index=object_stream_index@entry=-1)
    at /usr/src/qpdf-11.9.0-1.1build1/libqpdf/QPDFWriter.cc:1790
#6  0x00007ffff75603f2 in QPDFWriter::writeStandard (this=0x7fffffffbdb0) at /usr/src/qpdf-11.9.0-1.1build1/libqpdf/QPDFWriter.cc:3013
#7  QPDFWriter::write (this=0x7fffffffbdb0) at /usr/src/qpdf-11.9.0-1.1build1/libqpdf/QPDFWriter.cc:2200
#8  0x00007ffff7e8e12d in _cfPDFToPDFQPDFProcessor::emit_file (this=0x55555559b320, f=<optimized out>, doc=0x7fffffffbea0, take=<optimized out>)
    at cupsfilters/pdftopdf/qpdf-pdftopdf-processor.cxx:876
#9  0x00007ffff7e822a8 in cfFilterPDFToPDF (inputfd=<optimized out>, outputfd=<optimized out>, inputseekable=<optimized out>, data=<optimized out>, 
    parameters=<optimized out>) at cupsfilters/pdftopdf/pdftopdf.cxx:985
#10 0x00007ffff7f519bf in ppdFilterCUPSWrapper (argc=argc@entry=7, argv=argv@entry=0x7fffffffe398, 
    filter=0x7ffff7f4fce0 <ppdFilterPDFToPDF(int, int, int, cf_filter_data_t*, void*)>, parameters=parameters@entry=0x0, 
    JobCanceled=JobCanceled@entry=0x555555558014 <JobCanceled>) at ppd/ppd-filter.c:178
#11 0x00005555555550db in main (argc=7, argv=0x7fffffffe398) at filter/pdftopdf.c:66
```

Regardless of the FIPS mode QPDF tries to use MD5 hashing mechanism from gnutls - and since gnutls is FIPS-aware it refuses to return it because it's not FIPS approved.

`QPDF::compute_data_key` in `libqpdf/QPDF_encryption.cc` seems to be [unconditionally using MD5](https://github.com/qpdf/qpdf/blob/2ef866db3c00e5b1617cff7b3774bf5addcd4ca1/libqpdf/QPDF_encryption.cc#L452).

I'd like to avoid _hacking_ gnutls to make printing work.

I believe the best ways to proceed here could be:
* switch to a FIPS-approved algorithm that would make it work everywhere
* detect FIPS mode and replace MD5 with something else in that case

I would love to learn qpdf team view on this.

dargad:
Surrounding the gnutls_hash_init(&hash_ctx, GNUTLS_DIG_MD5) call with the convenience macros allows MD5 to be used internally even in FIPS 140 mode and should not affect at all systems not running FIPS.

Fixes #1566.

### DIFF
--- a/libqpdf/QPDFCrypto_gnutls.cc
+++ b/libqpdf/QPDFCrypto_gnutls.cc
@@ -11,9 +11,17 @@ QPDFCrypto_gnutls::QPDFCrypto_gnutls() :
     encrypt(false),
     cbc_mode(false),
     aes_key_data(nullptr),
-    aes_key_len(0)
+    aes_key_len(0),
+    fips_mode(gnutls_fips140_mode_enabled())
 {
     memset(digest, 0, sizeof(digest));
+
+    if (fips_mode) {
+        // Relax FIPS mode for the lifetime of this object
+        gnutls_fips140_set_mode(
+            GNUTLS_FIPS140_LAX,
+            GNUTLS_FIPS140_SET_MODE_THREAD);
+    }
 }
 
 QPDFCrypto_gnutls::~QPDFCrypto_gnutls()
@@ -26,6 +34,13 @@ QPDFCrypto_gnutls::~QPDFCrypto_gnutls()
     }
     aes_key_data = nullptr;
     aes_key_len = 0;
+
+    if (fips_mode) {
+        // Restore saved FIPS mode
+        gnutls_fips140_set_mode(
+            static_cast<gnutls_fips_mode_t>(fips_mode),
+            GNUTLS_FIPS140_SET_MODE_THREAD);
+    }
 }
 
 void
--- a/libqpdf/qpdf/QPDFCrypto_gnutls.hh
+++ b/libqpdf/qpdf/QPDFCrypto_gnutls.hh
@@ -53,6 +53,7 @@ class QPDFCrypto_gnutls: public QPDFCryptoImpl
     char digest[64];
     unsigned char const* aes_key_data;
     size_t aes_key_len;
+    unsigned fips_mode;
 };
 
 #endif // QPDFCRYPTO_GNUTLS_HH


m-holger:
@dargad  , thanks, this is very helpful and definite progress.

There are two questions on my mind:

-  Is there any danger that we accidentally drop into strict mode if the tread is currently in relaxed mode? Expanding the macros we currently have 
```cpp
if (gnutls_fips140_mode_enabled()) {              
	gnutls_fips140_set_mode(                 
		GNUTLS_FIPS140_LAX,              
		GNUTLS_FIPS140_SET_MODE_THREAD);
}
int code = gnutls_hash_init(&hash_ctx, GNUTLS_DIG_MD5);

if (gnutls_fips140_mode_enabled()) {              
			gnutls_fips140_set_mode(                 
				GNUTLS_FIPS140_STRICT,           
				GNUTLS_FIPS140_SET_MODE_THREAD); 
}
```  

I am wondering whether we should restore the current mode instead:

```cpp

unsigned current_fips_mode = 0;
if ((current_mode = gnutls_fips140_mode_enabled())) {              
	gnutls_fips140_set_mode(                 
		GNUTLS_FIPS140_LAX,              
		GNUTLS_FIPS140_SET_MODE_THREAD);
}
int code = gnutls_hash_init(&hash_ctx, GNUTLS_DIG_MD5);

if (gnutls_fips140_mode_enabled()) {               
			gnutls_fips140_set_mode(                 
				current_mode,           
				GNUTLS_FIPS140_SET_MODE_THREAD); 
}
```  

- Given that most other calls to GnuTLS when decrypting existing PDF documents will have the same problem, and given that QPDFCrypto_gnutls objects are short-lived, should we move the calls to the con/destructors?

dargad:
> I am wondering whether we should restore the current mode instead:
> 
> ```c++
> unsigned current_fips_mode = 0;
> if ((current_mode = gnutls_fips140_mode_enabled())) {              
> 	gnutls_fips140_set_mode(                 
> 		GNUTLS_FIPS140_LAX,              
> 		GNUTLS_FIPS140_SET_MODE_THREAD);
> }
> int code = gnutls_hash_init(&hash_ctx, GNUTLS_DIG_MD5);
> 
> if (gnutls_fips140_mode_enabled()) {               
> 			gnutls_fips140_set_mode(                 
> 				current_mode,           
> 				GNUTLS_FIPS140_SET_MODE_THREAD); 
> }
> ```
A believe this will be more future-proof if there's a chance there will be more than 1 place where switching the mode could occur.

> * Given that most other calls to GnuTLS when decrypting existing PDF documents will have the same problem, and given that QPDFCrypto_gnutls objects are short-lived, should we move the calls to the con/destructors?

If they are short-lived then it would make sense. Just to clarify: you mean to store the current mode in the constructor and switch to LAX mode and then switch back to the saved mode in the destructor implying the mode should be stored somewhere as a new QPDFCrypto_gnutls field?
If so - that makes sense and would protect the whole implementation from gnutls returning `NULL` for any other call in FIPS mode.

m-holger:
> implying the mode should be stored somewhere as a new QPDFCrypto_gnutls field?

yes

dargad:
I have tested changing the FIPS mode for the lifetime of the object. Seems to be working as expected.

m-holger:
@dargad , thank's again for taking this on

PR #1589, opened 2025-11-15
Introduce `global` namespace for managing qpdf global options and lim…

m-holger:
…its; add support for configurable global limits via `qpdf_global_get_uint32` and `qpdf_global_set_uint32`.


jberkenbilt:
@m-holger I skimmed this and am fine with the idea. I think I don't have the mindspace to review the whole thing, but if there is some specific thing you want me to comment on, let me know. I'm generally in favor of this change.

PR #1612, opened 2025-11-30
Document new release signing procedure

jberkenbilt:
@m-holger feel free to come in after me and change your release signing email address if I picked the wrong one, or if you don't want to be involved in releases, you can take your name off the list. I don't mind continuing to do releases, but if we both agree, I believe the new process removes the need for me to be the one to actually perform the release mechanics. I wouldn't release without your consent anyway.

### DIFF
--- a/README-maintainer.md
+++ b/README-maintainer.md
@@ -607,31 +607,33 @@ When done, the following should happen:
 
 ## CREATING A RELEASE
 
+* Releases are signed using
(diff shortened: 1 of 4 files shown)


jberkenbilt:
copilot pointed out that I am still using gpg -s to sign the release tags...I'm not going to fix that. If you ever do actually do a release, you can update the procedure to fit what you actually do. For now, I think this should suffice. I will merge when CI is green, and then I'm done for the day. We can always change it.

PR #1629, opened 2025-12-12
fix(doc): Fix a typo

fml2:


m-holger:
@fml2 , thanks for taking the trouble of correcting the manual. Much appreciated

PR #1631, opened 2025-12-21
Add 6 new fuzz targets for improved OSS-Fuzz coverage

skypher:
## Summary
- Add 6 new fuzzers targeting previously uncovered high-complexity code paths:
  - `qpdf_job_fuzzer`: QPDFJob JSON configuration parsing
  - `qpdf_transform_fuzzer`: Page transformations (flattenRotation, copyAnnotations, rotatePage)
  - `qpdf_embed_fuzzer`: Embedded file operations
  - `qpdf_annot_fuzzer`: Annotation handling and AcroForm operations
  - `qpdf_linearization_fuzzer`: Linearization checking and writing
  - `qpdf_object_fuzzer`: PDF object manipulation and tree operations

- Each fuzzer includes:
  - Appropriate memory limits for decompression pipelines
  - Seed corpus with relevant test files
  - Fuzzing dictionary where applicable (JSON keywords, annotation types)
  - Options file for max_len configuration

- Targets functions like `flattenRotation` (complexity 11,472) and `copyAnnotations` (complexity 11,205) that previously had 0% coverage per Fuzz Introspector

## Motivation
Current OSS-Fuzz coverage is ~52%. These fuzzers aim to improve coverage toward the 80% threshold by exercising previously untested code paths in page transformation, annotation handling, linearization, and object manipulation.

## Test plan
- [x] All fuzzers compile successfully with standalone runner
- [x] All fuzzers run without crashes on seed corpus files
- [x] CMakeLists.txt correctly routes each fuzzer to its specific seed corpus

### DIFF
--- a/fuzz/qpdf_linearization_fuzzer.cc
+++ b/fuzz/qpdf_linearization_fuzzer.cc
@@ -56,13 +56,10 @@ FuzzHelper::testLinearization()
     std::cerr << "info: isLinearized\n";
     bool is_linearized = q->isLinearized();
 
-    std::cerr << "info: checkLinearization\n";
-    q->checkLinearization();
-
     if (is_linearized) {
-        // Note: showLinearizationData() outputs to cout which creates noise,
-        // so we skip calling it. checkLinearization() exercises similar code paths.
-        std::cerr << "info: PDF is linearized, skipping showLinearizationData\n";
+        // checkLinearization() is only valid if isLinearized() returns true
+        std::cerr << "info: checkLinearization\n";
+        q->checkLinearization();
     }
 
     // Test writing with linearization enabled
--- a/fuzz/qpdf_object_fuzzer.cc
+++ b/fuzz/qpdf_object_fuzzer.cc
@@ -200,8 +200,7 @@ FuzzHelper::testTrees()
                             // Iterate through the tree to exercise traversal logic
                             int count = 0;
                             for (auto it = nth.begin(); it != nth.end() && count < 50; ++it, ++count) {
-                                it->first;  // Get key
-                                it->second; // Get value
+                                continue;
                             }
                         } catch (...) {
                             // Name tree errors expected
@@ -228,8 +227,7 @@ FuzzHelper::testTrees()
             // Iterate through the tree to exercise traversal logic
             int count = 0;
             for (auto it = nth.begin(); it != nth.end() && count < 50; ++it, ++count) {
-                it->first;  // Get key (page number)
-                it->second; // Get value (label dict)
+                continue;
             }
         }
     } catch (QPDFExc const& e) {


fuzz/qpdf_linearization_fuzzer.cc
```diff
@@ -0,0 +1,180 @@
...
+{
+    std::shared_ptr<QPDF> q = getQpdf();
+
+    // Test linearization checks
+    std::cerr << "info: isLinearized\n";
+    bool is_linearized = q->isLinearized();
+
+    std::cerr << "info: checkLinearization\n";
+    q->checkLinearization();
+
```
m-holger: It is invalid to call `q2->checkLinearization()` if `q2->isLinearized()` fails. This is is the reason for the "Fuzzing" CI failure.

fuzz/qpdf_object_fuzzer.cc
```diff
@@ -0,0 +1,293 @@
...
+                            std::cerr << "info: name tree " << key << '\n';
+                            QPDFNameTreeObjectHelper nth(names.getKey(key), *q);
+                            nth.find("test");
+                            nth.hasName("test");
+
+                            // Iterate through the tree to exercise traversal logic
+                            int count = 0;
+                            for (auto it = nth.begin(); it != nth.end() && count < 50; ++it, ++count) {
+                                it->first;  // Get key
+                                it->second; // Get value
```
m-holger: The other two CI failures are due to the expressions above being unused. Replace with a continue statement

skypher:
Great catches, added some fixes!

m-holger:
@skypher , thank you for your effort. I am afraid I haven't had time to review in detail, but I have 3 initial comments:

1. It would be much easier if you produced separate PRs for each new fuzzer. It would break down the review into more manageable chunks. I also don't really fancy adding 6 new fuzzers at the same time in case they are too good at identifying issues that need fixing (they identified one minor issue already).
2. I don't think the CMake changes are necessary. The existing fuzzers are named `qpdf_ ...` because they are the result of splitting a single `qpdf` fuzzer to avoid spurious time-outs. Simply drop the `qpdf_` prefix from your fuzzer names.
3. The new fuzzers should be added to CI testing. The last commit of #1634 does this for all 6 fuzzers. It should be pretty clear from that commit how to add a single fuzzer.

 I noticed that you did similar work for other projects. It would be interesting to know what the background to your work is.

skypher:
> @skypher , thank you for your effort. I am afraid I haven't had time to review in detail, but I have 3 initial comments:
> 
> 1. It would be much easier if you produced separate PRs for each new fuzzer. It would break down the review into more manageable chunks. I also don't really fancy adding 6 new fuzzers at the same time in case they are too good at identifying issues that need fixing (they identified one minor issue already).
> 2. I don't think the CMake changes are necessary. The existing fuzzers are named `qpdf_ ...` because they are the result of splitting a single `qpdf` fuzzer to avoid spurious time-outs. Simply drop the `qpdf_` prefix from your fuzzer names.
> 3. The new fuzzers should be added to CI testing. The last commit of [Add CI tests for new fuzzers in #1631 #1634](https://github.com/qpdf/qpdf/pull/1634) does this for all 6 fuzzers. It should be pretty clear from that commit how to add a single fuzzer.
> 
> I noticed that you did similar work for other projects. It would be interesting to know what the background to your work is.

Hi and happy new year!

And thanks also for showing such an interest in my little project. I've been an open source contributor for 20 years plus. A couple weeks ago I discovered the Google OSS Fuzz program and thought it was a great idea -- cost-neutral fuzzing for OSS projects. I curated a list of critical projects that either didn't have any fuzzing or seemed to have low coverage that could be improved, and qpdf is one of them!

Your feedback is really helpful. I'll put up a fresh iteration of this PR soon, incorporating the changes you've suggested.

skypher:
Closing to split into separate PRs per fuzzer as requested. Will submit 6 individual PRs with renamed fuzzers (dropping qpdf_ prefix) and CI tests.

m-holger:
@skyper , Happy New Year to you too.

Can I suggest you do one fuzzer first and than wait for comments before tackling the next one.

Based on my brief initial scan you have followed the existing fuzzers. This is very reasonable and, subject to review, I will be happy to accept that. That said, the existing fuzzers could do with refactoring and don't really follow current coding practice. If you want to volunteer to some of that on, that would also be very welcome. If so, just let me know and I can give you some pointers.

skypher:
> @Skypher , Happy New Year to you too.
> 
> Can I suggest you do one fuzzer first and than wait for comments before tackling the next one.
> 
> Based on my brief initial scan you have followed the existing fuzzers. This is very reasonable and, subject to review, I will be happy to accept that. That said, the existing fuzzers could do with refactoring and don't really follow current coding practice. If you want to volunteer to some of that on, that would also be very welcome. If so, just let me know and I can give you some pointers.

That's valuable info! Let me prepare an update. Which fuzzer do you think should be the first one to be added?

m-holger:
> Which fuzzer do you think should be the first one to be added?

Going by the areas currently worked on, transform and annot are most likely spot new issues early.

skypher:
> @Skypher , Happy New Year to you too.
> 
> Can I suggest you do one fuzzer first and than wait for comments before tackling the next one.
> 
> Based on my brief initial scan you have followed the existing fuzzers. This is very reasonable and, subject to review, I will be happy to accept that. That said, the existing fuzzers could do with refactoring and don't really follow current coding practice. If you want to volunteer to some of that on, that would also be very welcome. If so, just let me know and I can give you some pointers.

Could you elaborate on your plans for refactoring? I'd be happy to take this on.

m-holger:
> I'd be happy to take this on.

Great news! Thanks

> Could you elaborate on your plans for refactoring?

I'll come back on that early next week

m-holger:
@skypher, sorry, took me a bit longer than planned.

Assume that there is going to be a single call to a qpdf::global method to set all the limits currently set in the doChecks methods as well as  the call to qpdf->setMaxWarnings in getQPDF. This is going to be easier to maintain for us as well as making it easier for other projects that have qpdf as a dependency to fuzz their project rather than fuzzing qpdf and our dependencies. 

My priorities for refactoring the fuzzers are:

- add timer output (see info methods of the qpdf_pages_fuzzer) to aid in initial assessment of timeouts.

- reduce boiler plate. Much of the code is common to all fuzzers, and further code is common to all fuzzer that create a QPDF object / have a PDF file as input. Derive the various FuzzHelper classes from common base classes that implement the common code. 

- there is no need to have separate run and doChecks methods. As long as the test method called at the very end of doChecks is named consistenly, the combind run/doChecks method can be moved to the base class.  

- it is probably not a good idea create more than in QPDF object in a single run as this increases the change of meaningless timeouts, especially looking forward. It is highly likely that the number of checks following a xref reconstruction (the expected case during fuzzing) is going to increase, including expensive checks of complex structures. This means that the QPDF object can be created and configured when needed rather than having a separate method to create it, and we definitely don't need to use a shared pointer.

- a QPDF object can only be written once. Having a separate method to create the QPDFWriter and the use of shared pointers is also unnecessary.

- QPDFExc is derived from std::runtime_error and is easily recognisable. No need to treat them separately from run_time errors.

Note that nothing of the above is set in stone if you can think of something better.

Preferred coding style uses snake_case for method and variable names, trailing underscores for member variable and no unnecessary "this->".

skypher:
@m-holger That helps a lot, thanks! I'll get started with these today. It will take me a couple days. Btw, my username is "skypher" with an "h" :-)

m-holger:
>  Btw, my username is "skypher" with an "h" :-)

@skypher, ooops, sorry.

PR #1634, opened 2025-12-22
Add CI tests for new fuzzers in #1631

m-holger:


m-holger:
Demo only for how to add fuzzer CI tests

PR #1663, opened 2026-01-25
Add --deduplicate-image-xobjects to merge identical image XObjects

### LINKED ISSUE #448: Option to remove duplicated streams
When qpdf merges files, it may create files with multiple streams that have identical content, or such cases may exist for other reasons. Sometimes, this is intentional, and in fact, qpdf explicitly supports duplicating streams when merging pages in case someone wants to modify a copied page. However, in other cases, it may be desirable for a user to replace duplicated streams with a single stream and change all references to point to the single stream. If qpdf eventually learns how to do this, it will be explicitly selected. Ideas could include an operation on a QPDF to deduplicate all streams, an option on QPDF to deduplicate when adding new streams, or an option to identify identical streams.

This issue serves as a placeholder to track this feature, which has been requested by email.

qooxzuub:
This adds an optimization feature that identifies stream objects with identical content and identical dictionaries and merges them.

* Added `QPDF::deduplicateStreams` and helper logic in `QPDF_deduplicate.cc`.
* Added --deduplicate-streams CLI flag.
* Added regression tests.

Fixes #448


qooxzuub:
Demo below. We manufacture a file /tmp/ss.pdf with duplicate content streams and then deduplicate them, roughly halving the file size. 

(/tmp/s_copy.pdf is a copy of /tmp/s.pdf; if we don't make a copy then the output /tmp/ss.pdf will not have duplicate streams, I guess because qpdf notices it's the same input so can make optimizations.)

```bash
$ build/qpdf/qpdf --empty --pages /tmp/s.pdf /tmp/s_copy.pdf -- /tmp/ss.pdf
$ build/qpdf/qpdf /tmp/ss.pdf --deduplicate-streams /tmp/ss2.pdf
$ for x in s ss ss2; do pdf=/tmp/$x.pdf; echo $pdf; hexapdf usage $pdf ; done
/tmp/s.pdf
Content            5037B (4)
Fonts             175.8K (49/33)
Images            742.3K (2)
ObjectStreams      7065B (2)
Outline               0B (1/1)
XObjects          450.8K (7)
Total               1.4M (199)
/tmp/ss.pdf
Content             9.8K (8)
Fonts             371.0K (98)
Images              1.4M (4)
XObjects          901.6K (14)
Total               2.7M (384)
/tmp/ss2.pdf
Content             9.8K (8)
Fonts             363.1K (79)
Images            742.3K (2)
XObjects          450.8K (7)
Total               1.4M (297)
```
If you run with QPDF_DEDUPE_DEBUG set, it says that it consolidated 60 duplicate streams.

m-holger:
@qooxzuub , thank you for your effort. Unfortunately for the reason already given, I will not be reviewing any PR to remove all duplicated streams. 

I did however have a very brief look at your work and my initial view is that even if the PR was restricted to specific stream types it would require significant changes to be acceptable. If you plan to rework the PR let me know what your plans are and I can provide more detailed comments.

qooxzuub:
Thanks. Yes, I was quite naive; it is clear now that restricting to certain types of streams is necessary.

I am working on revising the feature along these lines:

1. rename it to deduplicateXobjects
2. if the PDF has /Encrypt, then never touch it
3. only ever deduplicate XObjects, and only consider those without a /Group key

So far I have implemented this but haven't yet updated the tests. I'm a bit lost with qpdf testing conventions generally.

Thank you for taking a look at the previous version. Any comments would be great. 

(I'm also not sure if it's better to close this PR and make a fresh one, or just commit here...)

qooxzuub:
I’ve squashed the history and reworked the feature per earlier feedback (thanks, @m-holger).

This now introduces `QPDF::deduplicateXobjects`, restricted to unencrypted PDFs and to XObject streams only, with explicit exclusions for context-sensitive cases (e.g. transparency groups, OC, structure).

Tests and release notes are included. Happy to adjust scope or naming if needed.

qooxzuub:
Re-running the demo with --deduplicate-xobjects instead of --deduplicate-streams:
```
/tmp/s.pdf
Content            5037B (4)
Fonts             175.8K (49/33)
Images            742.3K (2)
ObjectStreams      7065B (2)
Outline               0B (1/1)
XObjects          450.8K (7)
Total               1.4M (199)
/tmp/ss.pdf
Content             9.8K (8)
Fonts             371.0K (98)
Images              1.4M (4)
XObjects          901.6K (14)
Total               2.7M (384)
/tmp/ss2.pdf
Content             9.8K (8)
Fonts             363.1K (79)
Images            742.3K (2)
XObjects          450.8K (7)
Total               1.6M (316)
```
When run with QPDF_DEDUPE_DEBUG set, it says that it consolidated 9 duplicate streams.

m-holger:
@qooxzuub , unfortunately our coding conventions are not well documented, but you may want to look at  [code-formatting](https://github.com/qpdf/qpdf/blob/main/.github/copilot-instructions.md#code-formatting) and [coding-conventions](https://github.com/qpdf/qpdf/blob/main/.github/copilot-instructions.md#coding-conventions) for what little guidance there is.

 I am still not convinced that it is safe to merge widget appearances and therefore I do not want to include form XObjects at the moment. Let's limit this to image streams only.

We also need to keep in mind that de-duplicating streams is expensive and therefore we need to be efficient. In practical terms this means that we need to use qpdf's internal API.

One important aspect of this PR is the comparing two objects for equality by value. Could you do a separate PR for this. Rather than burying this functionality in the QPDF class, which represents a PDF document, please implement this as BaseHandle::equals. Can you have a look at BaseHandle::copy or BaseHandle::unparse as an example of how this should be implemented and let me have your questions. Can you also have a look at appendix J of the PDF spec for guidance on the comparison of numeric objects

libqpdf/QPDF_deduplicate.cc
```diff
@@ -0,0 +1,529 @@
...
+        return check_dict_equivalence(d1, d2, cache, depth);
+    }
+
+    return o1.unparse() == o2.unparse();
+}
+
+static bool
+check_dict_equivalence(
+    QPDFObjectHandle o1,
+    QPDFObjectHandle o2,
```
m-holger: avoid passing object handles by value if you can pass them as const&

libqpdf/QPDF_deduplicate.cc
```diff
@@ -0,0 +1,529 @@
...
+    EquivalenceCache& cache,
+    int depth)
+{
+    std::set<std::string> k1 = o1.getKeys();
+    std::set<std::string> k2 = o2.getKeys();
```
m-holger: getKeys is extremely expensive,it is much cheaper to iterate over the dictionaries

libqpdf/QPDF_deduplicate.cc
```diff
@@ -0,0 +1,529 @@
...
+
+    if (k1 != k2) {
+        return false;
+    }
+
+    for (const auto& key: k1) {
```
m-holger:
```cpp
for (auto const& [key, value]: Dictionary(o1)) {
    auto const& other = o2[key];
    if (value.size() != other.size()) {
        return false;
    }
    if (!are_equivalent(value, other .....
```

qooxzuub:
Thanks for the review and the pointers on coding conventions. I really appreciate the guidance.

1. Scope (images vs form xobjects)
Agreed, limiting to image xobjects is a much safer starting point.

2. Performance and internal API
That makes sense. I will switch to using the internal API for the traversal. (Since I'm new to the codebase, if you have a specific file or class in mind that demonstrates the preferred way to iterate dictionary keys/array items "raw," a pointer would be very helpful!)

4. Implementing annex J (equality logic)
Right, `unparse()` is insufficient because it compares syntax rather than semantics. I will revise the comparison logic to follow the annex J rules (promoting integers to reals, canonicalizing strings/names, and handling dictionary nulls).

5. `BaseHandle` implementation and naming
I am happy to move this logic to `BaseHandle`, but I have a proposal regarding the name and signature to address your performance concerns and the specific structure of PDF images.

I want to clarify that a shallow, non-recursive check (which 'equals' often implies) is insufficient even for images.

For example, consider two image xobjects I1 and I2 that are pixel-identical, pointing to SMask objects S1 and S2, respsectively, which are themselves identical to each other. A shallow equality check for I1, I2 would see different object IDs for the SMasks and return false, preventing the images from merging. To catch this, we must recurse into the SMask dictionary.

However, once we introduce recursion, we hit the safety warning in annex J.4.2: "recursion in dictionaries... has the potential to lead to an infinite recursive situation." Even if rare in images, we must protect against cycles.

Because of this, I propose naming the method `is_equivalent(other, &cache)`.

* Name: 'is_equivalent' distinguishes this deep structural graph traversal from a simple identity or shallow value check.
* Signature: It accepts an EquivalenceCache (or visited map). This allows us to handle the required recursion safely (breaking cycles) and efficiently (memoizing common sub-structures like SMasks so we don't re-compare them thousands of times).

Does this approach work for you?

m-holger:
See https://github.com/qpdf/qpdf/pull/1665#issuecomment-3873541645 for my comments

johnwhitington:
(Please forgive the drive-by comment, but this topic interests me.)

As far as I'm aware, there are only two places in PDF where de-duplicating objects is wrong:

1. The spec explicitly says an annotation may only be referenced from a single page, so top-level annotation objects may not be de-duplicated. There is no reason to think this extends to objects referenced from the annotations.
2. Page objects should not be de-duplicated. The spec says nothing about this, but a certain Well Known Implementation gets very confused if a page object is used more than once.

If you know of any others, please let me know - because it would mean my own software needs fixing...

m-holger:
> Please forgive the drive-by comment

any contributions are welcome

> As far as I'm aware, there are only two places in PDF where de-duplicating objects is wrong ... If you know of any others ...

Not specifically, at least not where two conforming objects could be identical. However. even where  the spec does not explicitly forbids sharing of objects it is not necessarily safe in practice to deduplicate, and I prefer to be cautious in this respect. In particular, I  will not touch anything related to the state of a widget annotation.

libqpdf/QPDF_deduplicate.cc
```diff
@@ -0,0 +1,105 @@
+#include <qpdf/ObjectHandle.hh>
+#include <qpdf/QPDF.hh>
```
m-holger: Formatting:

The primary header file is QPDF.hh - this should be the first include, followed by a blank line. Next should be the remaining project headers, another blank line and then the standard includes.

API:

Use the internal API, i.e. use QPDF_private.hh (instead of QPDF.hh) and   QPDFObjectHandle_private.

libqpdf/QPDF_deduplicate.cc
```diff
@@ -0,0 +1,105 @@
...
+#include <set>
+#include <vector>
+
+// ----------------------------------------------------------------------------
+// Helpers
+// ----------------------------------------------------------------------------
+
+static bool
+is_image_xobject(QPDFObjectHandle obj)
+{
```
m-holger: This is not required (see later comment). If it were
- there is allready a QPDFObjectHandle method for this
- this should be a method rather than a static function
- if it were a static function, the oh should be passed as a const&

libqpdf/QPDF_deduplicate.cc
```diff
@@ -0,0 +1,105 @@
...
+        if (current.isIndirect()) {
+            QPDFObjGen id = current.getObjGen();
+            if (visited_ids.count(id)) {
+                continue;
+            }
+            visited_ids.insert(id);
+        }
+
+        if (current.isDictionary()) {
+            for (const auto& key: current.getKeys()) {
```
m-holger: As an aside, never use getKeys as it is very expensive. Use
```cpp
if (Dictionary d = current) {
    for (auto const& [key, val]: d) {
        // ...
    }
}
```

libqpdf/QPDF_deduplicate.cc
```diff
@@ -0,0 +1,105 @@
...
+    }
+}
+
+// ----------------------------------------------------------------------------
+// Core Deduplication
+// ----------------------------------------------------------------------------
+
+void
+QPDF::deduplicateImageXobjects()
+{
```
m-holger: Rather than having a huge QPDF class, we are trying to use smaller component classes. This should be a method of QPDF::Doc::Objects::Streams

libqpdf/QPDF_deduplicate.cc
```diff
@@ -0,0 +1,105 @@
...
+    {
+        QPDFObjectHandle handle;
+        double length;
+    };
+
+    std::vector<ImageInfo> image_candidates;
+    for (auto& obj: this->getAllObjects()) {
+        if (is_image_xobject(obj)) {
+            auto len = obj.getDict().getKey("/Length").getNumericValue();
+            image_candidates.push_back({obj, len});
```
m-holger: Use emplace_back instead of push_back

libqpdf/QPDF_deduplicate.cc
```diff
@@ -0,0 +1,105 @@
...
+{
+    struct ImageInfo
+    {
+        QPDFObjectHandle handle;
+        double length;
+    };
+
+    std::vector<ImageInfo> image_candidates;
+    for (auto& obj: this->getAllObjects()) {
+        if (is_image_xobject(obj)) {
```
m-holger: `ìf (obj.Subtype() == "/Image") {` (needs #1685 )

libqpdf/QPDF_deduplicate.cc
```diff
@@ -0,0 +1,105 @@
...
+// ----------------------------------------------------------------------------
+// Core Deduplication
+// ----------------------------------------------------------------------------
+
+void
+QPDF::deduplicateImageXobjects()
+{
+    struct ImageInfo
+    {
+        QPDFObjectHandle handle;
```
m-holger: All images are streams. Use typed Stream rather than generic QPDFObjectHandle. (If we used   QPDFObjectHandle, convention would be to name it oh).

libqpdf/QPDF_deduplicate.cc
```diff
@@ -0,0 +1,105 @@
...
+        double length;
+    };
+
+    std::vector<ImageInfo> image_candidates;
+    for (auto& obj: this->getAllObjects()) {
```
m-holger: No "this". Don't create a new vector containing all objects. Iterate over m->obj_cache instead.
obj should be of Stream type.

libqpdf/QPDFObjectHandle.cc
```diff
@@ -449,7 +449,12 @@ BaseHandle::equivalent_to(BaseHandle const& other, int depth) const
             if (!s1.m->stream_dict.equivalent_to(s2.m->stream_dict, depth - 1)) {
                 return false;
             }
-            return s1.m->stream_data->view() == s2.m->stream_data->view();
+            // stream_data is only populated for replaceStreamData() streams;
+            // file-backed streams must be read via getRawStreamData().
+            if (s1.m->stream_data && s2.m->stream_data) {
+                return s1.m->stream_data->view() == s2.m->stream_data->view();
+            }
+            return Stream{obj}.getRawStreamData() == Stream{other.obj}.getRawStreamData();
```
m-holger, line 457: well spotted. ouch

libqpdf/QPDF_deduplicate.cc
```diff
@@ -0,0 +1,105 @@
...
+static bool
+is_image_xobject(QPDFObjectHandle obj)
+{
+    return obj.isStream() && obj.getDict().getKey("/Subtype").isNameAndEquals("/Image");
+}
+
+static void
+update_references_iterative(
+    std::vector<QPDFObjectHandle>& all_objects,
+    const std::map<QPDFObjGen, QPDFObjectHandle>& replacements)
```
m-holger: Visiting each object for references to stream duplicates would be prohibitively expensive. The approach to take is to replace the duplicate stream object with a QPDF_Reference to the other stream. This will need some tweaking to the object replacement code to ensure that obj_gen return by the reference is that of the underlying stream and not that of the duplicate.

As an aside, even if we were to visit all objects, we would not use a full tree search as this would be unnecessarily expensive. We would only use a full tree search if we wanted to ensure that we encountered objects in a specific order. This is not the case here, and we would therefore iterate over all objects, stopping  each search at the object boundary (i.e. wherever we encounter indirect references).

libqpdf/QPDF_deduplicate.cc
```diff
@@ -0,0 +1,105 @@
...
+    while (!work_stack.empty()) {
+        QPDFObjectHandle current = work_stack.back();
+        work_stack.pop_back();
+
+        if (current.isIndirect()) {
+            QPDFObjGen id = current.getObjGen();
+            if (visited_ids.count(id)) {
+                continue;
+            }
+            visited_ids.insert(id);
```
m-holger: As an aside, this should be
```cpp
if (!visited_ids.insert(current.id_gen()).second) {
    continue;
}
```
Also, avoid `count` to check for containment in sets and maps. C++20 provides `contains`.

libqpdf/QPDF_deduplicate.cc
```diff
@@ -0,0 +1,105 @@
...
+    struct ImageInfo
+    {
+        QPDFObjectHandle handle;
+        double length;
+    };
+
+    std::vector<ImageInfo> image_candidates;
+    for (auto& obj: this->getAllObjects()) {
+        if (is_image_xobject(obj)) {
+            auto len = obj.getDict().getKey("/Length").getNumericValue();
```
m-holger: Length is always an integer. Only use getNumericValue for objects that can be real. No need for the len variable. Also, use `Length()`, see #1685 .

libqpdf/QPDF_deduplicate.cc
```diff
@@ -0,0 +1,105 @@
...
+
+    std::map<QPDFObjGen, QPDFObjectHandle> replacements;
+    std::vector<ImageInfo> masters;
+
+    for (auto& candidate: image_candidates) {
+        bool found_master = false;
+        for (auto const& master: masters) {
+            if (candidate.length == master.length &&
+                candidate.handle.equivalent_to(master.handle)) {
+                replacements[candidate.handle.getObjGen()] = master.handle;
```
m-holger: Replace the duplicate stream with a reference immediately. No need for the masters vector or the found_masters var.

libqpdf/QPDF_deduplicate.cc
```diff
@@ -0,0 +1,105 @@
...
+        if (is_image_xobject(obj)) {
+            auto len = obj.getDict().getKey("/Length").getNumericValue();
+            image_candidates.push_back({obj, len});
+        }
+    }
+
+    std::map<QPDFObjGen, QPDFObjectHandle> replacements;
+    std::vector<ImageInfo> masters;
+
+    for (auto& candidate: image_candidates) {
```
m-holger: In the worst case of n streams, no duplicates you are doing n * (n/2) comparisons. Before iterating, sort image_candidates by size and note that you only need to check from the  current candidate until there is a change in length.

libqpdf/QPDF_deduplicate.cc
```diff
@@ -0,0 +1,105 @@
...
+            }
+        }
+        if (!found_master) {
+            masters.push_back(candidate);
+        }
+    }
+
+    if (!replacements.empty()) {
+        std::vector<QPDFObjectHandle> all_objs = this->getAllObjects();
+        all_objs.push_back(this->getTrailer());
```
m-holger: No need to check the trailer - it doesn't contain images.

m-holger (changes requested):
Sorry for the delay, I've been a snowed under with real life.

I've provided some initial comments. There are a couple more general points:

1. merging images is expensive and only worthwhile if it has a noticeable benefit. We will need a length threshold below which we would no merge, and an option for the user to set the threshold if he is not happy with our default.

2. if there is a private API method, avoid using the equivalent QPDFObjectHandle method. Have a look at what is defined in BaseHandle.

3. if an object has a specific type (such as stream) and the type is supported by a typed handle in QPDFObjectHandle_private.hh, prefer the typed handle over a generic QPDFObjectHandle.

libqpdf/QPDF_deduplicate.cc
qooxzuub: done

libqpdf/QPDF_deduplicate.cc
qooxzuub: done

libqpdf/QPDF_deduplicate.cc
qooxzuub: not done yet...

libqpdf/QPDF_deduplicate.cc
qooxzuub: done

libqpdf/QPDF_deduplicate.cc
qooxzuub: ok

libqpdf/QPDF_deduplicate.cc
qooxzuub: done

libqpdf/QPDF_deduplicate.cc
qooxzuub: done

libqpdf/QPDF_deduplicate.cc
qooxzuub: I tried QPDF_Reference but id_gen() returns the og of the slot rather than the reference target and I got stuck. So we still have a reference-rewriting pass. Can revisit if there's a way to make QPDF_Reference work correctly here?

libqpdf/QPDF_deduplicate.cc
qooxzuub: The QPDF_Reference issue tripped me up again here

libqpdf/QPDF_deduplicate.cc
qooxzuub: done

libqpdf/QPDF_deduplicate.cc
qooxzuub: done

zvezdochiot:
Hi @qooxzuub .

`--deduplicate-fonts` and other?

Example:
* [singh2017-orig.pdf](https://github.com/user-attachments/files/26882786/singh2017-orig.pdf) (94 fonts, 707858b)
* [singh2017-split_and_merge.pdf](https://github.com/user-attachments/files/26882795/singh2017-split_and_merge.pdf) (149 fonts, 987937b)

qooxzuub:
> Hi @qooxzuub .
> 
> `--deduplicate-fonts` and other?
> 
> Example:
> 
> * [singh2017-orig.pdf](https://github.com/user-attachments/files/26882786/singh2017-orig.pdf) (94 fonts, 707858b)
> * [singh2017-split_and_merge.pdf](https://github.com/user-attachments/files/26882795/singh2017-split_and_merge.pdf) (149 fonts, 987937b)

Yes, this is technically feasible, but not in this PR. I can take a look once this PR is done. (Earlier iterations of this PR did deduplicate every stream it found (including font data) but that is not safe. So the trick is to do it safely.)

libqpdf/QPDF_deduplicate.cc
qooxzuub: now done

qooxzuub:
> 1. merging images is expensive and only worthwhile if it has a noticeable benefit. We will need a length threshold below which we would no merge, and an option for the user to set the threshold if he is not happy with our default.

I added a threshold parameter but it is not exposed in the CLI and it defaults to 0. 0 is easier to test, and I don't know what the correct default should be. Happy to change any of this, of course.

PDFNewb:
Hi.  This feature is very interesting to me.  

I tested it on some very (very) large PDF files (20+GB) and it's extraordinarily slow -- the process seems almost entirely I/O bound, and uses relatively little RAM (compared to the size of the files) -- it takes 6 to 12 hours to process a single file on a server with 32CPUs and 256GB of RAM.  Even running the process inside a RAMdisk doesn't appear to make much of a difference.

Is it possible to make this more efficient by loading more components into RAM, and possibly reducing the number of comparisons?

I'm not a C++ programmer, so it's difficult for me to read the code and identify where the problem is, but I'm happy to assist with more testing on the data I have available to me.

qooxzuub:
Are you able to share a problematic file? Ideally not 20gb, a cut down version which is still a bit slow to process would be better,  say one that takes 15-30s. I can then try profiling to see what's going on.

PDFNewb:
Unfortunately, no.  This is confidential data.  However, the process is working very well, it's just slow.

I processed a 27GB file, consisting of a few thousand copies of the same form that contains unique data that we send to clients, which took 880 minutes to run, (even inside a RAMdisk!) -- but the resulting file was less than 500MB when it completed.

You could probably simulate this by merging several hundred copies of the same file that contains the same graphics.

qooxzuub:
> Unfortunately, no. This is confidential data. However, the process is working very well, it's just slow.
> 
> I processed a 27GB file, consisting of a few thousand copies of the same form that contains unique data that we send to clients, which took 880 minutes to run, (even inside a RAMdisk!) -- but the resulting file was less than 500MB when it completed.
> 
> You could probably simulate this by merging several hundred copies of the same file that contains the same graphics.

@PDFNewb please could you try the latest version? It runs much faster for me on a file like that.

PDFNewb:
Yup, 880+ minutes -> 5m14s.  Bravo!  Amazing job...  I/O's are down to practically nothing, and I can process one thread per CPU without any I/O wait.

Only other request is to add some output for when --verbose or --progress is enabled, since there's *zero* output during processing.

qooxzuub:
> Yup, 880+ minutes -> 5m14s. Bravo! Amazing job... I/O's are down to practically nothing, and I can process one thread per CPU without any I/O wait.

Great! Thanks for the real-world testing, this is useful.
 
> Only other request is to add some output for when --verbose or --progress is enabled, since there's _zero_ output during processing.

I added a simple announcement when running with --verbose. Percentage progress reporting during a long operation seems atypical for qpdf so I didn't implement that.

PR #1665, opened 2026-02-02
Implement BaseHandle::equivalent_to

qooxzuub:
Add deep structural comparison for PDF objects following ISO 32000-2 Annex J logic.

* Implement numeric promotion for integer and real types.
* Expose EquivalenceCache to allow memoization across multiple calls, supporting efficient deduplication workflows.
* Use cache to ensure termination on circular references and provide efficient graph traversal.
* Add convenience wrapper to call without providing a cache
* Add test suite in libtests/equivalence.

Per discussion with @m-holger in #1663.

### DIFF
--- a/libqpdf/QPDFObjectHandle.cc
+++ b/libqpdf/QPDFObjectHandle.cc
@@ -327,6 +327,148 @@ BaseHandle::copy(bool shallow) const
     return {}; // unreachable
 }
 
+// This method determines structural equivalence up to a given depth.
+// The default depth is 10.
+//
+// Nomenclature note: ISO 32000-2 Annex J uses the term "equal" for this
+// strict recursive comparison (J.4.1). We use "equivalent_to" here to
+// implement Annex J's "equality", distinguishing it from C++ shallow
+// pointer equality.
+//
+// Implementation notes:
+//
+// (1) We deviate from Annex J by comparing raw streams only, without
+// decoding.
+//
+// (2) Loop detection is expensive and is avoided. If either object has
+// a cycle in its forward orbit, this implementation will return false.
+
+bool
+BaseHandle::equivalent_to(BaseHandle const& other, int depth) const
+{
+    // A. Identity, size & limit checks
+    if (obj == other.obj) {
+        return true;
+    }
+    if (depth < 0) {
+        return false;
+    }
+    size_t size1 = size();
+    size_t size2 = other.size();
+    if (size1 != size2) {
+        return false;
+    }
+    // B. Structural comparison
+    qpdf_object_type_e t1 = resolved_type_code();
+    qpdf_object_type_e t2 = other.resolved_type_code();
+    if (t1 == ::ot_reference) {
+        return referenced_object().equivalent_to(other, depth - 1);
+    }
+    if (t2 == ::ot_reference) {
+        return equivalent_to(other.referenced_object(), depth - 1);
+    }
+    if (t1 != t2) {
+        if ((t1 == ::ot_integer || t1 == ::ot_real) && (t2 == ::ot_integer || t2 == ::ot_real)) {
+            // Numeric equivalence per Annex J
+            return oh().getNumericValue() == other.oh().getNumericValue();
+        }
+        // normalize uninitialized and null
+        return (t1 == ::ot_uninitialized && t2 == ::ot_null) ||
+            (t2 == ::ot_uninitialized && t1 == ::ot_null);
+    }
+    switch (t1) {
+    case ::ot_uninitialized:
+    case ::ot_null:
+        return true;
+    case ::ot_boolean:
+        return std::get<QPDF_Bool>(obj->value).val == std::get<QPDF_Bool>(other.obj->value).val;
+    case ::ot_string:
+        return std::get<QPDF_String>(obj->value).val == std::get<QPDF_String>(other.obj->value).val;
+    case ::ot_name:
+        return std::get<QPDF_Name>(obj->value).name == std::get<QPDF_Name>(other.obj->value).name;
+    case ::ot_array:
+        {
+            auto const& a1 = std::get<QPDF_Array>(obj->value);
+            auto const& a2 = std::get<QPDF_Array>(other.obj->value);
+            // sizes size1, size2 were calculated above and checked to be equal
+            if (!a1.sp && !a2.sp) {
+                for (size_t i = 0; i < size1; ++i) {
+                    if (!a1.elements[i].equivalent_to(a2.elements[i], depth - 1)) {
+                        return false;
+                    }
+                }
+                return true;
+            }
+            // at least one array is sparse
+            auto get_item = [](QPDF_Array const& arr, size_t idx) -> BaseHandle const& {
+                if (arr.sp) {
+                    auto it = arr.sp->elements.find(idx);
+                    if (it == arr.sp->elements.end()) {
+                        static QPDFObjectHandle null_oh = Null();
+                        return null_oh;
+                    }
+                    return it->second;
+                }
+                return arr.elements[idx];
+            };
+            for (size_t i = 0; i < size1; ++i) {
+                if (!get_item(a1, i).equivalent_to(get_item(a2, i), depth - 1)) {
+                    return false;
+                }
+            }
+            return true;
+        }
+    case ::ot_dictionary:
+        {
+            auto const& map1 = std::get<QPDF_Dictionary>(obj->value).items;
+            auto const& map2 = std::get<QPDF_Dictionary>(other.obj->value).items;
+            auto it2 = map2.begin();
+            auto end2 = map2.end();
+            for (auto const& [key1, value1]: map1) {
+                if (value1.null()) {
+                    continue;
+                }
+                while (it2 != end2 && it2->second.null()) {
+                    ++it2;
+                }
+                if (it2 == end2 || key1 != it2->first ||
+                    !value1.equivalent_to(it2->second, depth - 1)) {
+                    return false;
+                }
+                ++it2;
+            }
+            while (it2 != end2 && it2->second.null()) {
+                ++it2;
+            }
+            return it2 == end2;
+        }
+    case ::ot_stream:
+        {
+            auto const& s1 = std::get<QPDF_Stream>(obj->value);
+            auto const& s2 = std::get<QPDF_Stream>(other.obj->value);
+            if (!s1.m->stream_dict.equivalent_to(s2.m->stream_dict, depth - 1)) {
+                return false;
+            }
+            return s1.m->stream_data->view() == s2.m->stream_data->view();
+        }
+    case ::ot_operator:
+        throw std::logic_error("Internal error in BaseHandle::equivalent_to: found ot_operator");
+    case ::ot_inlineimage:
+        throw std::logic_error("Internal error in BaseHandle::equivalent_to: found ot_inlineimage");
+    case ::ot_integer:
+        return std::get<QPDF_Integer>(obj->value).val ==
+            std::get<QPDF_Integer>(other.obj->value).val;
+    case ::ot_real:
+        return oh().getNumericValue() == other.oh().getNumericValue();
+    case ::ot_unresolved: // cannot determine equivalence so return false
+    case ::ot_reference:  // handled above
+    case ::ot_destroyed:  // should not happen
+    case ::ot_reserved:   // should not happen
+        return false;
+    }
+    return false; // unreachable
+}
+
 std::string
 BaseHandle::unparse() const
 {
--- a/libtests/objects.cc
+++ b/libtests/objects.cc
@@ -304,6 +304,458 @@ test_2(QPDF& pdf, char const* arg2)
     assert(!default_limits());
 }
 
+// test equivalent_to
+static void
+test_3(QPDF& pdf, char const* arg2)
+{
+    // Scenario 1: Basic Equality: Name, Scalars
+    {
+        auto name = "/Test"_qpdf;
+        auto integer = Integer(42);
+        assert(name.equivalent_to("/Test"_qpdf));
+        assert(!name.equivalent_to(integer));
+    }
+    // Scenario 2: Numeric Types (Int vs Real)
+    {
+        auto integer = Integer(1);
+        auto real = QPDFObjectHandle::newReal("1.0");
+        assert(real.equivalent_to(integer));
+        assert(integer.equivalent_to(real));
+    }
+    // Scenario 3: Array Order Sensitivity
+    {
+        auto a1 = "[1 2]"_qpdf;
+        auto a2 = "[2 1]"_qpdf;
+        assert(!a1.equivalent_to(a2));
+        assert(!a2.equivalent_to(a1));
+    }
+    // Scenario 4: Dictionary Key Order Insensitivity
+    {
+        auto d1 = "<< /A 1 >>"_qpdf;
+        d1.replaceKey("/B", Integer(2));
+        auto d2 = "<< /B 2 >>"_qpdf;
+        d2.replaceKey("/A", Integer(1));
+        assert(d1.equivalent_to(d2));
+        assert(d2.equivalent_to(d1));
+    }
+    // Scenario 5: Direct vs Indirect Equality
+    {
+        auto obj = Integer(100);
+        auto indirect = pdf.makeIndirectObject(Integer(100));
+        assert(obj.equivalent_to(indirect));
+        assert(indirect.equivalent_to(obj));
+    }
+    // Scenario 6: Diamond Graph Isomorphism
+    {
+        auto d = pdf.makeIndirectObject(Integer(99));
+        auto b = pdf.makeIndirectObject(QPDFObjectHandle::newArray({d}));
+        auto c = pdf.makeIndirectObject(QPDFObjectHandle::newArray({d}));
+        assert(Array({b, c}).equivalent_to(Array({b, c})));
+    }
+    // Scenario 7: Circular References (Self-Loop): Compares as False
+    {
+        auto a1 = pdf.makeIndirectObject("[]"_qpdf);
+        a1.appendItem(a1);
+        auto a2 = pdf.makeIndirectObject("[]"_qpdf);
+        a2.appendItem(a2);
+        // The implementation rejects if there is any cycle, for performance reasons
+        assert(!a1.equivalent_to(a2));
+    }
+    // Scenario 8: Cross-Document Comparison (Objects from Different QPDF Instances)
+    {
+        QPDF pdf2;
+        pdf2.emptyPDF();
+        auto a1 = pdf.makeIndirectObject("[1]"_qpdf);
+        auto a2 = pdf2.makeIndirectObject("[1]"_qpdf);
+        auto a3 = pdf2.makeIndirectObject("[2]"_qpdf);
+        assert(a1.equivalent_to(a2));  // Same content, different documents
+        assert(a2.equivalent_to(a1));  // Same content, different documents
+        assert(!a1.equivalent_to(a3)); // Different content, different documents
+        assert(!a3.equivalent_to(a1)); // Different content, different documents
+    }
+    // Scenario 9: Stream Content: Match
+    {
+        assert(pdf.newStream("Stream data").equivalent_to(pdf.newStream("Stream data")));
+    }
+    // Scenario 10: Stream Content: Mismatch
+    {
+        auto s1 = pdf.newStream("Data A");
+        auto s2 = pdf.newStream("Data B");
+        assert(!s1.equivalent_to(s2));
+        assert(!s2.equivalent_to(s1));
+    }
+    // Scenario 11: Stream Dictionary Differences
+    {
+        auto s1 = pdf.newStream("same");
+        auto s2 = pdf.newStream("same");
+        s2.getDict().replaceKey("/Extra", QPDFObjectHandle::newName("/Value"));
+        assert(!s1.equivalent_to(s2));
+        assert(!s2.equivalent_to(s1));
+    }
+    // Scenario 12: J.3.6: Absent Keys vs Null
+    {
+        auto d0 = Dictionary::empty();
+        auto d1 = "<</Present null>>"_qpdf;
+        auto d2 = "<</Present << >> >>"_qpdf;
+        auto d3 = "<</Present [] >>"_qpdf;
+        assert(d0.equivalent_to(d1));
+        assert(d1.equivalent_to(d0));
+        assert(!d0.equivalent_to(d2));
+        assert(!d2.equivalent_to(d0));
+        assert(!d0.equivalent_to(d3));
+        assert(!d3.equivalent_to(d0));
+        assert(!d1.equivalent_to(d2));
+        assert(!d2.equivalent_to(d1));
+        assert(!d1.equivalent_to(d3));
+        assert(!d3.equivalent_to(d1));
+    }
+    // Scenario 13: String Syntax: Hex vs Literal (Annex J)
+    {
+        auto literal = "(A)"_qpdf;
+        auto hex = "<41>"_qpdf;
+        assert(literal.equivalent_to(hex));
+        assert(hex.equivalent_to(literal));
+    }
+    // Scenario 14: Name Syntax (Parser) vs Distinct Names (Model)
+    {
+        auto name1 = "/Name"_qpdf;
+        auto name2 = "/Na#6d#65"_qpdf;
+        assert(name1.equivalent_to(name2));
+        assert(name2.equivalent_to(name1));
+    }
+    // Scenario 15: Annex J Oddities: Keys, Octals, and Zeros
+    {
+        auto key1 = "<< /Key 1 >>"_qpdf;
+        auto key2 = "<< /K#65#79 1 >>"_qpdf;
+        auto lit_A = "(A)"_qpdf;
+        auto oct_A = "(\\101)"_qpdf;
+        auto zero_i = Integer(0);
+        auto zero_r = QPDFObjectHandle::newReal("-0.0");
+        auto r1 = QPDFObjectHandle::newReal("12.345");
+        auto r2 = QPDFObjectHandle::newReal("12.345000000000000");
+        auto i12 = Integer(12);
+        // note: we rely on double rounding here
+        auto r_lo = QPDFObjectHandle::newReal("11.99999999999999999999999999999999");
+        auto i12b = Integer(12);
+        auto r_hi = QPDFObjectHandle::newReal("12.00000000000000000000000000000000");
+        auto i1 = Integer(1);
+        auto r_1 = QPDFObjectHandle::newReal("1.");
+        assert(key1.equivalent_to(key2));
+        assert(key2.equivalent_to(key1));
+        assert(lit_A.equivalent_to(oct_A));
+        assert(oct_A.equivalent_to(lit_A));
+        assert(zero_i.equivalent_to(zero_r));
+        assert(zero_r.equivalent_to(zero_i));
+        assert(r1.equivalent_to(r2));
+        assert(r2.equivalent_to(r1));
+        assert(i12.equivalent_to(r_lo));
+        assert(r_lo.equivalent_to(i12));
+        assert(i12b.equivalent_to(r_hi));
+        assert(r_hi.equivalent_to(i12b));
+        assert(i1.equivalent_to(r_1));
+        assert(r_1.equivalent_to(i1));
+    }
+    // Scenario 16: Nested Containers
+    {
+        assert(Dictionary({{"/K", "[5]"_qpdf}}).equivalent_to(Dictionary({{"/K", "[5]"_qpdf}})));
+    }
+    // Scenario 17: Boolean and Null mismatch
+    {
+        auto b_true = QPDFObjectHandle::newBool(true);
+        auto b_false = QPDFObjectHandle::newBool(false);
+        auto null = QPDFObjectHandle::newNull();
+        auto null2 = QPDFObjectHandle::newNull();
+        auto one = Integer(1);
+        auto zero = Integer(0);
+        assert(null.equivalent_to(null));
+        assert(null.equivalent_to(null2));
+        assert(!b_true.equivalent_to(b_false));
+        assert(!b_true.equivalent_to(null));
+        assert(!b_true.equivalent_to(one));
+        assert(!b_true.equivalent_to(zero));
+        assert(!b_false.equivalent_to(null));
+        assert(!b_false.equivalent_to(one));
+        assert(!b_false.equivalent_to(zero));
+        assert(!null.equivalent_to(one));
+        assert(!null.equivalent_to(zero));
+        assert(!one.equivalent_to(zero));
+    }
+    // Scenario 18: Stream Semantics (J.3.7) - Strictness Check
+    {
+        auto s1 = pdf.newStream("test stream");
+        auto s2 = pdf.newStream("DIFFERENT_RAW_BYTES");
+        auto s3 = pdf.newStream("test stream");
+        s2.getDict().replaceKey("/Filter", QPDFObjectHandle::newName("/FlateDecode"));
+        s3.getDict().replaceKey("/Filter", QPDFObjectHandle::newName("/FlateDecode"));
+        assert(!s1.equivalent_to(s2));
+        assert(!s2.equivalent_to(s1));
+        assert(!s1.equivalent_to(s3));
+        assert(!s3.equivalent_to(s1));
+        assert(!s2.equivalent_to(s3));
+        assert(!s3.equivalent_to(s2));
+    }
+    // Scenario 19: Dictionary Value Type Mismatch
+    {
+        auto d1 = "<< /Key 1 >>"_qpdf;
+        auto d2 = "<< /Key (1) >>"_qpdf;
+        assert(!d1.equivalent_to(d2));
+        assert(!d2.equivalent_to(d1));
+    }
+    // Scenario 20: Mixed Direct vs Indirect Nesting
+    {
+        assert(
+            QPDFObjectHandle::newArray({Integer(7)})
+                .equivalent_to(QPDFObjectHandle::newArray({pdf.makeIndirectObject(Integer(7))})));
+    }
+    // Scenario 21: Dictionary Subset vs Superset
+    {
+        auto d1 = "<< /A 1 /B 2 >>"_qpdf;
+        auto d2 = "<< /A 1 >>"_qpdf;
+        assert(!d1.equivalent_to(d2));
+        assert(!d2.equivalent_to(d1));
+    }
+    // Scenario 22: Stream Semantic Decode Equivalence
+    {
+        auto s1 = pdf.newStream("Hello World");
+        auto s2 = pdf.newStream("HELLO WORLD RAW");
+        s2.getDict().replaceKey("/Filter", "/FlateDecode"_qpdf);
+        s2.getDict().replaceKey("/DecodeParms", Dictionary::empty());
+        assert(!s1.equivalent_to(s2));
+        assert(!s2.equivalent_to(s1));
+    }
+    // Scenario 23: Indirect Object Identity Independence
+    {
+        auto i1 = pdf.makeIndirectObject(Integer(123));
+        auto i2 = Integer(123);
+        assert(i1.equivalent_to(pdf.makeIndirectObject(Integer(123))));
+        assert(i1.equivalent_to(i2));
+        assert(i2.equivalent_to(i1));
+    }
+    // Scenario 24: Deep Recursive Structure (Stack Safety)
+    {
+        QPDFObjectHandle a1 = "[]"_qpdf;
+        QPDFObjectHandle a2 = "[]"_qpdf;
+        QPDFObjectHandle cur1 = a1;
+        QPDFObjectHandle cur2 = a2;
+        for (int i = 0; i < 200; ++i) {
+            auto n1 = "[]"_qpdf;
+            auto n2 = "[]"_qpdf;
+            cur1.appendItem(n1);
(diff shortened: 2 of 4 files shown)


libqpdf/QPDFObjectHandle.cc
```diff
@@ -2425,3 +2426,200 @@ operator""_qpdf(char const* v, size_t len)
...
+// strict recursive comparison (J.4.1) and reserves "equivalent" for looser
+// semantic matches (e.g., decoded streams). We use "is_equivalent" here to
+// implement Annex J's "equality", distinguishing it from C++ shallow
+// pointer equality.
+
+bool
+BaseHandle::is_equivalent(BaseHandle const& other, EquivalenceCache& cache) const
+{
+    // 1. Identity Optimization
+    if (this->obj == other.obj) {
```
m-holger: "this->" is redundant. Please remove

include/qpdf/QPDFEquivalenceCache.hh
```diff
@@ -0,0 +1,36 @@
+#ifndef QPDFEQUIVALENCECACHE_HH
+#define QPDFEQUIVALENCECACHE_HH
+
```
m-holger: private header files should only be created in libqpdf/qpdf, not in include/qpdf

include/qpdf/QPDFEquivalenceCache.hh
```diff
@@ -0,0 +1,36 @@
...
+     * @brief Memoization cache for is_equivalent.
+     * @warning Valid only while QPDFObject lifetimes are stable.
+     * Risk of ABA collision if objects are freed and reallocated.
+     */
+    struct EquivalenceCache
+    {
+        struct PointerPairHash
+        {
+            size_t
+            operator()(const std::pair<QPDFObject*, QPDFObject*>& p) const
```
m-holger: qpdf convention is "Type const&" reather than "const Type&"

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2425,3 +2426,200 @@ operator""_qpdf(char const* v, size_t len)
...
+        return true;
+    }
+    if (!this->obj || !other.obj) {
+        return false;
```
m-holger: I am not sure why you are return false in this case. `!obj` means that the object is null (see https://github.com/qpdf/qpdf/wiki/PDF-null-objects-vs-qpdf-null-objects)

For the purpose of this method I would suggest using the `size`method to check for null-like objects at this point:
```cpp
if (size() != other.size()) {
    return (t1 == ::ot_array) == (t2 == ::ot_array);
}
```

I don't think it is worth having runtime checks for destroyed or reserved objects here - use assertions (`util::qpdf_expect`).

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2425,3 +2426,200 @@ operator""_qpdf(char const* v, size_t len)
...
+    bool result = false;
+    qpdf_object_type_e t1 = this->type_code();
+    qpdf_object_type_e t2 = other.type_code();
+
+    auto oh1 = this->oh();
+    auto oh2 = other.oh();
+
+    // 3. Numeric Promotion (Integer == Real)
+    if ((t1 == ::ot_integer || t1 == ::ot_real) && (t2 == ::ot_integer || t2 == ::ot_real)) {
+        double v1, v2;
```
m-holger: There is no need to for this if  both objects are of the same type:
```cpp
if (t1 != t2) {
     if ((t1 == ::ot_integer || t1 == ::ot_real) && (t2 == ::ot_integer || t2 == ::ot_real)) {
        return oh().getNumericValue() == other.oh().getNumericValue() ;
    }
    return false;
}
```

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2425,3 +2426,200 @@ operator""_qpdf(char const* v, size_t len)
...
+        if (oh1.getValueAsNumber(v1) && oh2.getValueAsNumber(v2)) {
+            result = (v1 == v2);
+        }
+    } else if (t1 != t2) {
+        result = false;
+    } else {
+        // 4. Structural Comparison
+
+        switch (t1) {
```
m-holger: This switch should directly retrieve the values from the `QPDFObjects` rather than using the higher level accessor methods. Look at `BaseHandle::copy`or `unparse`as an example.

m-holger:
@qooxzuub , sorry for the delay in dealing with this. 

> Because of this, I propose naming the method is_equivalent 

I don't have a problem with the use of equivalent, but prefer "equivalent_to".

> Signature: It accepts an EquivalenceCache (or visited map). This allows us to handle the required recursion safely

I prefer not to use maps for loop detection as this is expensive (see https://github.com/qpdf/qpdf-dev/discussions/26).  We do not expect loops except in (most likely deliberately) damaged files and therefore restricting the depth of search is adequate.

> [EquivalenceCache]  allows us to handle ... efficiently ... common sub-structures like SMasks so we don't re-compare them thousands of times

Realistically, the only case where caching comparison results is likely to be beneficial is when comparing stream objects. What is the likelihood that different images share Masks, alternates or Metadata? I am not convinced caching is beneficial.

m-holger (changes requested):

libqpdf/QPDFObjectHandle.cc
qooxzuub: done

include/qpdf/QPDFEquivalenceCache.hh
qooxzuub: deleted

include/qpdf/QPDFEquivalenceCache.hh
qooxzuub: OK, hope I got them all.

(`libqpdf/QPDFObjectHandle.cc:Name::analyzeJSONEncoding(const std::string& name)` is not mine)

libqpdf/QPDFObjectHandle.cc
qooxzuub: Didn't understand the size() hint. null/uninitialized handling should be cleaner now.

```
+        if (t1 != t2) {
+            // normalize uninitialized and null
+            return (
+                (t1 == ::ot_uninitialized && t2 == ::ot_null) ||
+                (t2 == ::ot_uninitialized && t1 == ::ot_null));
+        }
```

libqpdf/QPDFObjectHandle.cc
qooxzuub: I fast-tracked all numeric comparisons now. I think they are very common. There should be no redundant calls.

```
+        // Numeric equivalence per Annex J
+        if ((t1 == ::ot_integer || t1 == ::ot_real) && (t2 == ::ot_integer || t2 == ::ot_real)) {
+            if (t1 == ::ot_integer && t2 == ::ot_integer) {
+                return std::get<QPDF_Integer>(h1.obj->value).val ==
+                    std::get<QPDF_Integer>(h2.obj->value).val;
+            }
+            return h1.oh().getNumericValue() == h2.oh().getNumericValue();
+        }
```

libqpdf/QPDFObjectHandle.cc
qooxzuub: Thanks, done

libqpdf/QPDFObjectHandle.cc
m-holger: On reflection, my suggestion should have been 
```cpp
if (size() != other.size()) {
    return false;
}
```

BaseHandle::size returns 0 for empty arrays and all null (and null-like) objects. For image streams there is no difference between an entry with an empty array and a missing entry, and I am pretty sure the same will apply in all other cases where we will want to merge objects

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2425,3 +2425,195 @@ operator""_qpdf(char const* v, size_t len)
...
+                return true;
+            }
+
+        case ::ot_dictionary:
+            {
+                auto const& map1 = std::get<QPDF_Dictionary>(h1.obj->value).items;
+                auto const& map2 = std::get<QPDF_Dictionary>(h2.obj->value).items;
+
+                if (map1.size() != map2.size()) {
+                    return false;
```
m-holger: This is not valid. Internally the maps can hold null objects, which are ignored for the comparison, therefore you can have equality even if the sizes don't match. 

This also means that the comparison loop will need to be adjusted to something like
```cpp
for (auto const& [key, value]: map1) {
    if (value.null()) {
        continue;
    }
    while (it2 != end2 && !it2->second.null()) {
        ++it2;
    }
    if (it2 == end2 || it2->first < key || !value.equivalent_to(it2->second, depth - 1)) {
        return false;
    } 
}
```

qooxzuub: done

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2425,3 +2425,200 @@ operator""_qpdf(char const* v, size_t len)
...
+BaseHandle::equivalent_to(BaseHandle const& other) const
+{
+    // Identity check
+    if (obj == other.obj) {
+        return true;
+    }
+
+    // usage of std::set avoids custom hash boilerplate for std::pair.
+    // O(log N) is acceptable here given the recursion depth limit.
+    using VisitedSet = std::set<std::pair<QPDFObjGen, QPDFObjGen>>;
```
m-holger: Loop detection using sets is expensive and we avoid it if possible. Limiting the search depth is sufficient, but reduce the limit. I think a depth of 10 will be more than sufficient for image streams.

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2425,3 +2425,200 @@ operator""_qpdf(char const* v, size_t len)
...
+
+    // Recursive lambda
+    auto internal = [](auto&& self,
+                       BaseHandle const& h1,
+                       BaseHandle const& h2,
+                       VisitedSet& visited,
+                       int depth) -> bool {
+        // A. Identity & limit checks
```
m-holger: There is no need for a lambda. Just call the method itself recursively

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2425,3 +2425,200 @@ operator""_qpdf(char const* v, size_t len)
...
+        if (h1.obj == h2.obj) {
+            return true;
+        }
+        if (depth < 0) {
+            return false;
+        }
+
+        qpdf_object_type_e t1 = h1.resolved_type_code();
+        qpdf_object_type_e t2 = h2.resolved_type_code();
```
m-holger: Contrary to what you are saying below, `::ot_reference` is a possible return value of `resolved_type_code`.

This would need to be handled at this point by calling `equivalent_to` with/on the underlying object.

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2425,3 +2425,200 @@ operator""_qpdf(char const* v, size_t len)
...
+
+        // Numeric equivalence per Annex J
+        if ((t1 == ::ot_integer || t1 == ::ot_real) && (t2 == ::ot_integer || t2 == ::ot_real)) {
```
m-holger: This test is only required if `t1 != t2`.  Move into the following if statement and let the switch statement take care of the equality case.

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2425,3 +2425,200 @@ operator""_qpdf(char const* v, size_t len)
...
+            if (t1 == ::ot_integer && t2 == ::ot_integer) {
+                return std::get<QPDF_Integer>(h1.obj->value).val ==
+                    std::get<QPDF_Integer>(h2.obj->value).val;
+            }
+            return h1.oh().getNumericValue() == h2.oh().getNumericValue();
+        }
+
+        if (t1 != t2) {
+            // normalize uninitialized and null
+            return (
```
m-holger: Drop the redundant outer brackets

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2425,3 +2425,200 @@ operator""_qpdf(char const* v, size_t len)
...
+                (t1 == ::ot_uninitialized && t2 == ::ot_null) ||
+                (t2 == ::ot_uninitialized && t1 == ::ot_null));
+        }
+
+        // B. Cycle detection
+        //
+        // Cycle detection tracks indirect objects (IDs) only.  PDF file
+        // format guarantees direct objects form trees; cycles require
+        // indirection.  Pathological in-memory direct cycles are invalid
+        // for serialization and caught by depth limit.
```
m-holger: As stated above, we don't want to use a set for loop prevention. FYI, there are a couple of issues with the code. At a basic level, that a source object descends into a destination object multiple times does not necessarily indicate a loop. On the other hand, even if the source object is no indirect a loop can occur.  If we wanted to use sets for loop detection we would need to use set of the og of the object descended into and pass the set by value to ensure that the set only contains ancestors only rather than all objects visited.

See also below.

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2425,3 +2425,200 @@ operator""_qpdf(char const* v, size_t len)
...
+        if (h1.obj->og.isIndirect() && h2.obj->og.isIndirect()) {
+            auto pair = std::make_pair(h1.obj->og, h2.obj->og);
```
m-holger: Similarly, this should use `h2.id_gen()`

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2425,3 +2425,200 @@ operator""_qpdf(char const* v, size_t len)
...
+                (t2 == ::ot_uninitialized && t1 == ::ot_null));
+        }
+
+        // B. Cycle detection
+        //
+        // Cycle detection tracks indirect objects (IDs) only.  PDF file
+        // format guarantees direct objects form trees; cycles require
+        // indirection.  Pathological in-memory direct cycles are invalid
+        // for serialization and caught by depth limit.
+        if (h1.obj->og.isIndirect() && h2.obj->og.isIndirect()) {
```
m-holger: `h2.obj` could be null which would cause a crash. This should be `if (h2.indirect()) {`.

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2425,3 +2425,200 @@ operator""_qpdf(char const* v, size_t len)
...
+
+    // Recursive lambda
+    auto internal = [](auto&& self,
+                       BaseHandle const& h1,
+                       BaseHandle const& h2,
+                       VisitedSet& visited,
+                       int depth) -> bool {
+        // A. Identity & limit checks
+        if (h1.obj == h2.obj) {
+            return true;
```
m-holger: ... which means this test is a duplicate

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2425,3 +2425,200 @@ operator""_qpdf(char const* v, size_t len)
...
+        // format guarantees direct objects form trees; cycles require
+        // indirection.  Pathological in-memory direct cycles are invalid
+        // for serialization and caught by depth limit.
+        if (h1.obj->og.isIndirect() && h2.obj->og.isIndirect()) {
+            auto pair = std::make_pair(h1.obj->og, h2.obj->og);
+            if (visited.count(pair)) {
+                return true;
+            }
+            visited.insert(pair);
+        }
```
m-holger: We would want to avoid using two separate calls for the check:
```cpp
if (!visited.insert(h2.id_gen()).second) {
    return false;
}
```
or combining into a single if statement:
```cpp
if (h2.indirect() && !visited.insert(h2.id_gen()).second) {
    return false;
}
```

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2425,3 +2425,200 @@ operator""_qpdf(char const* v, size_t len)
...
+        case ::ot_name:
+            return std::get<QPDF_Name>(h1.obj->value).name ==
+                std::get<QPDF_Name>(h2.obj->value).name;
+
+        case ::ot_array:
+            {
+                auto const& a1 = std::get<QPDF_Array>(h1.obj->value);
+                auto const& a2 = std::get<QPDF_Array>(h2.obj->value);
+
+                size_t s1 = a1.sp ? a1.sp->size : a1.elements.size();
```
m-holger: Use existing method: ` size_t s1 = a1.size()j`

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2425,3 +2425,200 @@ operator""_qpdf(char const* v, size_t len)
...
+                size_t s2 = a2.sp ? a2.sp->size : a2.elements.size();
+                if (s1 != s2) {
+                    return false;
+                }
+
+                auto get_item = [](QPDF_Array const& arr, size_t idx) -> BaseHandle {
```
m-holger: There would is no need to incur the overhead of creating a new BaseHandle - return a const& .

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2425,3 +2425,200 @@ operator""_qpdf(char const* v, size_t len)
...
+                    if (arr.sp) {
+                        auto it = arr.sp->elements.find(idx);
+                        return (it != arr.sp->elements.end())
```
m-holger: The brackets are redundant and should be omitted.

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2425,3 +2425,200 @@ operator""_qpdf(char const* v, size_t len)
...
+                            ? it->second
+                            : BaseHandle(QPDFObjectHandle::newNull());
```
m-holger: For the internal API, we never use QPDFObjectHandle::newNull() as a temporary null object. Typically, we would return an immutable null aka QPDFObjectHandle(). In this case, returning an explicit  null makes sense, but creating a new null object every time we need one does not. Use something like
```cpp
 static QPDFObjectHandle null_oh = Null();
 return it != arr.sp->elements.end() ? it->second : null_oh;
```

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2425,3 +2425,200 @@ operator""_qpdf(char const* v, size_t len)
...
+                    }
+                    return arr.elements[idx];
+                };
+
+                for (size_t i = 0; i < s1; ++i) {
```
m-holger: Sparse arrays are extremely rare. It would make sense to have one version of the loop for usual case and only use this version if one of the arrays is sparse.

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2425,3 +2425,200 @@ operator""_qpdf(char const* v, size_t len)
...
+            {
+                auto const& map1 = std::get<QPDF_Dictionary>(h1.obj->value).items;
+                auto const& map2 = std::get<QPDF_Dictionary>(h2.obj->value).items;
+
+                auto it1 = map1.begin();
+                auto it2 = map2.begin();
+                auto end1 = map1.end();
+                auto end2 = map2.end();
+
+                while (it1 != end1 || it2 != end2) {
```
m-holger: We try to only use iterators where it is necessary. Use a range based for loop to iterate over the current dictionary and use iterators for the other dictionary.

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2425,3 +2425,200 @@ operator""_qpdf(char const* v, size_t len)
...
+                    while (it1 != end1 && it1->second.null()) {
+                        ++it1;
+                    }
+                    while (it2 != end2 && it2->second.null()) {
+                        ++it2;
+                    }
+                    if (it1 == end1 || it2 == end2) {
+                        return (it1 == end1 && it2 == end2);
+                    }
+                    if (it1->first < it2->first || it1->first > it2->first) {
```
m-holger: `if (it1->first !=  it2->first ) {`

m-holger (changes requested):

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2425,3 +2425,200 @@ operator""_qpdf(char const* v, size_t len)
...
+                    ++it1;
+                    ++it2;
+                }
+                return true;
+            }
+
+        case ::ot_stream:
+            {
+                QPDFObjectHandle oh1(h1.obj);
+                QPDFObjectHandle oh2(h2.obj);
```
m-holger: Use qpdf::Stream rather than QPDFObjectHandle to call stream methods

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2425,3 +2425,200 @@ operator""_qpdf(char const* v, size_t len)
...
+
+                if (!self(self, oh1.getDict(), oh2.getDict(), visited, depth - 1)) {
+                    return false;
+                }
+
+                // Annex J deviation: compare compressed stream data only for efficiency.
```
m-holger: the remainder can be written more concisely as 
```cpp
return oh1.getRawStreamData()->view() == oh2.getRawStreamData()->view()
```

libtests/equivalence.cc
```diff
@@ -0,0 +1,831 @@
...
+        // -NDEBUG path
+        throw std::runtime_error("Verification failed");
+    }
+}
+
+// Helper to create a stream easily for tests
+QPDFObjectHandle
+make_stream(QPDF& pdf, std::string const& data)
+{
+    return pdf.makeIndirectObject(QPDFObjectHandle::newStream(&pdf, data));
```
m-holger: All streams are indirect objects. Replace calls to `make_stream(pdf, data)` with `pdf.newStream(data)`

libtests/equivalence.cc
```diff
@@ -0,0 +1,831 @@
+#include <qpdf/QPDF.hh>
+#include <qpdf/QPDFObjectHandle.hh>
+#include <qpdf/QUtil.hh>
+#include <cassert>
```
m-holger: including cassert is not allowed in qpdf code and will cause a CI failure. To use assert,  `#include <qpdf/assert_test.h>` is required as the first include. This will ensure that NDBUG is undefined.

libtests/equivalence.cc
```diff
@@ -0,0 +1,831 @@
...
+            assert_equivalent(roots.getArrayItem(0), roots.getArrayItem(1), false);
+        });
+    run(
+        "J.3.6: Absent Keys vs Null",
+        [](QPDF& pdf, QPDFObjectHandle& roots) {
+            auto d0 = QPDFObjectHandle::newDictionary();
+            auto d1 = QPDFObjectHandle::newDictionary();
+            auto d2 = QPDFObjectHandle::newDictionary();
+            auto d3 = QPDFObjectHandle::newDictionary();
+            d1.replaceKey("/Present", QPDFObjectHandle::newNull());
```
m-holger: This does not test what you think it tests. Calling `replaceKey` with a null object removes the entry and therefore the /Present key is absent in d1 rather than present with a null value.

libtests/equivalence.cc
```diff
@@ -0,0 +1,831 @@
...
+            roots.appendItem(d0);
+            roots.appendItem(d1);
+            roots.appendItem(d2);
+            roots.appendItem(d3);
+        },
+        [](QPDF& pdf, QPDFObjectHandle& roots) {
+            assert_equivalent(roots.getArrayItem(0), roots.getArrayItem(1), true);
+            assert_equivalent(roots.getArrayItem(0), roots.getArrayItem(2), false);
+            assert_equivalent(roots.getArrayItem(0), roots.getArrayItem(3), false);
+        });
```
m-holger: The whole run can be written more concisely as:
```cpp
    auto d0 = Dictionary::empty();
    assert(d0.equivalent_to(Dictionary({{"/Present", Null()}})));
    assert(!d0.equivalent_to(Dictionary({{"/Present", Dictionary::empty()}})));
    assert(!d0.equivalent_to(Dictionary({{"/Present", Array::empty()}})));
```

libtests/equivalence.cc
```diff
@@ -0,0 +1,831 @@
...
+    run(
+        "String Syntax: Hex vs Literal (Annex J)",
+        [](QPDF& pdf, QPDFObjectHandle& roots) {
+            roots.appendItem(QPDFObjectHandle::parse("(A)"));
+            roots.appendItem(QPDFObjectHandle::parse("<41>"));
+        },
```
m-holger: There is no real need for this here. You are testing the parser implementation rather than `equivalent_to`

libqpdf/QPDFObjectHandle.cc
qooxzuub: visited has been removed

libqpdf/QPDFObjectHandle.cc
qooxzuub: done

libqpdf/QPDFObjectHandle.cc
qooxzuub: done

qooxzuub (commented):
Thanks for your latest comments, hopefully now all fixed.

libqpdf/QPDFObjectHandle.cc
qooxzuub: I went with
```
+    case ::ot_stream:
+        {
+            auto const& s1 = std::get<QPDF_Stream>(obj->value);
+            auto const& s2 = std::get<QPDF_Stream>(other.obj->value);
+            if (!s1.m->stream_dict.equivalent_to(s2.m->stream_dict, depth - 1)) {
+                return false;
+            }
+            return s1.m->stream_data->view() == s2.m->stream_data->view();
+        }
```

libqpdf/QPDFObjectHandle.cc
qooxzuub: see above

libqpdf/QPDFObjectHandle.cc
qooxzuub: done

libqpdf/QPDFObjectHandle.cc
qooxzuub: I went with
```
+    if (t1 == ::ot_reference || t2 == ::ot_reference) {
+        auto q1 = qpdf();
+        auto q2 = other.qpdf();
+
+        auto r1 = (t1 == ::ot_reference && q1) ? q1->getObject(*this) : *this;
+        auto r2 = (t2 == ::ot_reference && q2) ? q2->getObject(other) : other;
+
+        if (r1.obj != obj || r2.obj != other.obj) {
+            return r1.equivalent_to(r2, depth - 1);
+        }
+
+        // reference resolution failed
+        return false;
+    }
```

libqpdf/QPDFObjectHandle.cc
qooxzuub: done

libqpdf/QPDFObjectHandle.cc
qooxzuub: done

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2425,3 +2425,153 @@ operator""_qpdf(char const* v, size_t len)
...
+        {
+            auto const& s1 = std::get<QPDF_Stream>(obj->value);
+            auto const& s2 = std::get<QPDF_Stream>(other.obj->value);
+            if (!s1.m->stream_dict.equivalent_to(s2.m->stream_dict, depth - 1)) {
+                return false;
+            }
+            return s1.m->stream_data->view() == s2.m->stream_data->view();
+        }
+    case ::ot_operator:
+        return std::get<QPDF_Operator>(obj->value).val ==
```
m-holger: sorry, should have spotted this before. Operators and InlineImages can only appear if we are parsing content streams. If either of them are encountered here something has seriously gone wrong. Probably should be an internal error.

libtests/objects.cc
```diff
@@ -304,14 +304,504 @@ test_2(QPDF& pdf, char const* arg2)
...
+    auto make_stream = [&pdf](std::string const& data) {
+        return pdf.makeIndirectObject(QPDFObjectHandle::newStream(&pdf, data));
+    };
+    // Scenario 1: Basic Equality: Name, Scalars, Operators, Inline Images
+    {
+        auto name1 = QPDFObjectHandle::newName("/Test");
+        auto name2 = QPDFObjectHandle::newName("/Test");
+        auto integer = QPDFObjectHandle::newInteger(42);
+        auto op1 = QPDFObjectHandle::newOperator("BT");
+        auto op2 = QPDFObjectHandle::newOperator("BT");
```
m-holger: no need to test for operators and inline iimages.

libtests/objects.cc
```diff
@@ -304,14 +304,504 @@ test_2(QPDF& pdf, char const* arg2)
     assert(!default_limits());
 }
 
+static void
+test_3(QPDF& pdf, char const* arg2)
+{
+    auto make_stream = [&pdf](std::string const& data) {
+        return pdf.makeIndirectObject(QPDFObjectHandle::newStream(&pdf, data));
```
m-holger: All streams are indirect - the call to `makeIndirectObject`  is unnecessary, which also makes `make_stream` unnecessary. Also, use `pdf.newStream(data)` in preference to `QPDFObjectHandle::newStream(&pdf, data)`

libtests/objects.cc
```diff
@@ -304,14 +304,504 @@ test_2(QPDF& pdf, char const* arg2)
     assert(!default_limits());
 }
 
+static void
+test_3(QPDF& pdf, char const* arg2)
+{
```
m-holger, line 310: add comment `// test compare_to`

libtests/objects.cc
```diff
@@ -304,14 +304,504 @@ test_2(QPDF& pdf, char const* arg2)
...
+    // Scenario 2: Numeric Types (Int vs Real)
+    {
+        auto integer = QPDFObjectHandle::newInteger(1);
+        auto real = QPDFObjectHandle::newReal("1.0");
+        assert(real.equivalent_to(integer));
+        assert(integer.equivalent_to(real));
+    }
+    // Scenario 3: Array Order Sensitivity
+    {
+        auto a1 = QPDFObjectHandle::parse("[1 2]");
```
m-holger: In test code, parsing is acceptable, however the less verbose `"[1 2]"_qpdf` would be preferred. In library code, this should be written as `Array({Integer(1), Integer(2))`. (needs `using namespace qpdf;`)

libtests/objects.cc
```diff
@@ -304,14 +304,504 @@ test_2(QPDF& pdf, char const* arg2)
...
+        assert(!a1.equivalent_to(a2));
+        assert(!a2.equivalent_to(a1));
+    }
+    // Scenario 4: Dictionary Key Order Insensitivity
+    {
+        auto d1 = QPDFObjectHandle::newDictionary();
+        d1.replaceKey("/A", QPDFObjectHandle::newInteger(1));
+        d1.replaceKey("/B", QPDFObjectHandle::newInteger(2));
+        auto d2 = QPDFObjectHandle::newDictionary();
+        d2.replaceKey("/B", QPDFObjectHandle::newInteger(2));
```
m-holger: Similarly,
```cpp
        auto d1 = "<< /A 1 /B 2 >>"_qpdf;
        auto d2 = Dictionary({{"/B", Integer(2)}, {"/A", Integer(1)}});
```

libtests/objects.cc
```diff
@@ -304,14 +304,504 @@ test_2(QPDF& pdf, char const* arg2)
...
+        d2.replaceKey("/A", QPDFObjectHandle::newInteger(1));
+        assert(d1.equivalent_to(d2));
+        assert(d2.equivalent_to(d1));
+    }
+    // Scenario 5: Direct vs Indirect Equality
+    {
+        auto obj = QPDFObjectHandle::newInteger(100);
+        auto indirect = pdf.makeIndirectObject(obj);
+        assert(obj.equivalent_to(indirect));
+        assert(indirect.equivalent_to(obj));
```
m-holger, line 346: This doesn't actually test direct vs indirect - `obj` is a reference to `indirect`

libtests/objects.cc
```diff
@@ -304,14 +304,504 @@ test_2(QPDF& pdf, char const* arg2)
...
+    {
+        auto name1 = QPDFObjectHandle::newName("/Test");
+        auto name2 = QPDFObjectHandle::newName("/Test");
+        auto integer = QPDFObjectHandle::newInteger(42);
+        auto op1 = QPDFObjectHandle::newOperator("BT");
+        auto op2 = QPDFObjectHandle::newOperator("BT");
+        auto iim1 = QPDFObjectHandle::newInlineImage("foo");
+        auto iim2 = QPDFObjectHandle::newInlineImage("foo");
+        auto iim3 = QPDFObjectHandle::newInlineImage("bar");
+        assert(name1.equivalent_to(name2));
```
m-holger: Don't create variables if you only use a value once. Use `assert(name1.equivalent_to(Name("/Test")));` or `assert(name1.equivalent_to("/Test"_qpdf));` 

(or even  `assert(Name("/Test").equivalent_to(Name("/Test")));` )

libtests/objects.cc
```diff
@@ -304,14 +304,504 @@ test_2(QPDF& pdf, char const* arg2)
...
+    }
+    // Scenario 36: Distinct Direct Null Objects
+    {
+        QPDFObjectHandle h1 = QPDFObjectHandle::newNull();
+        QPDFObjectHandle h2 = QPDFObjectHandle::newNull();
+        assert(h1.equivalent_to(h2));
+    }
+    // Scenario 37: Distinct Indirect Nulls (Different IDs)
+    {
+        auto null1 = pdf.makeIndirectObject(QPDFObjectHandle::newNull());
```
m-holger: or `auto null1 = pdf.newIndirectNull();`

libtests/objects.cc
```diff
@@ -304,14 +304,504 @@ test_2(QPDF& pdf, char const* arg2)
...
+        auto null2 = pdf.makeIndirectObject(QPDFObjectHandle::newNull());
+        assert(null1.equivalent_to(null2));
+    }
+    // Scenario 38: Broken References to Different Missing Objects
+    {
+        auto bad1 = pdf.getObjectByID(999999, 0);
```
m-holger: Preferred: `auto bad1 = pdf.getObject(999999, 0);`

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2425,3 +2425,153 @@ operator""_qpdf(char const* v, size_t len)
...
+    size_t size1 = size();
+    size_t size2 = other.size();
+    if (size1 != size2) {
+        return false;
+    }
+
+    // B. Structural comparison
+    qpdf_object_type_e t1 = resolved_type_code();
+    qpdf_object_type_e t2 = other.resolved_type_code();
+    if (t1 == ::ot_reference || t2 == ::ot_reference) {
```
m-holger: Express more simply as:
```cpp
    if (t1 == ::ot_reference) {
        return referenced_object().equivalent_to(other, depth - 1);
    }
    if (t2 == ::ot_reference) {
        return equivalent_to(other.referenced_object(), depth - 1);
    }
```
for `referenced_object` see #1673

libqpdf/QPDFObjectHandle.cc
```diff
@@ -2425,3 +2425,153 @@ operator""_qpdf(char const* v, size_t len)
...
+//
+// (1) We deviate from Annex J by comparing raw streams only, without
+// decoding.
+//
+// (2) Loop detection is expensive and is avoided. If either object has
+// a cycle in its forward orbit, this implementation will return false.
+
+bool
+BaseHandle::equivalent_to(BaseHandle const& other, int depth) const
+{
```
m-holger: Lets move the method to the other BaseHandle methods (between BaseHandle::copy and BaseHandle::unparse() feels reasonable)

m-holger (changes requested):
Some of the tests go further than necessary, eg normalizing of strings and names, which is really part of the parser tests. However, since you already have written them lets keep them.

libtests/equivalence.cc
qooxzuub: fixed (tests are now in objects.cc)

libtests/objects.cc
qooxzuub: done: `// test equivalent_to`

libtests/objects.cc
qooxzuub: fixed

libqpdf/QPDFObjectHandle.cc
qooxzuub: done

libqpdf/QPDFObjectHandle.cc
qooxzuub: done

libtests/objects.cc
qooxzuub: done

libtests/objects.cc
qooxzuub: done

libtests/objects.cc
qooxzuub: should be less of these now

libqpdf/QPDFObjectHandle.cc
qooxzuub: done

libtests/objects.cc
qooxzuub: done

libtests/objects.cc
qooxzuub: fixed

libtests/objects.cc
qooxzuub: fixed

libtests/objects.cc
qooxzuub: mostly done, but I kept some separate key insertion as this is supposedly testing that key insertion order is irrelevant to equivalent_to.

qooxzuub (commented):
Thanks for the latest review

m-holger:
@qooxzuub , could you squash the commits. Thanks

qooxzuub:
@m-holger Yes, commits squashed.

Issue #1333, opened 2025-01-20
Install latest version in Centos


ridgey-dev:
I want to install the latest version of qpdf in Centos.

When doing it via yum install it gets me the old version 5.0.1, so I tried to install it manually:

`FROM bref/php-83-fpm-dev:2.2.21

RUN yum install -y \
    gcc-c++ \
    gzip \
    make \
    libjpeg-devel \
    tar \
    wget \
    zip

COPY --from=composer:2.7.2 /usr/bin/composer /usr/bin/composer

# Install cmake
RUN wget https://cmake.org/files/v3.24/cmake-3.24.3-linux-x86_64.tar.gz && \
    tar -xvzf cmake-3.24.3-linux-x86_64.tar.gz && \
    mv cmake-3.24.3-linux-x86_64 /opt/cmake && \
    ln -s /opt/cmake/bin/cmake /usr/local/bin/cmake && \
    ln -s /opt/cmake/bin/ctest /usr/local/bin/ctest && \
    ln -s /opt/cmake/bin/cpack /usr/local/bin/cpack

# Install qpdf
RUN wget https://github.com/qpdf/qpdf/releases/download/v11.9.1/qpdf-11.9.1.tar.gz && \
    tar -xvzf qpdf-11.9.1.tar.gz && \
    cd qpdf-11.9.1 && \
    mkdir build && \
    cd build && \
    cmake .. && \
    make && \
    make install
`

Building it doesn't give any errors, but executing qpdf --version gives the following error:

`qpdf: error while loading shared libraries: libqpdf.so.29: cannot open shared object file: No such file or directory`

How can I fix this?

jberkenbilt:
From the `build` directory, you should replace your `make` and `make install` commands with
```
cmake --build .
cmake --install .
```
You should also specify a release type, e.g. 
cmake .. -DCMAKE_BUILD_TYPE=RelWithDebInfo

It would also help if you didn't have the old OS package installed. Does this help?

ridgey-dev:
Thanks for you answer.

I fixed it by setting the LD_LIBRARY_PATH:

`# Install cmake
RUN wget https://cmake.org/files/v3.24/cmake-3.24.3-linux-x86_64.tar.gz \
    && tar -xvzf cmake-3.24.3-linux-x86_64.tar.gz \
    && mv cmake-3.24.3-linux-x86_64 /opt/cmake \
    && ln -s /opt/cmake/bin/cmake /usr/local/bin/cmake \
[...]

PR #1666, opened 2026-02-06
Update installation instructions in README

SuperCowProducts:
Added information about setting LD_LIBRARY_PATH for installation.

Thank you 🙏 @ridgey-dev ! (https://github.com/qpdf/qpdf/issues/1333#issuecomment-2604112067)

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ cmake -S . -B build
 cmake --build build --config Release
 ```
 
-Installation can be done with `cmake --install`. Packages can be made with `cpack`.
+Installation can be done with `cmake --install` (you may need to set the `LD_LIBRARY_PATH` variable, with a command such as `export LD_LIBRARY_PATH=/usr/local/lib64:$LD_LIBRARY_PATH`). Packages can be made with `cpack`.
 
 The tests use `qtest`, and the test driver is invoked by `ctest`. To see the real underlying tests,
 run `ctest --verbose` so that you can see `qtest`'s output. If you need to turn off qtest's color output,
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ cmake -S . -B build
 cmake --build build --config Release
 ```
 
-Installation can be done with `cmake --install` (you may need to set the `LD_LIBRARY_PATH` variable, with a command such as `export LD_LIBRARY_PATH=/usr/local/lib64:$LD_LIBRARY_PATH`). Packages can be made with `cpack`.
+Installation can be done with `cmake --install` (you may need to set the `LD_LIBRARY_PATH` variable in your `.bashrc` or similar, with a command such as `export LD_LIBRARY_PATH=/usr/local/lib64:$LD_LIBRARY_PATH`). Packages can be made with `cpack`.
 
 The tests use `qtest`, and the test driver is invoked by `ctest`. To see the real underlying tests,
 run `ctest --verbose` so that you can see `qtest`'s output. If you need to turn off qtest's color output,


SuperCowProducts:
There may be a way to avoid putting `LD_LIBRARY_PATH` in `.bashrc` by building it in a specific way, but that goes beyond my technical expertise and honestly if it works I see no harm in doing so.

jberkenbilt:
Thanks for this. I love it when people insert those little tidbits that we forget everyone doesn't know. It drive me crazy when I see something that's missing some crucial piece of information. These tiny fixes make this a better product.

jberkenbilt:
I'm okay with taking this as is as it's better than what's there now. Technically, you should only have to do this if you are install qpdf in a location other than where your system already looks for libraries. System defaults can be found and configured in /etc/ld.so.conf and /etc/ld.so.conf.d on Linux, and there are other mechanisms on the mac (which uses DYLD_LIBRARY_PATH), but if you install it somewhere that it isn't picked up automatically, there's nothing wrong with setting LD_LIBRARY_PATH. It is usually possible to make runpath information in the binary, but it's not always a best practice on Linux. Macos has an approach using applications that gets around this whole thing (while introducing its own complications). Unless @m-holger feels differently, I'm good with merging this README change as given on the basis that it is better than what's there now and that I don't want to get distracted wordsmithing this additional information.

jberkenbilt (approved):
Good as is -- will merge unless @m-holger beats me to it or has something he'd like to see changed.

SuperCowProducts:
> Thanks for this. I love it when people insert those little tidbits that we forget everyone doesn't know. It drive me crazy when I see something that's missing some crucial piece of information. These tiny fixes make this a better product.

As someone who works on ARM machines I can recount a good amount of 'past experiences' ;)

I feel making installation easier is a useful strategy for wider FOSS adoption: silly errors stop us from using awesome software built by the community!

jberkenbilt:
> I feel making installation easier is a useful strategy for wider FOSS adoption: silly errors stop us from using awesome software built by the community!

Indeed -- I remember earlier in my career, doing some Java thing (not my favorite environment) and being told to "add something to log4j" with no hint anywhere of where that was, how to reach it, etc., and this was long before LLMs. I spent hours of reading source code to figure it the file I needed to edit was inside a jar file. One or two lines of docs would have been all that was needed. I have also been known to do pull requests that add one or two sentences to a README -- basically put the thing in that would have saved you time if it had been there. :-) :clap:

m-holger:
@SuperCowProducts , thanks for your effort.

SuperCowProducts:
> I'm okay with taking this as is as it's better than what's there now. Technically, you should only have to do this if you are install qpdf in a location other than where your system already looks for libraries. System defaults can be found and configured in /etc/ld.so.conf and /etc/ld.so.conf.d on Linux, and there are other mechanisms on the mac (which uses DYLD_LIBRARY_PATH), but if you install it somewhere that it isn't picked up automatically, there's nothing wrong with setting LD_LIBRARY_PATH. It is usually possible to make runpath information in the binary, but it's not always a best practice on Linux. Macos has an approach using applications that gets around this whole thing (while introducing its own complications). Unless @m-holger feels differently, I'm good with merging this README change as given on the basis that it is better than what's there now and that I don't want to get distracted wordsmithing this additional information.

@jberkenbilt I eventually discovered `qpdf` had been installing in `/usr/local/lib64/` instead of `/usr/lib64/` - that might have been it!

jberkenbilt:
@SuperCowProducts Thanks for the follow-up on the path. Makes sense!

PR #1667, opened 2026-02-06
Add PDF named/explicit destination helpers

qooxzuub:
- QPDFExplicitDestinationObjectHelper: wrap arrays, validate grammar, detect remote. 
- QPDFNamedDestinationObjectHelper: resolve named destinations recursively. 
- QPDFNamedDestinationDocumentHelper: unify legacy /Dests and modern /Names/Dests, iterate, map to page index. 
- Refactor QPDFOutlineDocumentHelper to use new helpers. 
- Add test PDF and destinations.cc covering legacy, modern, valid, and invalid destinations.

This consolidates named destination resolution into a new DocumentHelper class along with supporting object helpers. The motivation is to be able to move destination resolution logic from wrappers (pikepdf) into qpdf library to improve performance.

### DIFF
--- a/include/qpdf/QPDFExplicitDestinationObjectHelper.hh
+++ b/include/qpdf/QPDFExplicitDestinationObjectHelper.hh
@@ -17,19 +17,19 @@ class QPDF_DLL QPDFExplicitDestinationObjectHelper: public QPDFObjectHelper
      * is an array.
      */
     QPDF_DLL
-    QPDFObjectHandle get_explicit_array() const;
+    QPDFObjectHandle getExplicitArray() const;
 
     /**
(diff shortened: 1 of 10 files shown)


m-holger:
@qooxzuub , thanks for having a stab at this.

I haven't had time to look at this in any detail, but can you have a look QPDFFormFieldObjectHelper and FormNode. This is still work in progress, but it should give you some idea of the direction we are going. Feel free to ask any questions.

qooxzuub:
@m-holger thanks, the last commit follows QPDFFormFieldObjectHelper style more closely. Is this any nearer the mark?

PR #1677, opened 2026-03-21
Fix bug in `QPDFObjectHandle::getNumericValue` (fixes #1676)

m-holger:
When run using a locale that does not use "."  as the decimal point (e.g. one that uses "," instead) real number were incorrectly parsed.

### DIFF
--- a/libqpdf/CMakeLists.txt
+++ b/libqpdf/CMakeLists.txt
@@ -107,6 +107,7 @@ include(FindPkgConfig)
 include(CheckTypeSize)
 include(CheckIncludeFile)
 include(CheckCSourceCompiles)
+include(CheckCXXSourceCompiles)
 include(CheckCSourceRuns)
 include(CheckSymbolExists)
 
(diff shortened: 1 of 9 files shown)


m-holger:
`atol` and `std::from_chars` behave slightly differently when converting strings to doubles:

1. `atol` accepts leading white-space while `from_chars` doesn't.
2. `atol` accepts reals with exponent part, while `from_chars` with the selected options doesn't.
3. `from_chars` allows us to detect trailing characters.

None of these are of any concern for reals create by qpdf, but could be an issue for user created reals using `newReal`. I believe that the intended API for `newReal` is that the supplies string should conform with rules for reals defined in the PDF spec, but this has not been clearly stated or enforced, and in the case of leading or trailing white-space has made no practical difference so far. We therefore have to consider that enforcing the intended API now might break existing code.

My proposal would be to replicate the existing behaviour for now, but add warnings if `newReal` is called with a string that is not a valid PDF real. At the same time, we should clarify the API and from state that from qpdf 13 any deviation will be considered an error.

In practical terms, we should check for and remove any leading white-space in `newReal`, as well detecting any empty strings. Trailing characters should be dealt with in `getNumericValue`.

jberkenbilt:
@m-holger I think we need accept anything that could reasonably parsed as a real number if we accept it today because people may be calling newReal on the output of some string formatting operation. I know there are files in the wild that use scientific notation (even though invalid), and it would surprise me if there weren't files that use `,` as the decimal point because I think people probably just use sprintf or similar. We could warn, but I'd lean against since those warnings are visible to the user. But I could see a case for warning in hopes that the developer sees the warnings. I agree that we could safely make it an error if newReal is called on numbers that, after stripping spaces, are not valid, so `1.5e7` is valid, but `12eggs` is not valid. However we should make sure we output a correct PDF real even if the input string is invalid. I know we've churned on this, and I can't recall the current state. If you disagree with any of this, let me know your reasoning.

I defer to you on where the various checks should live. Let me know if I've misunderstood your concerns.

Issue #514, opened 2021-03-19
Assertion failure triggerable after attempting to add duplicate foreign pages


jbarlow83:
I came across this assertion failure while trying to get QPDF to accept adding the same foreign page twice to a file. (Suppose the page is a separator we're adding between sections of a document.)

I managed to trigger the following assertion failure:

```
Assertion failed: (this->m->pageobj_to_pages_pos.size() == QIntC::to_size(npages)), function insertPage, file libqpdf/QPDF_pages.cc, line 264.
```

In short it seems to me that after a failure to add the same foreign page twice, some internal state is left inconsistent.

The following pikepdf code will reproduce the issue:

```python
import pikepdf

g = pikepdf.open('tests/resources/graph.pdf')
p = pikepdf.new()

p.pages.append(g.pages[0])

try:
    p.pages.append(g.pages[0])  # Raises an exception - duplicate page reference
except pikepdf.PdfError:
    print("Hit the expected exception")

foreign = p.copy_foreign(g.pages[0])
print("We're going to SIGABRT now...")
p.pages.append(foreign)

```

Or roughly equivalent C++:

```c++
#include <iostream>

#include <qpdf/QPDF.hh>
#include <qpdf/QPDFExc.hh>
#include <qpdf/QPDFSystemError.hh>
#include <qpdf/QPDFObjGen.hh>
#include <qpdf/QPDFXRefEntry.hh>
#include <qpdf/Buffer.hh>
#include <qpdf/BufferInputSource.hh>
#include <qpdf/QPDFWriter.hh>
#include <qpdf/QPDFPageDocumentHelper.hh>
#include <qpdf/Pl_Discard.hh>


int main (int argc, char **argv) {
    QPDF src;
    src.processFile(argv[1]);

    QPDF dst;
    dst.emptyPDF();

    auto page0 = src.getAllPages().at(0);

    dst.addPage(page0, false);

    try {
        dst.addPage(page0, false); 
    } catch (const QPDFExc& e) {
        std::cout << "Expected exception:" << std::endl;
        std::cout << e.what() << std::endl;
    }

    std::cout << "Make foreign:" << std::endl;
    auto foreign = dst.copyForeignObject(page0);
    auto foreign_copy = foreign.shallowCopy();
    std::cout << "Assertion failure happens here:" << std::endl;
    dst.addPage(foreign_copy, false);
    return 0;
}

```

[...]

PR #1703, opened 2026-04-23
Clear Copier scratch state on unwind in copyForeignObject

### LINKED ISSUE #1700: `copyForeignObject` leaves `Copier::visiting` populated when an inner copy throws, so the next call asserts
`Objects::Foreign::Copier::visiting` is the cycle-detection set used during the foreign-object walk. `Copier::copied` asserts it's empty on entry and `reserve_objects` (`libqpdf/QPDF.cc:515-560`) relies on the matching `visiting.erase(foreign_og)` at line 559 to clean up on the way out. If anything inside the walk throws (the easy way: `QPDFExc("Too many warnings - file is too badly damaged")` from a damaged dict or stream length resolution), that erase never runs and `visiting` stays dirty.

Next call to `copyForeignObject` on the same destination `QPDF` hits:

```cpp
// libqpdf/QPDF.cc:482
util::assertion(
    visiting.empty(), "obj_copier.visiting is not empty at the beginning of copyForeignObject");
```

`util::assertion` throws `std::logic_error` (`libqpdf/qpdf/Util.hh:22-31`). `logic_error` isn't a subclass of `runtime_error`, so the documented catch-and-continue pattern

```cpp
try { dest.addPage(p); } catch (QPDFExc&) { continue; }
```

walks straight into `std::terminate`. The three `copyForeignObject` call sites in `QPDFJob.cc` (lines 1889, 2083, 2324) loop in that style internally, and external consumers like pikepdf do the same to skip damaged pages.

This is in the same family as #514 (assertion-fired-after-foreign-copy-failure), but a different state-leak path. Not a duplicate.

### Repro
[2026-04-23-copyForeignObject-visiting.zip](https://github.com/user-attachments/files/27007543/2026-04-23-copyForeignObject-visiting.zip)
Standalone, public-API only. No fuzzer harness. Build qpdf with sanitizers:

```
CC=clang CXX=clang++ CFLAGS=-fsanitize=address,undefined CXXFLAGS=-fsanitize=address,undefined \
  cmake -S . -B build_asan -DBUILD_SHARED_LIBS=OFF -DCMAKE_BUILD_TYPE=Debug
cmake --build build_asan -j --target qpdf
```

Compile the reproducer:

```
  -I include 2026-04-23-copyForeignObject-visiting-repro.cc \
  -o repro
```

Run on the attached 1510-byte PoC:

```
$ ./repro 2026-04-23-copyForeignObject-visiting-poc.bin
bufA=1009 bytes, bufB=497 bytes
B has 1 pages
--- addPage call 1 ---
  QPDFExc: A (offset 944): Too many warnings - file is too badly damaged
--- addPage call 2 ---
  *** logic_error escaped public API: obj_copier.visiting is not empty at the beginning of copyForeignObject
```

First `addPage` throws `QPDFExc`, exactly as advertised. Second `addPage` on the same destination throws `logic_error`, which the documented catch-set doesn't cover.

### Fix idea

A scope guard in `Copier::copied` so `visiting` always gets cleared:

```cpp
struct VisitingGuard { VisitingNNTree& v; ~VisitingGuard() { v.clear(); } };
VisitingGuard _g{visiting};
```

Probably want the same treatment for `to_copy` so the copier is fully reusable after a partial call, but that's a separate call.

Happy to send a PR with a libtest that calls `copyForeignObject` twice on the attached PDF. Let me know if you'd prefer the minimal guard or a broader pass over the Copier's exception-safety.

### Scope

Not a memory-safety bug. ASan/UBSan ran for ~30M sanitizer-instrumented executions of the surrounding fuzzer with no heap/stack/UAF reports. The reproducer touches only `QPDF::copyForeignObject` and `QPDFPageDocumentHelper::addPage`; AcroForm helpers are not on the path.

MarkLee131:
Close #1700:
Copier::copied asserts visiting is empty on entry and relies on reserve_objects to leave it empty on normal exit. If any inner step throws (QPDFExc "Too many warnings - file is too badly damaged" during a damaged foreign dictionary walk is the easy repro), the matching visiting.erase calls are skipped and the set stays dirty. The next copyForeignObject on the same destination QPDF hits util::assertion(visiting.empty(), ...) and throws std::logic_error, which is outside the documented throw set (QPDFExc / runtime_error).

Consumers that loop copyForeignObject / addPage and catch QPDFExc between iterations (pikepdf, PDFArranger, and QPDFJob.cc:1889/2083/2324 internally) therefore terminate via std::terminate on a second call.

Fix with a small RAII guard that clears visiting and to_copy on unwind so the Copier is always reusable after a partial call. The assertion is left in place as documentation.

Same family of bug as #514; different state-leak path.

### DIFF
--- a/libqpdf/QPDF.cc
+++ b/libqpdf/QPDF.cc
@@ -479,6 +479,23 @@ Objects::Foreign::Copier::copied(QPDFObjectHandle const& foreign)
     // Note that we explicitly allow use of copyForeignObject on page objects. It is a documented
     // use case to copy pages this way if the intention is to not update the pages tree.
 
+    // Ensure per-call scratch state (visiting, to_copy) is cleared on unwind so that an exception
+    // thrown inside reserve_objects or replace_indirect_object cannot leave the Copier in a state
+    // that trips the assertion below on the next call. Without this guard a first
+    // copyForeignObject that fails mid-walk (for example with QPDFExc "Too many warnings") makes
+    // every subsequent copyForeignObject on the same destination throw std::logic_error, which
+    // is outside the documented throw set and terminates the caller.
+    struct ScopeClear
+    {
+        QPDFObjGen::set& v;
+        std::vector<QPDFObjectHandle>& tc;
+        ~ScopeClear()
+        {
+            v.clear();
+            tc.clear();
+        }
+    } _guard{visiting, to_copy};
+
     util::assertion(
         visiting.empty(), "obj_copier.visiting is not empty at the beginning of copyForeignObject");
 


m-holger:
@MarkLee131 , the existing behavior is the intended behavior. A previous exception has left the QPDF object in an invalid state and continued use of the object is unsafe. Simply clearing the cache simply hides the problem rather than fixing it.

MarkLee131:
Thanks @m-holger. Two clarifications, in case there's a misunderstanding about what the patch clears and which QPDF is involved: 
1. The guard only clears the Copier's per-call scratch state (`visiting`, `to_copy`). `object_map`, which holds the partial reservations from the failed call, is untouched, so no object-level state is hidden.
2. In the reproducer, the *source* is the damaged QPDF. The *destination* (which owns `Copier::visiting`) was loaded successfully and isn't mutated in any user-visible way by the failed call. The leak is in the destination's scratch state, not its document state.
3. qpdf itself relies on the catch-and-continue pattern: `QPDFJob.cc:1889`, `:2083`, `:2324` all call `copyForeignObject` repeatedly on the same destination inside  `catch (QPDFExc&) { continue; }`. If continued use after a `QPDFExc` is unsafe, those sites are unsafe today.
4.  Even if the intended contract is "destination is unusable after a failed foreign copy", signalling that with `std::logic_error` (a programmer-error type, not a`runtime_error`) bypasses every documented `catch (QPDFExc&)` / `catch (std::runtime_error&)` site. A `QPDFExc` would at least be catchable.

Happy to close the PR if the destination really must be discarded after any failed `copyForeignObject`, but in that case the `QPDFJob.cc` call sites probably need updating too. Otherwise I'd appreciate a second look at the patch as a scratch-state-only fix.

m-holger:
I am still looking into the issues, but at the moment my working assumption is that `setMaxWarnings` needs a fix. 

> Even if the intended contract is "destination is unusable after a failed foreign copy", signalling that with std::logic_error (a programmer-error type, not a runtime_error) bypasses every documented catch (QPDFExc&) / catch (std::runtime_error&) site. A QPDFExc would at least be catchable.

A logic error is derived from `std::exception` and is catchable. It is correct to throw a logic error because we encountered a bug and therefore all current qpdf objects (including QPDF objects, object handles and helpers) are suspect and should be discarded. I would expect all but the simplest applications to catch all exceptions and either terminate gracefully or abort the current operation and move on to the next operation with a fresh set of input files.

When you say "documented catch", is there anything in the qpdf documentation that suggests we do not throw logic errors (or other exceptions not derived from runtime_error)? If so, it needs to be fixed.

PR #1704, opened 2026-04-23
Throw QPDFExc instead of logic_error for foreign handle in writer

### LINKED ISSUE #1701: `fixCopiedAnnotations` returns success but leaves the destination page with foreign references; next `write()` aborts
Same error you saw in #548 ("Fail to merge pdfs with acroform"), recurring through a different code path in the same helper. The 10.4 fix at `4a648b9a` covered `/DR` resource indirection. The annotation walk in `fixCopiedAnnotations` still produces output where at least one entry under `/Annots` is a `QPDFObjectHandle` owned by the source `QPDF`. The helper returns cleanly (no exception, no warning that anything is wrong with the result), and then `QPDFWriter::write()` hits:

```cpp
// QPDFWriter.cc:1400 (reached via QPDF_Dictionary.cc:202 / QPDF_Array.cc:34)
throw std::logic_error(
    "QPDFObjectHandle from different QPDF found while writing.  "
    "Use QPDF::copyForeignObject to add objects from another file.");
```

Same `logic_error`-escapes-the-public-API issue as the other two reports today: `logic_error` is outside qpdf's documented throw-set, so consumers catching `QPDFExc` / `std::runtime_error` get `std::terminate`.

The cross-QPDF use case is documented as supported in `include/qpdf/QPDFAcroFormDocumentHelper.hh:211-225` ("allowing the from page to be in a different QPDF"), so a write right after `fixCopiedAnnotations` is the intended pattern, not misuse.

### Repro

[2026-04-23-fixCopiedAnnotations.zip](https://github.com/user-attachments/files/27007616/2026-04-23-fixCopiedAnnotations.zip)

Same build as in #1688. Compile:

```
  -I include 2026-04-23-fixCopiedAnnotations-repro.cc \
  -o repro
```

Run:

```
$ ./repro 2026-04-23-fixCopiedAnnotations-poc.bin
[... lots of "operation for dictionary attempted on object of type null/integer" warnings ...]
--- fixCopiedAnnotations ---
  no throw
--- QPDFWriter::write ---
  *** logic_error escaped public API: QPDFObjectHandle from different QPDF found while writing.  Use QPDF::copyForeignObject to add objects from another file.
```

`fixCopiedAnnotations` returns silently. The crash is on the next operation a caller would normally do.

### Fix idea

Two directions, no strong opinion on which is better here:

1. Make `fixCopiedAnnotations` itself catch the discovery of an un-copied foreign handle in its output and throw `QPDFExc` (or warn + drop the entry). That puts the failure at the site of the bug rather than in the writer, and keeps it inside the documented throw-set.
2. Fix the underlying walk so every annotation reference does get rewritten. Probably the right long-term fix: the walk has early-out paths for non-standard `/Annots` entries (null, integer, dangling reference) where `copyForeignObject` is skipped but the original foreign handle ends up retained in the output. Either still produce a local handle or drop the entry.

Related history in this helper:

- #548 (closed by `4a648b9a` in 10.4): same error message, fix only covered the `/DR` resource path.
- #1620 (closed by #1621): UAF via deque iterator invalidation. Different bug class but same "what if the source annotation walk is malformed?" pattern. Worth a once-over.

Happy to send a PR if you tell me which direction you want.

MarkLee131:
impl::Writer::enqueue used util::assertion to check that every indirect handle pulled into the write queue was owned by the destination QPDF. util::assertion throws std::logic_error, which sits outside the documented throw set (QPDFExc / std::runtime_error) and kills consumers that catch QPDFExc.

This is reachable from at least one helper path: QPDFAcroFormDocument Helper::fixCopiedAnnotations on damaged input can return success but leave the destination page with foreign indirect handles nested inside annotation sub-dicts. QPDFWriter::write then walks the graph, hits a foreign handle, and calls std::terminate through the consumer's QPDFExc-only catch.

The underlying helper bug is real and deserves its own fix, but for the writer's side it is appropriate to surface the condition as the documented exception type so callers can recover. Turn the assertion into an explicit QPDFExc throw.

Fixes #1701.

### DIFF
--- a/libqpdf/QPDFWriter.cc
+++ b/libqpdf/QPDFWriter.cc
@@ -1389,16 +1389,27 @@ void
 impl::Writer::enqueue(QPDFObjectHandle const& object)
 {
     if (object.indirect()) {
-        util::assertion(
-            // This owner check can only be done for indirect objects. It is possible for a direct
-            // object to have an owning QPDF that is from another file if a direct QPDFObjectHandle
-            // from one file was insert into another file without copying. Doing that is safe even
-            // if the original QPDF gets destroyed, which just disconnects the QPDFObjectHandle from
-            // its owner.
-            object.qpdf() == &qpdf,
-            "QPDFObjectHandle from different QPDF found while writing.  "
-            "Use QPDF::copyForeignObject to add objects from another file." //
-        );
+        // This owner check can only be done for indirect objects. It is possible for a direct
+        // object to have an owning QPDF that is from another file if a direct QPDFObjectHandle
+        // from one file was insert into another file without copying. Doing that is safe even
+        // if the original QPDF gets destroyed, which just disconnects the QPDFObjectHandle from
+        // its owner.
+        //
+        // This used to be a util::assertion which throws std::logic_error. Damaged input
+        // combined with helpers like QPDFAcroFormDocumentHelper::fixCopiedAnnotations can leave
+        // stray foreign indirect handles in the graph; since logic_error is outside the
+        // documented throw set (QPDFExc / std::runtime_error), consumers that catch QPDFExc
+        // would crash via std::terminate. Surface the condition as a QPDFExc instead so the
+        // catch-and-continue pattern used by tools like pikepdf / PDFArranger works.
+        if (object.qpdf() != &qpdf) {
+            throw QPDFExc(
+                qpdf_e_internal,
+                qpdf.getFilename(),
+                "",
+                0,
+                "QPDFObjectHandle from different QPDF found while writing.  "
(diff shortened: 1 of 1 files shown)


m-holger:
Thank you for your PR.

> ... it is appropriate to surface the condition as the documented exception type so callers can recover. 

qpdf throws std::logic_error when it detects that an invariant has been violated in line with common C++ practice, indicating a programming error. In this case it is not flaggesd as an internal error as there is the possibility that is due to a user programming error, although in this case it looks like it is due to a qpdf bug.    `qpdf_e_internal` is generally only used when C++ exception are translated into error codes for qpdf's C API.

PR #1705, opened 2026-04-23
Skip out-of-range object ids in Objects::compressible

### LINKED ISSUE #1702: `QPDFWriter::write` throws uncaught `logic_error` from `Objects::compressible()` after AcroForm helpers mutate the document
`Objects::compressible()` (called from the writer's object-stream-generation path) caches `max_obj = qpdf.getObjectCount()` once at the top, then walks the document graph. Every encountered object id is bounds-checked against that cached value:

```cpp
// libqpdf/QPDF_objects.cc:2057-2078
const size_t max_obj = qpdf.getObjectCount();
...
while (!queue.empty()) {
    auto obj = queue.back();
    queue.pop_back();
    if (obj.getObjectID() > 0) {
        QPDFObjGen og = obj.getObjGen();
        const size_t id = toS(og.getObj() - 1);
        if (id >= max_obj) {
            throw std::logic_error(
                "unexpected object id encountered in getCompressibleObjGens");
        }
        ...
```

On the attached PoC, the walk encounters at least one object whose id is out of range relative to `getObjectCount()`. I didn't fully nail down which case my input hits. The two plausible candidates:

- An AcroForm helper allocated new objects (via `newIndirect` or `copyForeignObject`) whose ids exceed the original parsed-xref `max_obj`, and `getObjectCount()` is returning the stale parsed value rather than the live max.
- The damaged source has an `/Annots` or `/Parent` chain pointing at an object id outside the xref, and that reference survives through the helper calls and into the walk.

Either way, this is an internal sanity check that's reachable from a documented public-API sequence (parse → AcroForm helper → write). `std::logic_error` is outside the throw-set the writer is documented to use, so a consumer catching `QPDFExc` / `std::runtime_error` ends up at `std::terminate`.

This is the same general pattern as the other two reports I'm filing today (the `Copier::visiting` leak and the `fixCopiedAnnotations` foreign-ref leftover), just at a different site.

### Repro

[2026-04-23-getCompressibleObjGens.zip](https://github.com/user-attachments/files/27007674/2026-04-23-getCompressibleObjGens.zip)

```
  -I include 2026-04-23-getCompressibleObjGens-repro.cc \
  -o repro
```

```
$ ./repro 2026-04-23-getCompressibleObjGens-poc.bin
WARNING: object 41 0: operation for dictionary attempted on object of type null: ignoring key replacement request
WARNING: object 46 0: operation for dictionary attempted on object of type integer: returning null for attempted key retrieval
WARNING: object 46 0: operation for dictionary attempted on object of type integer: ignoring key replacement request
--- QPDFWriter::write ---
  *** logic_error escaped public API: unexpected object id encountered in getCompressibleObjGens
```

The reproducer calls only public helpers (`QPDFAcroFormDocumentHelper::generateAppearancesIfNeeded`, `disableDigitalSignatures`, `fixCopiedAnnotations`, `transformAnnotations`) and then `QPDFWriter::write`.

### Fix idea

Two reasonable directions:

1. Re-read `getObjectCount()` each iteration. Cheap, trivially correct if the reachability is via stale cached `max_obj`.
2. Downgrade the `throw` to a `continue` (warn-and-skip). This matches the position you took in #514: *"assertions belong in test code. In this case, the assertion failure is simply a side effect of the already-reported data loss, but it doesn't necessarily result in data loss."* Same applies here. The writer should just skip the bogus object and finish the file.

MarkLee131:
compressible() caches max_obj = qpdf.getObjectCount() and bounds-checks every object id during the graph walk. Any id outside that range triggered a std::logic_error that escaped QPDFWriter::write, outside the documented throw set.

Dangling xref references and helper-mutated graphs both produce this case on malformed input. Consistent with the position in #514 that "assertions belong in test code", skip the object instead of throwing. The existing warning path already reports the underlying data loss.

Fixes #1702.

### DIFF
--- a/libqpdf/QPDF_objects.cc
+++ b/libqpdf/QPDF_objects.cc
@@ -2073,8 +2073,10 @@ Objects::compressible()
             QPDFObjGen og = obj.getObjGen();
             const size_t id = toS(og.getObj() - 1);
             if (id >= max_obj) {
-                throw std::logic_error(
-                    "unexpected object id encountered in getCompressibleObjGens");
+                // Object id is outside the xref-derived max. Can occur on damaged input with
+                // dangling references or after helpers mutate the graph. Skip rather than throwing
+                // an internal logic_error that escapes QPDFWriter::write.
+                continue;
             }
             if (visited[id]) {
                 continue;


m-holger:
@MarkLee131 , thank you for the PR. However, your fix is not correct. Either the object is invalid, in which case it should not have entered the object table, or it is valid, in which case the calculation of max_obj is incorrect. In either case, the underlying bug needs to be fixed.

qpdf's approach to assertions can be found [here](https://github.com/qpdf/qpdf/blob/main/README-developer.md#coding-rules).

MarkLee131:
Hi @m-holger, Apologies for jumping to the patch too fast.
Looking again, getObjectCount() already calls fixDanglingReferences(), so my "stale max_obj" hypothesis in #1702 doesn't hold. Two candidates seem more plausible to me now: 
1. The walk reaches an indirect reference whose target was never added to obj_cache, in which case the upstream resolution should have produced a null, and the bug is there.                                                          
2. fixDanglingReferences short-circuits via m->fixed_dangling_refs after the first call, so any new dangling reference introduced by AcroForm helpers between the first fixDanglingReferences() and the writer's call is invisible to the bounds check. 

Would you like me to instrument compressible() against the PoC to identify which og triggers and how it entered the walk, then come back with either a targeted fix or just a closed PR if the right answer turns out to live elsewhere? 
Happy to close this PR in the meantime if you'd prefer to track the investigation on the issue.

PR #1718, opened 2026-05-08
Normalize page rotation to [0, 360) in XObject placement

### LINKED ISSUE #717: Page rotation of -90 incorrectly used by `add_overlay`
I believe the two output files generated by this script should look identical but they do not: the rotation of -90 is not correctly applied for out.pdf, but the rotation of 270 is correctly applied in out2.pdf.

```
import pikepdf

rect = pikepdf.Rectangle(pikepdf.Array([0,0,500,500]))
src_pdf = pikepdf.open("/tmp/in.pdf")
src_pdf.pages[0].Rotate = -90

out_pdf = pikepdf.new()
out_pdf.add_blank_page()
out_pdf.pages[0].add_overlay(src_pdf.pages[0], rect)
out_pdf.save("/tmp/out.pdf")

src_pdf.pages[0].Rotate = 270
out_pdf = pikepdf.new()
out_pdf.add_blank_page()
out_pdf.pages[0].add_overlay(src_pdf.pages[0], rect)
out_pdf.save("/tmp/out2.pdf")
```

This is pikepdf 10.5.1, libqpdf 12.3.2.

qooxzuub:
The PDF spec asks for Rotation to be a multiple of 90. `QPDFPageObjectHelper.cc` checks for 90, 180, 270 only. This PR normalizes page rotation values to [0,360) so that, for example, pages with /Rotate = -90 are processed correctly. 

Fixes https://github.com/pikepdf/pikepdf/issues/717

### DIFF
--- a/libqpdf/QPDFPageObjectHelper.cc
+++ b/libqpdf/QPDFPageObjectHelper.cc
@@ -683,6 +683,10 @@ QPDFPageObjectHelper::getMatrixForTransformations(bool invert)
         }
 
         // Ignore invalid rotation angle
+        rotate %= 360;
+        if (rotate < 0) {
+            rotate += 360;
+        }
         switch (rotate) {
         case 90:
             matrix = QPDFObjectHandle::Matrix(0, -scale, scale, 0, 0, width * scale);
@@ -864,6 +868,10 @@ QPDFPageObjectHelper::flattenRotation(QPDFAcroFormDocumentHelper* afdh)
     int rotate = 0;
     if (rotate_oh.isInteger()) {
         rotate = rotate_oh.getIntValueAsInt();
+        rotate %= 360;
+        if (rotate < 0) {
+            rotate += 360;
+        }
     }
     if (!((rotate == 90) || (rotate == 180) || (rotate == 270))) {
         return;


libqpdf/QPDFPageObjectHelper.cc
```diff
@@ -683,6 +683,7 @@ QPDFPageObjectHelper::getMatrixForTransformations(bool invert)
         }
 
         // Ignore invalid rotation angle
+        rotate = ((rotate % 360) + 360) % 360;
```
m-holger: @qooxzuub , I would prefer `rotate  = rotate % 360 >= 0 ?  rotate % 360 : 360 + rotate % 360`. I think this would be clearer for eg Python speakers  who may wrongly assume that all languages define `%` sensibly.

qooxzuub: Hmm what about rotate=-360 though, we'd get 360 as the result. Which doesn't affect the logic I guess, but seems odd (and not the python result)?

How about 
```
rotate %= 360;
if (rotate < 0) {
    rotate += 360;
}
```

m-holger: > what about rotate=-360 though, we'd get 360 as the result.

I think that works correctly in my (edited) code. However, I prefer your suggestion anyway.

m-holger:
@jbarlow83  , this PR is going to close https://github.com/pikepdf/pikepdf/issues/717 when merged. Are you ok with that?

m-holger:
@qooxzuub , thanks for the PR.

Doing a quick check, there may be other places where /Rotate not in [0, 360) is not handled correctly. Do you want to take this on or stick with the current scope?

jbarlow83:
@m-holger All clear

qooxzuub:
> @qooxzuub , thanks for the PR.
> 
> Doing a quick check, there may be other places where /Rotate not in [0, 360) is not handled correctly. Do you want to take this on or stick with the current scope?

I did have a very quick look earlier and it seems simpler to stick with current scope.

PR #1731, opened 2026-06-08
Prevent construction of direct object cycles (fixes #1730)

### LINKED ISSUE #1730: Direct (non-indirect) object cycles cause stack overflow in unparse(), writeJSON(), and disconnect()
### Summary

A cycle built entirely from **direct** objects (no indirect references) sends `QPDFObjectHandle::unparse()`, `writeJSON()`, and `disconnect()` into unbounded recursion and crashes the process with a stack overflow. The cycle can only be constructed in memory via the libqpdf (it cannot arise from parsing a PDF) but any code using the public API, including bindings such as pikepdf, can build it.

### Reproducer

```cpp
#include <qpdf/QPDFObjectHandle.hh>

int main() {
    auto a = QPDFObjectHandle::newDictionary();   // direct, ObjGen (0,0)
    auto b = QPDFObjectHandle::newDictionary();   // direct, ObjGen (0,0)
    a.replaceKey("/Kids", QPDFObjectHandle::newArray({b}));  // a -> b (shared)
    b.replaceKey("/Kids", QPDFObjectHandle::newArray({a}));  // b -> a (shared)
    a.unparse();   // stack overflow (also: a.writeJSON(2, &pl), a.disconnect())
}
```

Confirmed: SIGSEGV (exit 139) on `unparse()` and `writeJSON()` against libqpdf 12.x.

### Why it happens

A `QPDFObjectHandle` is a thin handle over a `shared_ptr<QPDFObject>`. The container mutators (`replaceKey`, `appendItem`/`Array::push_back`, `insertItem`, `setItem`) store the supplied handle **by value, sharing the same underlying object** — the only gate is `checkOwnership()`, which merely rejects objects owned by a *different* `QPDF`. So `b.replaceKey("/Kids", newArray({a}))` makes `b` reachable from `a` and vice-versa through shared underlying objects, closing the loop.

`unparse()`/`writeJSON()`/`disconnect()` recurse into children and short-circuit only on **indirect** references (`N 0 R`). A direct cycle never hits that guard, so recursion never terminates.

### Why it's API-only and not a parser bug

The PDF file format can only point backwards through indirect references (`obj` / `R`); every object read from a file is indirect, and the indirect short-circuit breaks any file-borne cycle. A purely-direct cycle is therefore **impossible to load from a PDF** and can only be assembled programmatically in memory. This is likely why it has gone unnoticed: it's a libqpdf API-robustness gap, not a malformed-input issue. It is reachable from C++ and from any binding (the pikepdf report that surfaced this is https://github.com/pikepdf/pikepdf/issues/731; `obj['/Kids'] = [other]` lowers 1:1 to `replaceKey("/Kids", newArray({...}))`).

### Affected APIs

`unparse()`, `unparseResolved()`, `writeJSON()`, `disconnect()` — and by extension `QPDFWriter` and JSON output, which rely on these. (`QPDFWriter`'s normal object-graph walk terminates on real files precisely because file objects are indirect.)

### Possible fix

Reject the cycle-closing insert at construction time. Every cycle must be closed by some mutator call adding the back-edge, and all mutators already funnel through `checkOwnership()`; adding a cycle check there catches the closing edge before it is ever stored. The check can be skipped entirely for leaf values and indirect references (so objects loaded from a file, which are all indirect, pay nothing), keeping the cost proportional to the size of any hand-built *direct* subgraph being inserted.

jbarlow83:
A cycle built from direct (non-indirect) objects cannot occur in a PDF file, since every backward reference in a file is indirect, but could be assembled in memory through the API. unparse(), writeJSON(), and disconnect() then recurse without bound and crash with a stack overflow.

The container mutators (replaceKey, appendItem, insertItem, setArrayItem, setArrayFromVector) now call BaseHandle::checkForCycle alongside the existing checkOwnership check and throw std::logic_error if the change would close a direct cycle. The scan stops at indirect objects, so it costs nothing for objects read from a file, and gating every mutator is sufficient because any cycle must be closed by some mutator call. Reference cycles built from indirect objects are unaffected.

### DIFF
--- a/include/qpdf/ObjectHandle.hh
+++ b/include/qpdf/ObjectHandle.hh
@@ -137,6 +137,8 @@ namespace qpdf
         inline void assign(qpdf_object_type_e required, BaseHandle&& other);
         inline void nullify();
 
+        void check_insertion(QPDFObjectHandle const& item) const;
+
         std::string description() const;
         inline QPDFObjectHandle const& get(std::string const& key) const;
 
--- a/libqpdf/QPDFObjectHandle.cc
+++ b/libqpdf/QPDFObjectHandle.cc
@@ -21,6 +21,8 @@
 #include <cstdlib>
 #include <cstring>
 #include <stdexcept>
+#include <unordered_set>
+#include <vector>
 
 using namespace std::literals;
 using namespace qpdf;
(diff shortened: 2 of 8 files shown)


m-holger:
@jbarlow83 , thanks for your PR.

I think the proper fix for #1730 will be to limit the recursion depth in the affected methods. That said, if a programmer makes an error, it seems only polite to throw an exception close to where the error is located rather than far away in a completely unrelated part of qpdf and for that reason we should also make the changes you are suggesting.

Unfortunately the new method is quite expensive. As you noted, this makes very little difference in most circumstances and in most of my tests was hardly noticeable, but where container mutators are heavily used the impact is significant. Using the `--json-input` option I see a 6% performance hit. This is mainly due to the use of sets to limit the search. By limiting the search depth instead we can bring this down to ~0.5% , and with further tweaks to ~0.25%.  The tweaks are:

- only push containers onto the stack
- as we are only interested in direct objects, use the simpler `raw_type_code`
- combine the method with `checkOwnership`  

I have added my suggested tweaks together with a couple of stylistic changes in #1732. Can you review and if you agree with my suggestions squash the commits.

m-holger:
I considered the problem of exponential explosion when suggesting the removal of seen but given that this is API only rather than applying to untrusted input files couldn't really see how this would come about without deliberate programmer action.

Looking at performance, with the restriction to containers only the cost goes down to ~1.75%, and for very simple files can actually be beneficial. Counter-intuitively (but logically) using (ordered) sets brings this down to ~0.75% while a vector based solution (see #1732) reduces it to ~0.1%, which definitely makes it worthwhile.

PR #1732, opened 2026-06-09
Suggested changes to #1731

m-holger:


jbarlow83:
I folded these changes into an update to #1731 as explained in comments there.

PR #1744, opened 2026-06-22
Make equivalent_to public and fix null stream_data handling

qooxzuub:
- Expose BaseHandle::equivalent_to() as a public QPDFObjectHandle API to allow structural equivalence checks from user code. This is intended for use by external bindings (pikepdf).

- Fix a potential null dereference when comparing stream objects, with a test. Stream comparison previously assumed stream_data was always present. When stream_data is not materialized, we now fall back to comparing raw stream data via getRawStreamData(). This was previously reviewed as part of https://github.com/qpdf/qpdf/pull/1663


qooxzuub:
@m-holger please could you review this when you get a moment? Thanks

PR #1755, opened 2026-08-15
Json leading zero

jberkenbilt:
Simple fix to leading zero. The PR at the time of creation contains unrelated commits that will merge with another PR. Only the latest commit is relevant to the fix.


jberkenbilt:
The test failure here is not real. This commit is also included in a subsequent PR, so this will merge with that.

PR #1762, opened 2026-08-18
Make QPDF::setLinearizationProgressCallback a private API method

m-holger:


jberkenbilt:
@m-holger Thanks -- I was going to try to figure out the idiom for doing this with the new code organization. This is exactly what I was looking for. The old implementation was generated by Claude. With this change, we don't need to bump to 12.5.0, so I will drop the version back to 12.4.1.

PR #1763, opened 2026-08-19
Fix possible integer overflow in #1759

m-holger:
Fixes oss_fuzz issue 547975127


jberkenbilt:
@m-holger thanks for jumping on this. Seeing the fix you chose helps acquaint me with all the stuff you've added, like `fits<T>`, which is pretty nice.

PR #1766, opened 2026-08-23
Reject suffixes in bounded numeric options

arhxam:
Bounded numeric command-line options that use the shared `to_int` and `to_uint32` validators check the value returned by `std::stoi` or `std::stoll`, but not how much of the argument was consumed. As a result, values such as `--compression-level=9junk` and `--parser-max-errors=2junk` are accepted as `9` and `2`.

This checks the conversion position in both bounded numeric helpers and routes partially parsed values through the existing usage-error path. Valid values, range validation, and diagnostic text are unchanged.

The regression covers both helper paths with signed-integer and `uint32_t` options, and the release notes record the user-visible validation change.

Tests:

- `TESTS=arg-parsing ctest --test-dir cmake-build-maintainer -R '^qpdf$' --output-on-failure` (37/37)
- full qpdf qtest suite (2,890 tests: 2,886 passed, 4 expected failures)
- remaining maintainer ctest suites (6/6 passed, including 1,439/1,439 fuzz tests)
- maintainer warning-as-error build
- `clang-format -i libqpdf/QPDFJob_config.cc`
- `git diff --check`

### DIFF
--- a/libqpdf/QPDFJob_config.cc
+++ b/libqpdf/QPDFJob_config.cc
@@ -24,8 +24,9 @@ to_int(std::string_view option, std::string const& value, int max, int min)
 {
     qpdf_expect(min < max);
     try {
-        int result = std::stoi(value);
-        if (result < min || result > max) {
+        std::size_t pos = 0;
+        int result = std::stoi(value, &pos);
+        if (pos != value.size() || result < min || result > max) {
             int_usage(option, max, min);
         }
         return result;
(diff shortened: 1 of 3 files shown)


jberkenbilt:
@m-holger I plan to merge this and then fix release notes. I will create a PR that includes this and a few others.

m-holger:
@jberkenbilt  agreed

jberkenbilt:
@arhxam This will come along for the ride when I merge #1777 with the release notes fixed to point to the correct version (released after you created your PR).

PR #1767, opened 2026-08-26
Add Linux ARM64 and macOS release binaries

### LINKED ISSUE #1588: macOS precompiled binaries
I'm considering integrating qpdf into [MetadataZero](https://github.com/metadatazero/metadatazero) to reliably strip PDF metadata, but I need to ship a cross-platform binary for Windows, Linux, and macOS (Intel and ARM).

Would you consider providing precompiled macOS binaries for both Intel and ARM?

Shion1305:
## Summary

- add a native Ubuntu 22.04 ARM64 build that packages `qpdf`, `fix-qdf`, and `zlib-flate`
- build and upload separate macOS Intel and Apple Silicon command-line distributions
- statically link qpdf and non-system dependencies where appropriate, then verify architecture and runtime dependencies before upload
- merge the new archives into the existing release artifact and document the new download choices

## Testing

- [fork workflow run](https://github.com/Shion1305/qpdf/actions/runs/32959685945) on GitHub-hosted runners: Linux ARM64, macOS Intel, and macOS Apple Silicon all passed their full build, CTest, package validation, and artifact upload jobs
- `build-scripts/build-mac arm64` on Apple Silicon macOS: all 7 CTest groups passed; the generated ZIP was extracted and all three commands were smoke-tested
- `build-scripts/build-linux-arm64` in a native Ubuntu 22.04 ARM64 container: all 7 CTest groups passed; ELF architecture, runtime dependencies, ZIP extraction, and command smoke tests passed
- `shellcheck build-scripts/build-mac build-scripts/build-linux-arm64`
- `bash -n build-scripts/build-mac build-scripts/build-linux-arm64`
- parsed `.github/workflows/main.yml` as YAML

The runner labels come from the current `actions/runner-images` catalog; local actionlint 1.7.7 predates `macos-15-intel` and reports that label as unknown.

This keeps the existing release authentication flow based on signed checksums. It does not add Apple Developer ID signing or notarization for the macOS executables.

Closes #1588

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -84,12 +84,36 @@ jobs:
           name: distribution-windows-${{ matrix.wordsize }}-${{ matrix.tool }}
           path: distribution
   macOS:
-    runs-on: macos-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - runner: macos-15-intel
+            architecture: x86_64
+          - runner: macos-15
+            architecture: arm64
+    runs-on: ${{ matrix.runner }}
     needs: Prebuild
     steps:
       - uses: actions/checkout@v7
-      - name: 'Mac build and test'
-        run: build-scripts/build-mac
+      - name: 'Mac build, test, and package'
+        run: build-scripts/build-mac ${{ matrix.architecture }}
+      - name: 'Upload binary distribution'
+        uses: actions/upload-artifact@v7
+        with:
+          name: distribution-macos-${{ matrix.architecture }}
+          path: distribution
+  LinuxArm64:
+    runs-on: ubuntu-22.04-arm
+    steps:
+      - uses: actions/checkout@v7
+      - name: 'Linux ARM64 build, test, and package'
+        run: build-scripts/build-linux-arm64
+      - name: 'Upload binary distribution'
+        uses: actions/upload-artifact@v7
+        with:
+          name: distribution-linux-arm64
+          path: distribution
   AppImage:
     runs-on: ubuntu-latest
     needs: Prebuild
@@ -171,7 +195,9 @@ jobs:
      needs:
        - Prebuild
        - Linux
+       - LinuxArm64
        - Windows
+       - macOS
        - AppImage
      steps:
        - name: Merge Artifacts
--- a/README-what-to-download.md
+++ b/README-what-to-download.md
@@ -46,6 +46,18 @@ Ubuntu. However, there are some downloads available for Linux as well.
   and was initially created for that purpose. The executables have their runpath set to looks for the qpdf library
   in `../lib` relative to the location of the executables, which makes this distribution relocatable.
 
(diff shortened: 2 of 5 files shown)


Shion1305:
Hi @jberkenbilt Thank you so much for maintaining this awesome oss!
Could you review this by any chance? I really want to see qpdf releases available in arm and macos!

jberkenbilt:
@Shion1305 Thanks for this contribution. The work you've done provides a valuable roadmap and a template for what the deliverable should look like. I plan on using it as a jumping off point. There are a some aspects of it that I don't think are quite right, so I'm not going to accept it as is, and I don't have time to iterate right now. I'm trying to put a little more time into qpdf, so I'll see if I can land this before the next release or the one after. Thanks for the nudge!

Shion1305:
@jberkenbilt Thanks for the response!
If you could point out roughly which parts don't look quite right, I'd be happy to look into them and update the PR. No need for detailed feedback, just a general direction would be enough!

jberkenbilt:
@Shion1305 Sure -- the Linux arm one looks mostly good, though I don't want to ship official binaries based off qpdf's internal crypto. From a security standpoint, while I have no reason to believe there are issues with it, people don't want to rely on crypto that is not heavily vetted and frequently updated, which is why qpdf can use gnutls or openssl. The official Windows builds use openssl. Another possible approach would be to replicate how the x86_64 binaries are built, which is to piggyback off the appimage build. I haven't tried it or looked at it for a while, but it's possible that that docker-based build may work fine on an arm64 runner and produce an arm64 appimage and linux binaries. You might also consider adding a quick smoke test with ldd to ensure that there are no dynamic library dependencies on the things you're liking statically. Maybe it would be possible to modify the app image scripts so they take a platform string. This is probably the first place I would have gone. I don't know if there are Linux arm64 app images, but if so, it would solve that problem as well. I can easily test with a linux vm on a mac.

For mac, my main concern is the source of the dependencies, where the dependency versions are specified, and how many extra things we have to build. I don't want the ongoing maintenance burden of constantly having to update versions there. We can either do something like hitching our wagon to homebrew or macports and enabling static linking of those libraries, relying on system libraries where possible, or using vcpkg as we do with the Windows builds. Take a look at https://github.com/qpdf/qpdf/blob/main/.github/workflows/vcpkg.yml and how the Windows builds consume this. Maybe it makes sense to do something similar for mac. That gives us reproducibility while keeping us updated and not having to pay the price of building zlib and jpeg each time.

Feel free to keep working on it, and we can keep iterating. If you get all the way there before I have time to work on it, then great. :-)

jberkenbilt:
@Shion1305 I plan on merging #1776 soon which will add Windows Arm64. If you touch vcpgk, wait for it or rebase on top of it to avoid merge conflicts. Unless you're working as I write this, it'll probably be merged before you get to anything. :-)

jberkenbilt:
@Shion1305 See #1779 for Arm64 -- I've tested this locally on my mac in a Linux VM. I'll look at mac builds tomorrow if I have time.

jberkenbilt:
closing; replaced by #1779

PR #1768, opened 2026-08-27
Fix InputSource::read_line when a line's EOL run reaches EOF

federicobond:
The guard before the seek tested eol rather than next_line, so a line with no character after its terminator seeks to last_offset + nposm one byte before the start of the line just read. At offset 0 that is a seek to -1, which fails with EINVAL, so a file whose only line is a PDF header failed with a system error instead of being reported as damaged. At any other offset the result is a valid position, so the input source was left one byte behind with no error at all.

In practice this turns
```
QPDFSystemError: seek to …, offset -1 (0): Invalid argument
```
into
```
QPDFExc: unable to find trailer dictionary while recovering damaged file
```
so a truncated file is reported like any other invalid or corrupt PDF.

Let me know if you would like an integration test for this case.

### DIFF
--- a/manual/release-notes.rst
+++ b/manual/release-notes.rst
@@ -21,6 +21,10 @@ more detail.
     - Detect and warn in check linearization when the xref stream reports the object containing a
       compressed object to itself be a compressed object.
 
+    - Report a file whose only line is a PDF header as damaged instead of failing with a
+      system error. Reading a line whose end-of-line run reaches the end of the file left the
+      input source one byte before the start of that line, which at offset 0 is a seek to -1.
+
   - Enhancements
 
     - Improve uniformity and accuracy of progress reporting when writing linearized files and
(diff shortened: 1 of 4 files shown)


manual/release-notes.rst
```diff
@@ -21,6 +21,10 @@ more detail.
     - Detect and warn in check linearization when the xref stream reports the object containing a
       compressed object to itself be a compressed object.
 
+    - Report a file whose only line is a PDF header as damaged instead of failing with a
```
jberkenbilt, line 24: 12.4.1 is already out, so these should be 12.4.2 release notes. 12.4.2 is on main, and we have a PR that updates the release notes already. Either you can rebase from main and put your release notes in 12.4.2, or if you don't get it, I will merge this and move the release notes section.

jberkenbilt (changes requested):
Release notes shouldn't be added to something that's already released -- see comment. I can fix or you can. Otherwise, the change looks good.

jberkenbilt:
@federicobond No action required from you -- I have fixed the release notes in #1777 which will carry this along for the ride.

manual/release-notes.rst
federicobond: Thank you, completely missed it, sorry 🤦

jberkenbilt: No worries. :-) I appreciate that you thought to add a release note!

PR #1770, opened 2026-08-30
Mono fixes

jberkenbilt:
This PR contains code contributed by a user who had been running with some linearization performance improvement patches that were coded by Claude Opus 4.7. The contributed fixes addressed a real issue in qpdf's implementation, and the thrust of the change was correct, but the implementation didn't go far enough and was not to a standard that would be acceptable in the qpdf codebase. This PR contains the original changes with only cosmetic adjustments (code formatting, re-ordering) followed by a few small commits that make some tactical improvements. This will not be merged, but is being kept as a historical reference. I will create a separate pull request with the real changes and link to it from here.

### DIFF
--- a/libqpdf/QPDF_linearization.cc
+++ b/libqpdf/QPDF_linearization.cc
@@ -142,19 +142,6 @@ Lin::optimize_internal(
         return;
     }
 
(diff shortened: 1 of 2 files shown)


jberkenbilt:
Superseded by #1771

PR #1771, opened 2026-08-30
Lin fixes

jberkenbilt:
This is my hand-coded fixes to the linearization issues fixed by #1770 and supersedes that PR.

| version | max memory | time |
|---|---|---|
| 12.4.1 | 30,386,672,757 | 5:00.73 |
| mono-fixes | 1,104,944,656 | 2:33.79 |
| this | 781,826,593 | 2:30.62 |

This also includes an enhancement to the output of `--show-linearization` and some functionality fixes to linearization output in some corner cases, fixing latent bugs that have been in the code for decades.


jberkenbilt:
I need to rethink the vector<int>s in QPDF_linearize.cc because of memory usage and fuzzing. I'm going to look more closely at ObjectTable.

jberkenbilt:
@m-holger I believe this is ready for review now at your convenience.

libqpdf/QPDF_linearization.cc
```diff
@@ -1270,39 +1344,68 @@ Lin::calculateLinearizationData(T const& object_stream_data)
...
+
+    std::vector<QPDFObjectHandle> uc_pages;
+    std::vector<QPDFObjGen> uc_page_ogs;
+    std::vector<QPDFObjectHandle> uc_thumbs;
+    std::vector<QPDFObjGen> thumb_ogs;
+    { // local scope
+        // Map all page objects to the containing object stream.  This should be a no-op in a
+        // properly linearized file.
+        for (auto oh: pages) {
+            auto uc = getUncompressedObject(oh, object_stream_data);
```
m-holger, line 1377: I don't think compressed pages are possible. We only get here after prepare has been called which triggers a page tree check and repair / exception

jberkenbilt: This is also on the code path for checking linearization, so in theory, you could have a file that claims to be linearized and has compressed page objects. I don't think that's precluded by anything in the code.

m-holger: Yes, I got confused. However, this still means that the file is not properly linearized and therefore we should do a linearization warning

PR #1776, opened 2026-09-05
Add Windows ARM64 build

ndabas:
As the title says. All builds and tests pass on my fork.

I updated some libjpeg interface code in `libqpdf/Pl_DCT.cc`, I'm not sure if the existing code would be classified as UB but I suspect it would be (jumping out of a `catch` block with `longjmp`.) In any case this was causing an infinite loop on Windows ARM64, so I rewrote it a bit to keep the same logic but avoid the livelock.

Note that we would need to run the `vcpkg cache` workflow from this branch before the main build -- I'm not sure if that is even possible, so we might need to add the cache file that includes arm64 packages manually perhaps to the cache release. (I have tested this and it works fine on my fork.)

Another thing I noticed when working on the code on my Windows ARM64 machine is that the `generate_auto_job` script assumes LF newlines and UTF-8 encoding, which is not true on Windows. I made some edits to that file so that it works identically even when called from Windows; let me know if you would like me to add those changes to this PR as well.


jberkenbilt:
@ndabas This is great -- I'll merge it if it passes CI. FWIW, I am close to dumping the mingw build entirely. I'm just one inconvenient problem away from dropping it.

ndabas:
@jberkenbilt Thanks! Did you see my note on the CI, since we need to pre-fill the vcpkg cache first?

>Note that we would need to run the vcpkg cache workflow from this branch before the main build -- I'm not sure if that is even possible, so we might need to add the cache file that includes arm64 packages manually perhaps to the cache release. (I have tested this and it works fine on my fork.)

I would agree with dropping mingw support at this point, those environments are deprecated in MSYS2 anyway.

jberkenbilt:
> Thanks! Did you see my note on the CI, since we need to pre-fill the vcpkg cache first?

Good catch -- I'll run it and restart.

jberkenbilt:
@ndabas Actually, since we have to rerun and you're active, if you could squash the editor config removal with the commit that added it, that would be cleaner. Alternatively, I don't really mind having it committed. It might save someone else some trouble. I'll post back here when I've got vcpkg cache updated.

ndabas:
@jberkenbilt I could squash that, the reason why I left it this way is because the `README-*.md` files being CRLF and the rest of the docs being LF threw me off for a bit while staring at the diff.

I think it'll be cleaner to squash and remove that entirely, because if I were to commit it, I would want it to be a bit more complete and consistent (tab sizes etc.) and that is a bit beyond the scope of this PR. I'll go ahead and do that now.

jberkenbilt:
@ndabas Thanks -- when vcpkg is finished in CI, I will approve the workflows and merge if all is good.

ndabas:
@jberkenbilt thanks, I see you bumped the max-parallel as well -- I did that too for testing in my own fork, but wasn't sure about the rationale behind setting it to 4 previously, so didn't add that to the commits in this PR.

ndabas:
@jberkenbilt I see the job failed because the vcpkg cache job didn't actually upload the release -- the script will only upload if the branch is `main`: https://github.com/qpdf/qpdf/blob/b35de9c17f139cd0b65c94826139b3cf8111d3b8/build-scripts/package-vcpkg#L35-L36

Maybe we do it manually? Or push a temp commit to run it irrespective of the branch.

jberkenbilt:
yeah, it's been a while. I uploaded it manually. This is such a rare event that I'm not sure it's worth automating. Basically I grabbed the artifact from the vcpkg job downloaded, ran the package script locally, and uploaded the artifact. I did all this before I saw your comment.

ndabas:
Excellent, thank you!

jberkenbilt:
Probably I will make it release if I build from a magic branch, like vcpkg-release, so I can do what I just did: manually run on a branch. We may be adding mac binaries or something. Anyway, once this merges, I will run the dispatch from main to get everything official. I had a workflow like this in my day job for pushing container images once....

jberkenbilt:
@ndabas Thanks for your attentiveness. This is the first PR to the build side of things I ever remember being willing to accept without modification.

ndabas:
Thank you for the kind words! I was actually looking to run pikepdf on Windows ARM64, so the rabbit hole led me here...

PR #1780, opened 2026-09-09
Read 32-bit linearization hint offsets as unsigned

zkrige:
`--check` reports a spurious linearization error on any linearized PDF larger than 2 GiB:

```
WARNING: first shared object offset mismatch: hint table = -1041572443; computed = 3253399113
qpdf: operation succeeded with warnings
```

`-1041572443 + 2^32 = 3253394853`, which is the computed value reinterpreted as signed, so the file is fine and the reader is not.

## Cause

`first_page_offset`, `first_shared_offset` and `first_object_offset` are declared `qpdf_offset_t`, but all three were read with `getBitsInt(32)`. `BitStream.hh` states the rule that breaks:

```c++
// Only call getBitsInt when requesting a number of bits that definitely fit in an int.
int getBitsInt(size_t nbits);
```

A 32-bit offset stops fitting in an `int` once the file passes 2^31 bytes.

The write side is already correct. `writeBits(toULL(t.first_shared_offset), 32)` stores the value losslessly, and 32 bits cover the whole 4 GiB range the hint tables allow, so qpdf was misreading files it had itself written correctly.

## Change

Read those three fields with `getBits(32)` through `QIntC::to_offset`. Nothing else in the hint tables is affected: every other 32-bit field is a count, a length or an object number that is genuinely `int`.

## Testing

- New case in `libtests/bits.cc` reading a 32-bit field holding 3253399113, with the expected output updated.
- `ctest` passes in full (`qpdf` and `libtests`).
- Verified against a real 3.03 GiB linearized PDF: `--check-linearization` reports the warning above before the change and `no linearization errors` after, on the same bytes. `--check` is otherwise clean both ways.

I do not have a redistributable file of that size to add as a test fixture, hence the unit-level test on `BitStream` rather than a qpdf-level one.


zkrige:
One boundary worth stating explicitly, since this change touches that code path.

The fix covers 2 GiB to 4 GiB, where the value was written correctly and read wrongly. Above 4 GiB the hint table cannot represent the offset at all: the field is 32 bits, so the ceiling is 2^32 by format, not by implementation.

A real example, a 4.49 GiB PDF linearized by qpdf 12.4.1:

```
check: hint table = 505681064; computed = 4800648360
```

`4800648360 - 505681064 = 4294967296`, exactly 2^32. `writeBits(toULL(...), 32)` truncated it and `--check-linearization` correctly reports the mismatch. Nothing in this PR changes that, and nothing can without a format change.

If it would help, I am happy to add a warning at write time when an offset exceeds 2^32, so the output is flagged as unlinearizable rather than silently wrong. Say the word and I will put it in a separate PR.

